PR #2, opened 2024-03-12
feat: 🚧 `pallet-file-system` init

snowmead:


pallets/file-system/src/types.rs
```diff
@@ -0,0 +1,61 @@
...
+use scale_info::TypeInfo;
+
+use crate::Config;
+
+/// Metadata for a storage request.
+///
+/// This is used to track the status of a storage request
+#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Debug, PartialEq, Eq, Clone)]
+#[scale_info(skip_type_params(T))]
+pub struct StorageRequestMetadata<T: Config> {
```
ffarall, line 13: Shouldn't this also hold the user that started the request? Not just he's multiaddress

pallets/file-system/src/utils.rs
```diff
@@ -0,0 +1,112 @@
...
+        // Check that storage request is not already registered.
+        ensure!(
+            !<StorageRequests<T>>::contains_key(&location),
+            Error::<T>::StorageRequestAlreadyRegistered
+        );
+
+        // Register storage request.
+        <StorageRequests<T>>::insert(&location, file_metadata);
+
+        // TODO: Maybe worth it to loop a few times until we find a free slot. (this should never fail and so a loop is required)
```
ffarall, line 58: Instead, we can have a `StorageValue` that holds the number of the current expiration block (`CurrentExpBlock`). The minimum expiration number of blocks should be the constant `StorageRequestTtl`, so when a new storage request is generated, you append it in `StorageReqeustExpiration[expiration_block]` where `expiation_block = max(StorageRequestTtl, CurrentExpBlock)`. That way, in O(1) you get the next available slot. When a block's vector is filled, you increment `CurrentExpBlock`, and then if `CurrentExpBlock < StorageRequestTtl`, then you update `CurrentExpBlock = StorageRequestTtl`.

pallets/file-system/src/utils.rs
```diff
@@ -0,0 +1,112 @@
...
+            Error::<T>::StorageRequestNotRegistered
+        );
+
+        // Check if the BSP is not already registered for this storage request.
+        ensure!(
+            !file_metadata.bsps_volunteered.contains(&who),
+            Error::<T>::BspAlreadyConfirmed
+        );
+
+        // TODO: Check that the threshold value is high enough to qualify as BSP for the storage request.
```
ffarall, line 95: Just for clarity, the threshold should be higher than the BSP's calculated XOR value.

pallets/file-system/src/utils.rs
```diff
@@ -0,0 +1,112 @@
...
+
+        // Add BSP to storage request metadata.
+        file_metadata
+            .bsps_volunteered
+            .try_push(who.clone())
+            .map_err(|_| Error::<T>::BspVolunteerFailed)?;
+        <StorageRequests<T>>::set(&location, Some(file_metadata.clone()));
+
+        // Check if maximum number of BSPs has been reached.
+        if file_metadata.bsps_volunteered.len() == T::MaxBspsPerStorageRequest::get() as usize {
```
ffarall, line 105: I believe the storage request should acutally be deleted after all BSPs have been *confirmed*. Otherwise, the `bsps_confirmed` field is useless, because a storage request will always be deleted before all BSPs are confirmed.

pallets/file-system/src/types.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// List of BSPs that have volunteered to store the data.
+    pub bsps_volunteered: BoundedVec<StorageProviderId<T>, MaxBspsPerStorageRequest<T>>,
+    /// List of BSPs that have proven they are storing the data.
+    ///
+    /// The storage request will be dropped/complete once all the minimum required BSPs have
+    /// submitted a proof of storage after volunteering to store the data.
+    pub bsps_confirmed: BoundedVec<StorageProviderId<T>, MaxBspsPerStorageRequest<T>>,
+}
+
+/// Alias for the `AccoundId` type used in the FileSystem pallet.
```
ffarall, line 39:
```suggestion
/// Alias for the `AccountId` type used in the FileSystem pallet.
```

pallets/file-system/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+
+#[frame_support::pallet]
+pub mod pallet {
+    use super::types::*;
+    use codec::HasCompact;
+    use frame_support::{
+        dispatch::DispatchResult,
+        pallet_prelude::*,
+        sp_runtime::traits::{AtLeast32Bit, CheckEqual, MaybeDisplay, SimpleBitOps},
+    };
```
ffarall, line 52:
```suggestion
        /// Numerical type for representing units of data. 
        ///
        /// Used to represent the price for data. For example, if the price is 1 DOT, and a unit
        /// here represents 1MB, then 1MB is 1 DOT.
```

pallets/file-system/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+        /// Notifies that a BSP has been accepted to store a given file.
+        AcceptedBspVolunteer {
+            who: T::AccountId,
+            location: FileLocation<T>,
+            fingerprint: Fingerprint<T>,
+            bsp_multiaddress: MultiAddress<T>,
+        },
+
+        /// Notifies the expiration of a storage request.
+        StorageRequestExpired { location: FileLocation<T> },
```
ffarall, line 145: I'd also add an event for when a storage request is fully completed (and therefore deleted). Anyway, we'll have to add many events down the line.

pallets/file-system/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+        /// Storage request already registered for the given file.
+        StorageRequestAlreadyRegistered,
+        /// Storage request not registered for the given file.
+        StorageRequestNotRegistered,
+        /// BSP already volunteered to store the given file.
+        BspVolunteerFailed,
+        /// BSP already confirmed to store the given file.
+        BspAlreadyConfirmed,
+        /// No slot available found in blocks to insert storage request expiration time.
+        StorageRequestExpiredNoSlotAvailable,
```
ffarall, line 160: Actually, this error should never happen

pallets/file-system/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+            }
+
+            let mut used_weight = db_weight.reads(1);
+
+            let mut expired_requests = match StorageRequestExpirations::<T>::take(&block) {
+                Some(requests) => requests,
+                None => return used_weight,
+            };
+
+            // remove expired storage requests requests
```
ffarall, line 274:
```suggestion
            // Remove expired storage requests
```

pallets/file-system/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+                        .try_into()
+                        .expect("u32 would fit into u64"),
+                ),
+            ) {
+                return Weight::zero();
+            }
+
+            let mut used_weight = db_weight.reads(1);
+
+            let mut expired_requests = match StorageRequestExpirations::<T>::take(&block) {
```
ffarall, line 269: Is it possible that this doesn't get executed for block.

pallets/file-system/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+                // that are relevant to them (e.g. include the SPs that have volunteered to store the file)
+                let _request = StorageRequests::<T>::take(&location);
+
+                StorageRequests::<T>::remove(&location);
+                used_weight += db_weight.writes(1);
+
+                Self::deposit_event(Event::StorageRequestExpired { location });
+            }
+
+            // we already checked we have enough `remaining_weight` to cover this `used_weight`
```
ffarall, line 286:
```suggestion
            // We already checked we have enough `remaining_weight` to cover this `used_weight`
```

pallets/file-system/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+
+            let mut expired_requests = match StorageRequestExpirations::<T>::take(&block) {
+                Some(requests) => requests,
+                None => return used_weight,
+            };
+
+            // remove expired storage requests requests
+            for location in expired_requests.drain(..) {
+                // TODO: should probably add some fields to the `StorageRequestExpired` to facilitate SPs filtering the events
+                // that are relevant to them (e.g. include the SPs that have volunteered to store the file)
```
ffarall, line 277:
```suggestion
                // That are relevant to them (e.g. include the SPs that have volunteered to store the file)
```

ffarall (changes requested):

pallets/file-system/src/utils.rs
snowmead: I realized that the `submit_proof` is in the storage proofs pallet, this means that for `bsps_confirmed` to be incremented, the storage proofs pallet needs to update this in the file system pallet. Or you can have the `submit_proof` in the file system pallet

pallets/file-system/src/lib.rs
snowmead: Yes definitely, I add the events when I use them, just to avoid having too much unused code

pallets/file-system/src/lib.rs
snowmead: I am not sure I understand, can you clarify

pallets/file-system/src/lib.rs
snowmead: In the off chance the storage request expiry fails to append to a `BoundedVec`, we need to return an error from the extrinsic, and it still should tell you what happened (even if it will never happen xD)

ffarall: Sounds reasonable. In that case I'd use the `expect_or_err` macro so that in tests this panics with unreachable, but in wasm it reverts with that error.

pallets/file-system/src/utils.rs
ffarall: This is part of a broader conversation we need to have on circular dependencies of File System with Storage Proofs pallet. The File System pallet should also be able to submit custom challenges to the Storage Proofs pallet, for file deletions or modifications. And on the other side, Storage Proofs pallet should be able to read from File System what are the Merkle Patricia Forest roots for BSPs and buckets. One option could be that we move the storage of roots to the Storage Providers pallet. Then Storage Proofs would depend just on Storage Providers, and File System would depend con Storage Providers + Storage Proofs.

pallets/file-system/src/lib.rs
ffarall: Yes sorry. I meant to say that there should be a way to keep track of the last block for which this `on_idle` was executed, and effectively cleared storage. This is in case that in some heavily used block, `on_idle` doesn't get to be executed, so it lags behind.

pallets/file-system/src/lib.rs
snowmead: good idea

pallets/file-system/src/lib.rs
snowmead: Put some TODOs for these scenarios.

pallets/file-system/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+        fn on_idle(block: BlockNumberFor<T>, remaining_weight: Weight) -> Weight {
+            let db_weight = T::DbWeight::get();
+
+            // Return early if the remaining weight is not enough to perform the operation
+            // TODO: fix this, it's not working as expected (when the remaining weight is 0, it stil goes through)
+            if !remaining_weight
+                .all_gte(db_weight.reads_writes(1, T::MaxExpiredStorageRequests::get().into()))
+            {
+                CurrentExpirationBlock::<T>::put(block.saturating_add(1u32.into()));
+                return Weight::zero();
```
ffarall, line 282: The logics of this `on_idle` still need to be polished, but it is ok for this PR's scope.

ffarall (approved):

PR #4, opened 2024-03-13
remove storage-hub-node and mv all to root workspace

snowmead:


README.md
timbrinded, line 1: probably worth appending instead of replacing

ffarall (approved):

PR #11, opened 2024-03-18
feat: pallet-file-system clear storage requests up to current block `on_idle`

snowmead:
Refactored `on_idle` to manage clearing storage requests from old blocks up to the current block if any based on `NextStartingBlockToCleanUp`. 

Iterating over every old block number until we reach the current block while clearing all expired storage requests from `StorageRequestExpirations`. 

This iteration maintains the `total_used_weight` to avoid exceeding the `remaining_weight` for the current block.

### DIFF
--- a/pallets/file-system/src/lib.rs
+++ b/pallets/file-system/src/lib.rs
@@ -313,24 +313,18 @@ pub mod pallet {
             let start_block = NextStartingBlockToCleanUp::<T>::get();
             let mut block_to_clean = start_block;
 
-            // Calculate the weight required to process the cleanup
-            let required_weight =
-                db_weight.reads_writes(1, (T::MaxExpiredStorageRequests::get() + 1u32).into()); // Adding 1 for `NextBlockToCleanup` write at the end
-
-            // Check if there is enough weight to process entire `on_idle` execution with a single block cleanup iteration
-            remaining_weight.all_gte(required_weight);
-
             // Total weight used to avoid exceeding the remaining weight
             let mut total_used_weight = Weight::zero();
 
+            let required_weight_for_iteration =
+                db_weight.reads_writes(1, T::MaxExpiredStorageRequests::get().into());
+
             // Iterate over blocks from the start block to the current block,
             // cleaning up storage requests until the remaining weight is insufficient
-            while block_to_clean <= current_block && remaining_weight.all_gte(total_used_weight) {
-                // Check if there is enough weight left to process the cleanup for one more block
-                if required_weight.all_gte(remaining_weight.saturating_sub(total_used_weight)) {
-                    break;
-                }
-
+            while block_to_clean <= current_block
+                && remaining_weight
+                    .all_gte(total_used_weight.saturating_add(required_weight_for_iteration))
+            {
                 let mut used_weight = db_weight.reads(1);
                 let expired_requests = StorageRequestExpirations::<T>::take(&block_to_clean);
 
--- a/pallets/file-system/src/lib.rs
+++ b/pallets/file-system/src/lib.rs
@@ -134,7 +134,7 @@ pub mod pallet {
     /// storage request expiration will be inserted in the block `StorageRequestTtl` ahead, and then
     /// this value will be reset to block number a `current_block` + `StorageRequestTtl`.
     #[pallet::storage]
-    #[pallet::getter(fn current_expiration_block)]
+    #[pallet::getter(fn next_available_expiration_insertion_block)]
     pub type NextAvailableExpirationInsertionBlock<T: Config> =
         StorageValue<_, BlockNumberFor<T>, ValueQuery>;
 
@@ -172,7 +172,7 @@ pub mod pallet {
         /// Notifies the expiration of a storage request.
         StorageRequestExpired { location: FileLocation<T> },
 
-        /// Notifies that a storage request has been deleted.
+        /// Notifies that a storage request has been revoked by the user who initiated it.
         StorageRequestRevoked { location: FileLocation<T> },
     }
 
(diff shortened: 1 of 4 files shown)


TDemeco (approved):
LGTM 👍

pallets/file-system/src/lib.rs
```diff
@@ -234,8 +240,20 @@ pub mod pallet {
         /// Revoke storage request
         #[pallet::call_index(2)]
         #[pallet::weight(Weight::from_parts(10_000, 0) + T::DbWeight::get().writes(1))]
-        pub fn revoke_storage_request(_origin: OriginFor<T>) -> DispatchResult {
-            todo!()
+        pub fn revoke_storage_request(
```
ffarall, line 243: This should actually iniciate a delete request, in case some BSPs or the MSP already confirmed the storage request and added the file to their tries. We have to make sure they deleted it. Please add it as a TODO at least

pallets/file-system/src/lib.rs
```diff
@@ -284,38 +302,60 @@ pub mod pallet {
...
+            // cleaning up storage requests until the remaining weight is insufficient
+            while block_to_clean <= current_block && remaining_weight.all_gte(total_used_weight) {
+                // Calculate the weight required to process the cleanup
+                let required_weight =
+                    db_weight.reads_writes(1, T::MaxExpiredStorageRequests::get().into());
+
+                // Check if there is enough weight left to process the cleanup for one more block
+                if required_weight.all_gte(remaining_weight.saturating_sub(total_used_weight)) {
+                    break;
+                }
```
ffarall, line 326: While it may work, this logic is a bit confusing and hard to follow, which could be problematic for debugging.

pallets/file-system/src/lib.rs
```diff
@@ -284,38 +302,60 @@ pub mod pallet {
...
+                for location in expired_requests {
+                    StorageRequests::<T>::remove(&location);
+                    used_weight += db_weight.writes(1);
+                    Self::deposit_event(Event::StorageRequestExpired { location });
+                }
+
+                // Accumulate the weight used for cleanup operations
+                total_used_weight += used_weight;
+                // Increment the block to clean up for the next iteration
+                block_to_clean = block_to_clean.saturating_add(1u32.into());
```
ffarall, line 341: This should be a `checked_add`

pallets/file-system/src/lib.rs
ffarall: Plus, if it actually happens that `remaining_weight.all_gte(total_used_weight)` is `false`, we're already screwed, because more weight than what was remaining has ALREADY been used, it's too late.

pallets/file-system/src/lib.rs
```diff
@@ -284,38 +302,60 @@ pub mod pallet {
...
+                // Check if there is enough weight left to process the cleanup for one more block
+                if required_weight.all_gte(remaining_weight.saturating_sub(total_used_weight)) {
+                    break;
+                }
+
+                let mut used_weight = db_weight.reads(1);
+                let expired_requests = StorageRequestExpirations::<T>::take(&block_to_clean);
+
+                // Remove expired storage requests for the block
+                for location in expired_requests {
```
ffarall, line 332: For example, in here, it is never checked if there is going to be enough weight left to cleanup all the expired requests (to execute all the `for` loop). Instead, you clear them, register the weight, and only after, in the next `while` loop iteration, you check if you went too far.

pallets/file-system/src/lib.rs
```diff
@@ -284,38 +302,60 @@ pub mod pallet {
...
+                // `NextStartingBlockToCleanUp` is always updated to start from the block we reached in the current `on_idle` call.
+                Some(start_block) if block_to_clean > start_block => {
+                    NextStartingBlockToCleanUp::<T>::put(block_to_clean);
+                }
+                // `NextStartingBlockToCleanUp` is initialized here for the first time if we have not cleaned up the current block.
+                None if block_to_clean == current_block => {
+                    NextStartingBlockToCleanUp::<T>::put(block_to_clean);
+                }
+                // In all other cases, this means we processed all the blocks up to the current block
+                _ => {}
```
ffarall, line 355: This is prone to confusion. I believe `NextStartingBlockToCleanUp::<T>::get()` should actually be `ValueQuery` and return `0` if it has never been accessed. That way we make sure that the cleanup always starts checking storage requests to delete right from the first block.

ffarall (changes requested):

pallets/file-system/src/lib.rs
snowmead: yes will do, kindof added this quickly in the pr just to have something there.

pallets/file-system/src/lib.rs
snowmead: This is not true because we check first here in the while if we can even enter the iteration to begin with. Within the iteration itself we check if we can process another set of deletions, if not we exit and we still haven’t exceeded the `remaining_weight`.

`while block_to_clean <= current_block && remaining_weight.all_gte(total_used_weight) { … }`

pallets/file-system/src/lib.rs
snowmead: Before this loop I check if we can clear the potential total maximum of expired storage requests.

```rust
// Calculate the weight required to process the 
let required_weight = db_weight.reads_writes(1, T::MaxExpiredStorageRequests::get().into());

// Check if there is enough weight left to process the cleanup for one more block
if required_weight.all_gte(remaining_weight.saturating_sub(total_used_weight)) {
    break;
}
```

snowmead: So there is no chance of exceeding the `remaining_weight` because I check for max.

pallets/file-system/src/lib.rs
snowmead: Ya I agree with this.

pallets/file-system/src/lib.rs
ffarall: Aaahh good point, yes

ffarall (approved):
LGTM

PR #12, opened 2024-03-19
refactor: ♻️ `storage-proofs` -> `proofs-dealer-trie`

ffarall:
This PR:
- Makes the old `storage-proofs` pallet be abstracted over the fact that the providers are storage providers (now they're just "Providers").
- Defines a trait to be implemented by another pallet, that would give it information of the Providers. This trait should be implemented by the Storage Providers pallet.

### DIFF
--- a/pallets/storage-proofs/src/lib.rs
+++ b/pallets/storage-proofs/src/lib.rs
@@ -399,9 +399,7 @@ pub trait ProvidersInterface {
     /// The type which represents a registered Provider.
     type Provider: Parameter + Member + MaybeSerializeDeserialize + Debug + Ord + MaxEncodedLen;
     /// The type corresponding to the staking balance of a registered Provider.
-    type Balance: fungible::Inspect<Self::AccountId>
-        + fungible::hold::Inspect<Self::AccountId>
-        + fungible::freeze::Inspect<Self::AccountId>;
+    type Balance: fungible::hold::Inspect<Self::AccountId>;
     /// The type corresponding to the root of a registered Provider.
     type MerkleHash: Parameter
         + Member
--- a/pallets/storage-proofs/src/lib.rs
+++ b/pallets/storage-proofs/src/lib.rs
@@ -22,7 +22,7 @@ use frame_support::{
     inherent::IsFatalError,
     pallet_prelude::*,
     sp_runtime::{
-        traits::{AtLeast32BitUnsigned, CheckEqual, MaybeDisplay, SimpleBitOps},
+        traits::{CheckEqual, MaybeDisplay, SimpleBitOps},
         RuntimeString,
     },
     traits::fungible,
@@ -45,7 +45,7 @@ pub mod pallet {
     use frame_system::pallet_prelude::*;
     use scale_info::prelude::fmt::Debug;
     use sp_trie::CompactProof;
-    use types::SpFor;
+    use types::ProviderFor;
 
     use crate::types::*;
     use crate::*;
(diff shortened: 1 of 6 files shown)


pallets/storage-proofs/src/lib.rs
```diff
@@ -462,8 +418,17 @@ pub trait FileSystemInterface {
         + MaxEncodedLen
         + FullCodec;
 
-    /// Check if a Merkle Patricia Forest root hash belongs to a given Storage Provider.
-    fn is_valid_root_for_sp(who: Self::StorageProvider, root: Self::ForestRoot) -> bool;
+    /// Check if an account is a registered Provider.
+    fn is_sp(who: Self::Provider) -> bool;
```
snowmead, line 422:
```suggestion
    fn is_provider(who: Self::Provider) -> bool;
```

pallets/storage-proofs/src/lib.rs
```diff
@@ -462,8 +418,17 @@ pub trait FileSystemInterface {
...
+
+    // Get Provider from AccountId, if it is a registered Provider.
+    fn get_sp(who: Self::AccountId) -> Option<Self::Provider>;
```
snowmead, line 425:
```suggestion
    fn get_provider(who: Self::AccountId) -> Option<Self::Provider>;
```

snowmead (approved):
I like the new name 👍🏼 

But I think `pallet-proofs-dealer` would be better, I don’t mind if you leave it as is.

pallets/storage-proofs/src/lib.rs
```diff
@@ -387,67 +390,20 @@ impl InherentError {
...
+pub trait ProvidersInterface {
     /// The type which can be used to identify accounts.
     type AccountId: Parameter + Member + MaybeSerializeDeserialize + Debug + Ord + MaxEncodedLen;
-    /// The type corresponding to the staking balance of a registered Storage Provider.
+    /// The type which represents a registered Provider.
+    type Provider: Parameter + Member + MaybeSerializeDeserialize + Debug + Ord + MaxEncodedLen;
+    /// The type corresponding to the staking balance of a registered Provider.
     type Balance: fungible::Inspect<Self::AccountId>
         + fungible::hold::Inspect<Self::AccountId>
         + fungible::freeze::Inspect<Self::AccountId>;
```
TDemeco, line 404: I get this Balance type but at the same time I'm not sure I can return a value that has all three traits in itself, which is why I think there's an issue with the return type below here

pallets/storage-proofs/src/lib.rs
```diff
@@ -462,8 +418,17 @@ pub trait FileSystemInterface {
...
+    fn is_provider(who: Self::Provider) -> bool;
+
+    // Get Provider from AccountId, if it is a registered Provider.
+    fn get_provider(who: Self::AccountId) -> Option<Self::Provider>;
+
+    /// Get the root for a registered Provider.
+    fn get_root(who: Self::Provider) -> Option<Self::MerkleHash>;
+
+    /// Get the stake for a registered  Provider.
+    fn get_stake(who: Self::Provider) -> Option<Self::Balance>;
```
TDemeco, line 431: This is the return type I mean. I think, for it to be a value and not a type or trait, it can't have all three required traits specified above. I might be mistaken though, let's chat about this.

TDemeco (changes requested):
LGTM but I left some comments that are important for a trait implementation in `pallet-providers`

TDemeco (approved):
LGTM 👍

PR #17, opened 2024-03-20
feat(proofs-dealer): ✨ Add `challenge` extrinsic

ffarall:


pallets/proofs-dealer/src/utils.rs
```diff
@@ -1,25 +1,51 @@
...
 impl<T> Pallet<T>
 where
     T: pallet::Config,
 {
-    // TODO: Document.
+    /// Add custom challenge to ChallengesQueue.
+    ///
+    /// Check if sender is a registered Provider. If it is not, charge a fee for the challenge.
+    /// This is to prevent spamming the network with challenges. If the challenge is already queued,
+    /// just return. Otherwise, add the challenge to the queue.
```
snowmead, line 30: I am not sure I understand the reason behind this. 
This opens up the possibility for a storage provider to stall the network (intentionally or unintentionally).

pallets/proofs-dealer/src/tests.rs
```diff
@@ -0,0 +1,347 @@
...
+            &1,
+            user_balance
+        ));
+
+        // Mock a FileKey.
+        let file_key = BlakeTwo256::hash(b"file_key");
+
+        // Dispatch challenge extrinsic twice.
+        assert_ok!(ProofsDealer::challenge(RuntimeOrigin::signed(1), file_key));
+        assert_ok!(ProofsDealer::challenge(RuntimeOrigin::signed(1), file_key));
```
snowmead, line 148: Should we allow multiple challenges for the same file_key? I only see this causing more transactions and work having to be done by storage providers.

snowmead (approved):
lgtm 👍🏼 just a couple of comments

TDemeco (approved):
LGTM 👍

pallets/proofs-dealer/src/utils.rs
ffarall: It doesn't really stall it, it just fills it up with challenges that they themselves are gonna have to answer to. And in any case, it goes to the regular challenges queue, not the priority, which is the one used for deletion requests and so on. I have a todo somewhere in the extrinsic that uses this about considering only allowing free challenges when there has been a change of an MSP, but that would introduce a dependency of this pallet with the File System pallet, that I'd like to avoid now.

snowmead: I like the last comment about specific scenarios when it is free. I don’t think it should be free all the time. But we can discuss about this another time.

pallets/proofs-dealer/src/tests.rs
snowmead: oh my bad misread.

PR #20, opened 2024-03-23
feat: ✨ Define and implement `ProofsDealerInterface` trait to loosely couple File System pallet to Proofs Dealer pallet

ffarall:


snowmead (approved):
Nice 👍🏼

PR #21, opened 2024-03-25
BSP Architecture and initial commit

links234:


storage-kit/src/traits.rs
```diff
@@ -0,0 +1,74 @@
...
+    fn set_metadata(&self, key: &FileKey, metadata: &FileMetadata) -> impl std::future::Future<Output = ()> + Send;
+
+    /// Get a file chunk from storage.
+	fn get_chunk(&self, key: &FileKey, chunk: u64) -> impl std::future::Future<Output = Option<FileChunk>> + Send;
+
+    /// Write a file chunk in storage.
+    fn write_chunk(&self, key: &str, chunk: u64, data: &FileChunk) -> impl std::future::Future<Output = ()> + Send;
+}
+
+pub trait Actor: Sized {
```
snowmead, line 20: I would add a bunch of documentation everywhere for the traits and functions.

storage-kit/src/p2p/actor.rs
```diff
@@ -0,0 +1,122 @@
...
+		// Listen on all interfaces on the specified port.
+		swarm.listen_on(format!("/ip4/0.0.0.0/tcp/{}", port).parse()?)?;
+
+		Ok(P2PModule {
+			swarm,
+		})
+	}
+}
+
+pub struct P2PEventLoop {
```
snowmead, line 62: Shouldn’t the `P2PEventLoop` have a hashmap of senders to send messages to other actors? 

Example:
```rust
senders: HashMap<ActorHandle<A>, Sender<C>>, // A being an ActorHandle and C being a Command type for an actor handle
```
This would give flexibility for the event loops like P2P or Blockchain module to send commands to each other but also give possibility to add other actors in the future.

storage-kit/src/types.rs
snowmead, line 1: We could possibly generate the types using `subxt` here with the macro to expand types from the scale metadata and have it public so all other modules can use it out of the box.

snowmead (commented):
A few questions and comments 👍🏼

bsp/Cargo.toml
```diff
@@ -6,3 +6,7 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+storage-kit = { path = "../storage-kit" }
```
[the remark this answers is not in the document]
links234: Agree, but I think we should do this later.

bsp/src/main.rs
```diff
@@ -1,3 +1,21 @@
...
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let filter = tracing_subscriber::EnvFilter::try_from_default_env()?;
+    tracing_subscriber::FmtSubscriber::builder()
+        .with_env_filter(filter)
+        .finish()
+        .try_init()?;
+
+    let mut storage_kit_manager = StorageKitBuilder::new()
+        .start()
```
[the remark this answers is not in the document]
links234: > I would rename this to build since below is when we actually start the process with start_as_bsp().

My concern is that build it is a bit misleading since it actually starts 2 threads (for now) in the background.

> I would have the builder be able to specify run_as(Roles::BSP), and then when you start below, it will execute all processes related to the roles you chose in the builder.

Is there a reason why would someone want to start as multiple roles at once? :D

I agree it is cleaner to make the builder just build. Will refactor this.

storage-kit/src/types.rs
links234: Good to know. If it works that easily, sure.

I added these types as a "mock" mostly.

storage-kit/src/p2p/actor.rs
links234: If there is no other way around, we can add it later.

However, I'd like to aim to have no cycles (or as little as possible) between actors since it's really easy to make deadlocks otherwise.

I don't see why the Actors have to talk directly with each other. I imagine the higher level tasks would "resolve" that.

links234: Also, this PR does not include yet the event dispatcher.

Each of the actors/modules will have access to a global event dispatcher (mostly to dispatch events). But there is no constraint why actor A couldn't register to actor B's events.

storage-kit/src/p2p/actor.rs
```diff
@@ -0,0 +1,122 @@
...
+use tokio::{select, sync::{mpsc::Receiver, oneshot}};
+use std::time::Duration;
+use libp2p::{futures::StreamExt, identify, identity::Keypair, noise, swarm::NetworkBehaviour, Multiaddr, Swarm};
+use tracing::*;
+
+use crate::{Actor, ActorEventLoop, Port};
+
+/// Defines max_negotiating_inbound_streams constant for the swarm.
+/// It must be set for large plots.
+const SWARM_MAX_NEGOTIATING_INBOUND_STREAMS: usize = 100000;
```
links234, line 11: _(question from author)_

Should we centralize **all** the constants in 1 file?

storage-kit/src/types.rs
links234: Would it be cleaner/easier to have a shared crate between storage kit and the node that would contain these types? Probably other things would fall there too (i.e. the Merkle tree implementations?)

bsp/src/main.rs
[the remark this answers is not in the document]
snowmead: I guess it’s either they run the executable twice, once as a BSP and the other as an MSP. Or they run as both for the same executable.

bsp/Cargo.toml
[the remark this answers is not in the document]
snowmead: I agree it isn’t necessary, but in the end it is a feature flag on the bsp and msp modules and the storage kit builder functions. I leave it up to you, if you want to do this in another pr.

storage-kit/src/p2p/actor.rs
snowmead: Ok I see what you mean. But I don’t understand how the higher level process would not rely on the Blockchain module to begin a storage request for example. The higher level process manages all executions in some sequence using commands. Do you intend to have some higher level processes listen on blockchain events coming from the blockchain module to begin a process? 

1. Higher level process listening to events <- *__does it listen to events, and from where?__*
2. Receives storage request <- *__how does this happen__*
3. Spawn task (or loop listener) blockchain module
a. Execute volunteer transaction
b. Return success to higher level process
4. Send command (storage request) P2P module
a. Connect to user and get file
b. Return success to higher level process
5. Successfully processed storage request

storage-kit/src/types.rs
snowmead: Ya probably.

storage-kit/src/p2p/actor.rs
snowmead: I prefer the latter actually.

storage-kit/Cargo.toml
```diff
@@ -6,3 +6,27 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
```
ffarall, line 8: Use workspace dependencies

ffarall (commented):
Overall I think you did a pretty amazing job, and your approach seems well guided in the decisions you took. Looking forward to discussing this further tomorrow and boil it down to some concrete tasks to do.

storage-kit/src/traits.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    /// The event loop associated with the actor.
+    /// If no custom event loop is needed, the default `EventLoop<Self>` can be used.
+    type EventLoop: ActorEventLoop<Self> + Send + 'static;
+
+    /// Handles a message received by the actor.
+    ///
+    /// * `message` - The message to be handled.
+    ///
+    fn handle_message(
```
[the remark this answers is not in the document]
links234: I'd really like to have the same interface as Actix where you do `Handle<T: Message>` for each _different_ message type but the problem is you end up either having to do dynamic dispatch or have to do very heavy code generation (at least seems that is possible this way). The latter wouldn't work on anything written outside the crate anyway.

Here `Message` is an `enum` that contains all possible commands/queries. For this specific part ("lower-level" modules) I think is the best approach anyway since you don't really get any benefit by isolating implementations for each message. We also get the static analysis and the autocomplete for `P2PCommand::`.

storage-kit/src/traits.rs
```diff
@@ -0,0 +1,128 @@
...
+        &mut self,
+        message: Self::Message,
+    ) -> impl std::future::Future<Output = ()> + std::marker::Send;
+}
+
+/// Trait representing an event loop for an actor.
+pub trait ActorEventLoop<T: Actor> {
```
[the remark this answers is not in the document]
links234: This is just for composability. If you want to do something very specific (i.e. manage a network resource like `P2PModule`) you will have to implement your own event loop, otherwise you can use the provided one (`EventLoop`).

storage-kit/src/p2p/actor.rs
links234: I've updated and implemented event busses in the meantime. Please check [storage-kit/src/tasks.rs](https://github.com/Moonsong-Labs/storage-hub/pull/21/files#diff-d0773164cc29978634c85691ef7dff6355c29c9a4117633ba5764f9b7635603d) to see how an event-triggered task would interact with the different actors/modules.
Check [this](https://github.com/Moonsong-Labs/storage-hub/pull/21/files#diff-68316775a6d5e33dfad7aff44fe5488f64547456625fb81e649a1a3aad9de422R22) to see how this task is registered.

Let me know if this is not clear enough or if it doesn't cover everything you were thinking of.

PR #24, opened 2024-03-26
feat ✨: BSP confirm storing data, BSP volunteer threshold calculation

snowmead:
This PR introduces the following features:
- Verify accounts are BSPs for relevant actions.
- BSP can stop storing a file.
- Challenge (with priority) when storage request is revoked.
- New `SubscribeProvidersInterface` which pushes important events to subscribers. 
  *Implemented by the file system pallet to react to BSPs signing up and signing off. It only supports a single subscriber implementer. In the future we can potentially make it accept a vec if needed.*
- Compute asymptotic thresholds eligibility when BSPs sign up or sign off
- Renamed `StorageProvidersInterface` -> `MutateProvidersInterface` and moved to support traits
- New `ReadProvidersInterface` trait inherits from `ProvidersInterface`.


support/traits/src/lib.rs
```diff
@@ -49,6 +50,87 @@ pub trait ProvidersInterface {
...
+    ) -> DispatchResult;
+
+    /// Remove a root from a bucket of a MSP, removing the whole bucket from storage
+    fn remove_root_bucket(bucket_id: Self::BucketId) -> DispatchResult;
+
+    /// Remove a root from a BSP. It will remove the whole BSP from storage, so it should only be called when the BSP is being removed.
+    /// todo!("If the only way to remove a BSP is by this pallet (bsp_sign_off), then is this function actually needed?")
+    fn remove_root_bsp(who: &Self::AccountId) -> DispatchResult;
+}
+
```
TDemeco, line 133: I don't know if you saw on Slack but Facu wanted to discuss this as this might not be needed. I'll ping you on the thread again

TDemeco (changes requested):
Looks good! Just commented some minor stuff

pallets/file-system/src/utils.rs
```diff
@@ -176,10 +202,91 @@ where
...
+        );
+
+        // Check that the proof is valid.
+        ensure!(
+            <T::ProofDealer as storage_hub_traits::ProofsDealerInterface>::verify_proof(
+                &bsp, &root, &proof,
+            )
+            .is_ok(),
+            Error::<T>::InvalidProof
+        );
```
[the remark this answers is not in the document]
snowmead: Yes, we should remove the old one when it is merged.

pallets/file-system/src/utils.rs
```diff
@@ -159,7 +161,31 @@ where
...
+            .saturating_sub(file_metadata.requested_at)
+            .saturated_into::<u32>();
+
+        let rate_increase = blocks_since_requested
+            .saturating_mul(T::AssignmentThresholdMultiplier::get())
+            .saturated_into::<T::AssignmentThreshold>();
+
+        let threshold = rate_increase.saturating_add(T::MinBspsAssignmentThreshold::get());
+
+        ensure!(bsp_threshold <= threshold, Error::<T>::ThresholdTooHigh);
```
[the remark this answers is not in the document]
snowmead: > Just a personal opinion so no change needed, but the naming of the variables is a bit confusing to me, and the order of operations in the code as well

I added comments and moved the casting to the `compute_bsp_xor` method to have it more readable here. I also renamed the BSP threshold variable.

> It's a minor edge case, but shouldn't it be valid if the XOR between the file's fingerprint and the BSP ID is equal to the current threshold?

The `ensure!` checks if the `bsp_threshold` is equal or less then the threshold.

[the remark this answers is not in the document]
TDemeco: I don't know how I missed that equal sign, sorry about that 😆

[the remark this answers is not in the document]
snowmead: np :)

support/traits/src/lib.rs
snowmead: As discussed, this trait method is removed.

TDemeco (approved):
LGTM 👍

pallets/file-system/src/utils.rs
```diff
@@ -159,7 +161,32 @@ where
...
+                .map_err(|_| Error::<T>::FailedToEncodeFingerprint)?,
+            &who.encode()
+                .try_into()
+                .map_err(|_| Error::<T>::FailedToEncodeBsp)?,
+        )?;
+
+        // Get number of blocks since the storage request was issued.
+        let blocks_since_requested = <frame_system::Pallet<T>>::block_number()
+            .saturating_sub(file_metadata.requested_at)
+            .saturated_into::<u32>();
```
ffarall: Why as `u32`?

pallets/file-system/src/utils.rs
```diff
@@ -294,4 +407,42 @@ where
...
+        file_key: &FileKey<T>,
+        size: StorageData<T>,
+    ) -> DispatchResult {
+        // Challenge BSP to force update its storage root to uninclude the file.
+        ensure!(
+            <T::ProofDealer as storage_hub_traits::ProofsDealerInterface>::challenge_with_priority(
+                file_key
+            )
+            .is_ok(),
+            Error::<T>::InvalidProof
```
ffarall, line 427: Is this the right error? Doesn't seem like a proof has anything to do here.

pallets/file-system/src/utils.rs
```diff
@@ -294,4 +407,42 @@ where
...
+
+    /// Calculate the XOR of the fingerprint and the BSP.
+    fn compute_bsp_xor(
+        fingerprint: &[u8; 32],
+        bsp: &[u8; 32],
+    ) -> Result<T::AssignmentThreshold, Error<T>> {
+        let mut xor_result = Vec::with_capacity(32);
+        for i in 0..32 {
+            xor_result.push(fingerprint[i] ^ bsp[i]);
+        }
```
ffarall, line 446:
```suggestion
        xor_result = fingerprint.iter()
            .zip(bsp.iter())
            .map(|(&x1, &x2)| x1 ^ x2)
            .collect()

```

pallets/file-system/src/utils.rs
```diff
@@ -176,10 +206,91 @@ where
...
+        location: FileLocation<T>,
+        root: FileKey<T>,
+        proof: Proof<T>,
+    ) -> DispatchResult {
+        let bsp = match <T::Providers as storage_hub_traits::ReadProvidersInterface>::get_provider(
+            who.clone(),
+        ) {
+            Some(bsp) => bsp,
+            None => return Err(Error::<T>::NotABsp.into()),
+        };
```
ffarall, line 220: Why using `expect_or_err` here? `location` is a parameter passed by the user, it is perfectly possible that it is wrong.

pallets/file-system/src/utils.rs
```diff
@@ -176,10 +206,91 @@ where
...
+
+        // Increment the number of confirmed storage providers.
+        <StorageRequests<T>>::try_mutate(&location, |file_metadata| -> DispatchResult {
+            let file_metadata = file_metadata
+                .as_mut()
+                .ok_or(Error::<T>::StorageRequestNotFound)?;
+
+            match file_metadata
+                .bsps_confirmed
+                .checked_add(&T::StorageRequestBspsRequiredType::one())
```
ffarall, line 267: Shouldn't we check if there are already enough BSPs confirmed at the beginning? Before doing heavy comptation like `verify_proof`?

pallets/file-system/src/tests.rs
```diff
@@ -293,18 +295,19 @@ fn bsp_volunteer_success() {
...
 
-        let user = RuntimeOrigin::signed(1);
-        let bsp = RuntimeOrigin::signed(2);
+        let owner = RuntimeOrigin::signed(AccountId32::new([1; 32]));
+        let bsp_account_id = AccountId32::new([2; 32]);
+        let bsp = RuntimeOrigin::signed(bsp_account_id.clone());
         let location = FileLocation::<Test>::try_from(b"test".to_vec()).unwrap();
-        let file_content = b"test".to_vec();
-        let fingerprint = BlakeTwo256::hash(&file_content);
+        // bypass the volunteer assignment threshold
```
ffarall, line 302: Add a `TODO` here, to remember to change this when we can actually verify the threshold properly

ffarall (changes requested):
Overall pretty good. Just some minor comments/fixes, and I'd like to see some tests for the bsp confirm extrinsic.

pallets/file-system/src/lib.rs
```diff
@@ -51,14 +51,54 @@ pub mod pallet {
...
+
+        /// The multiplier increases the threshold over time (blocks) which increases the
+        /// likelihood of a BSP successfully vulenteering to store a file.
+        type AssignmentThresholdMultiplier: Get<Self::ThresholdType>;
+
+        /// Horizontal asymptote which the volunteering threshold approaches as more BSPs are registered in the system.
+        type AssignmentThresholdAsymptote: Get<Self::ThresholdType>;
+
+        /// Asymptotic decay function for the assignment threshold.
+        type AssignmentThresholdDecayFunction: Get<Self::ThresholdType>;
```
ffarall, line 100: `AssignmentThresholdDecayFunction` sounds a bit weird as a name. How about `AssignmentThresholdDecayFactor`

pallets/file-system/src/lib.rs
```diff
@@ -188,37 +218,80 @@ pub mod pallet {
...
+                    .unwrap_or_else(|_| panic!("Threshold overflow")),
+            );
+            let bsp_assignment_threshold =
+                match asymptotic_decay_fn.checked_add(&T::AssignmentThresholdAsymptote::get()) {
+                    Some(threshold) => threshold,
+                    None => panic!("Threshold overflow"),
+                };
+
+            Self {
+                bsp_assignment_threshold: bsp_assignment_threshold,
```
ffarall, line 251:
```suggestion
                bsp_assignment_threshold,
```

rust-toolchain
```diff
@@ -1,5 +1,4 @@
 [toolchain]
-channel = "1.77.0"
```
ffarall, line 2: Why deleting this? I don't mind upgrading, but for tracking the versions used, I think this should be here. And in any case, if we're taking it out, it shouldn't be in this PR.

support/traits/src/lib.rs
```diff
@@ -47,6 +60,98 @@ pub trait ProvidersInterface {
     fn get_stake(
         who: Self::Provider,
     ) -> Option<<Self::Balance as fungible::Inspect<Self::AccountId>>::Balance>;
+
+    /// Get number of registered BSPs.
+    fn get_number_of_bsps() -> Self::SpCount;
```
ffarall, line 65: This should't be here. This is the interface used by the Proofs Dealer pallet, and is therefore abstracted over the concept of BSPs and MSPs, it is even unaware of the concept of "Storage" Providers, it just deals with providers. This should be in another interface, separate.

pallets/file-system/src/lib.rs
```diff
@@ -449,4 +562,40 @@ pub mod pallet {
...
+
+    impl<T: Config> storage_hub_traits::SubscribeProvidersInterface for Pallet<T> {
+        type Provider = StorageProviderId<T>;
+
+        fn subscribe_bsp_sign_up(_who: &Self::Provider) {
+            // Adjust bsp assignment threshold by applying the decay function after removing the asymptote
+            let mut bsp_assignment_threshold = BspsAssignmentThreshold::<T>::get();
+            let base_threshold = bsp_assignment_threshold
+                .checked_sub(&T::AssignmentThresholdAsymptote::get())
+                .unwrap_or_else(|| panic!("Underflow in threshold calculation during sign-up"));
```
ffarall, line 574: We should avoid this `panic`s in the runtime. Use `expect_or_else` instead. Besides, using `unwrap_or_else` and then panicking is kinda redundant. It is equivalent to using just `expect()`

pallets/providers/src/utils.rs
```diff
@@ -486,4 +512,8 @@ impl<T: pallet::Config> ProvidersInterface for pallet::Pallet<T> {
             None
         }
     }
+
```
ffarall, line 515: This here again, shouldn't be here

runtime/src/lib.rs
```diff
@@ -544,11 +548,25 @@ impl TrieVerifier for ProofTrieVerifier {
     }
 }
 
+type ThresholdType = i128;
+
+pub struct ThresholdDecayFunction;
+impl Get<ThresholdType> for ThresholdDecayFunction {
+    fn get() -> ThresholdType {
```
ffarall, line 555: Wouldn't this yield `0`? We're returning an integer division whose result is smaller than `1`

runtime/src/lib.rs
```diff
@@ -544,11 +548,25 @@ impl TrieVerifier for ProofTrieVerifier {
...
+}
+
 /// Configure the pallet template in pallets/template.
 impl pallet_file_system::Config for Runtime {
     type RuntimeEvent = RuntimeEvent;
+    type Providers = Providers;
+    type ProofDealer = ProofsDealer;
+    type ThresholdType = ThresholdType;
+    type AssignmentThresholdDecayFunction = ThresholdDecayFunction;
+    type AssignmentThresholdAsymptote = ConstI128<100i128>;
```
ffarall, line 567: Why `i128` instead of `u128`? This shouldn't be negative

ffarall (changes requested):

pallets/providers/src/mock.rs
```diff
@@ -69,7 +69,7 @@ impl pallet_balances::Config for Test {
     type RuntimeHoldReason = RuntimeHoldReason;
     type RuntimeFreezeReason = ();
     type FreezeIdentifier = ();
-    type MaxHolds = ConstU32<10>;
+    type MaxHolds = ConstU32<100>;
```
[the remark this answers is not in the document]
snowmead: I can't recall precisely, but the substrate runtime was panicking due to this number being too low since the providers pallet introduced new hold reasons.

pallets/file-system/src/types.rs
```diff
@@ -32,7 +32,7 @@ pub struct StorageRequestMetadata<T: Config> {
     ///
     /// This is useful when a BSP stops serving data and automatically creates a new storage request with no user multiaddresses, since
     /// SPs can prove and serve the data to be replicated to other BSPs without the user having this stored on their local machine.
-    pub data_server_sps: BoundedVec<StorageProviderId<T>, MaxBspsPerStorageRequest<T>>, // TODO: Change the Maximum data servers to be the maximum SPs allowed
+    pub data_server_sps: BoundedVec<T::AccountId, MaxBspsPerStorageRequest<T>>, // TODO: Change the Maximum data servers to be the maximum SPs allowed
```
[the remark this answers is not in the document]
snowmead: Ya it is confusing. I removed `StorageProviderId` and use `T::AccountId` instead.

pallets/file-system/src/utils.rs
snowmead: ya, caught this as well. It is not `.try_into()` from block number to `ThresholdType`

pallets/file-system/src/utils.rs
snowmead: sweet!

pallets/file-system/src/utils.rs
snowmead: true!

pallets/file-system/src/utils.rs
snowmead: yes!

rust-toolchain
snowmead: Yes, agreed. Unintentional.

support/traits/src/lib.rs
snowmead: If the entire trait is based on the perspective of the Proofs Dealer pallet, then there should be a dedicated trait for it. 

There is explicitely a `Read` and `Mutate` trait that the Providers pallet implements which any pallet like File System and Proofs Dealer can inherit, but the File System isn't agnostic over storage providers. It is directly tied to the Storage Hub ecosystem is not generic over any chain like Proofs Dealer pallet. 

We should create a trait which the Proofs Dealer pallet expects with all the necessary functions it needs to accomplish it's tasks (issuing challenges, generic over ACCOUNTS).

pallets/file-system/src/lib.rs
snowmead: Return `DispatchResult` now with `ThresholdArithmeticError`.

runtime/src/lib.rs
snowmead: I am not sure I understand how it would yield `0`... Why can't it return `0.5`?

It's 128-bit **signed** integer type.

pallets/providers/src/mock.rs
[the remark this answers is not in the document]
ffarall: Ok cool

support/traits/src/lib.rs
ffarall: Exactly, this trait was specified by the Proofs Dealer pallet, that is why it was called `ProvidersInterface`. It is exactly what you're describing. In any case, we can define another `ReadProvidersInterface` that is not abstracted.

rust-toolchain
ffarall: I'd just avoid upgrading the toolchain in this repo altogether. To keep the scope clear.

pallets/file-system/src/utils.rs
```diff
@@ -237,13 +361,13 @@ where
             Some(mut metadata) => {
                 // Remove BSP from storage request and challenge if has confirmed having stored this file.
                 if let Some(bsp) = <StorageRequestBsps<T>>::get(&location, &who) {
-                    // TODO: challenge the BSP to force update its storage
-
                     if bsp.confirmed {
```
ffarall, line 364: We should handle the case where a BSP is NOT confirmed, but still in the storage request as volunteer. This is not an expected behaviour, as it would mean that a BSP volunteered but it is trying to remove itself as BSP, before it even provided a proof of including that file in its Merkle Patricia Forest (i.e. confirm). We can handle it with an `expect_or_err` or similar.

pallets/file-system/src/utils.rs
```diff
@@ -294,4 +416,39 @@ where
         );
         block_to_insert_expiration
     }
+
+    /// Challenge `file_key` to force the BSP to update its storage root to uninclude the
+    /// file and reduce the data used by the storage provider.
+    fn challenge_and_reduce_data_used(
```
ffarall, line 422: There probably isn't much need for this function since we shouldn't challenge when a BSP stops storing a file.

pallets/file-system/src/utils.rs
```diff
@@ -203,7 +324,10 @@ where
         // Remove storage request.
         <StorageRequests<T>>::remove(&location);
 
-        // TODO: initiate deletion request for SPs.
+        // Challenge BSP to force update its storage root to uninclude the file.
```
ffarall, line 327:
```suggestion
        // Challenge BSPs who confirmed this storage request, to force update its storage root to uninclude the file.
```

ffarall: We should check first if this storage request has confirmed BSPs before going through the trouble of sending the priority challenge.

pallets/file-system/src/utils.rs
```diff
@@ -176,10 +203,104 @@ where
...
+            !<StorageRequestBsps<T>>::get(&location, &who)
+                .expect("BSP should exist since we checked it above")
+                .confirmed,
+            Error::<T>::BspAlreadyConfirmed
+        );
+
+        let mut file_metadata =
+            <StorageRequests<T>>::get(&location).ok_or(Error::<T>::StorageRequestNotFound)?;
+
+        // Check that the number of confirmed bsps is less than the required bsps and increment it.
```
ffarall, line 250: This is probably an old comment, as you're not actually checking that the confirmed BSPs is less than the required. In fact, it is correct that you're not checking it, because it shouldn't be possible for the confirmed BSPs to be greater or equal than the required. In that case, there shouldn't even be a storage request anymore.

ffarall: I support that you check it anyway, but with `expect_or_err`, as it is not an expected state of the chain.

ffarall: What you are checking is an overflow in the `checked_add`, so the error should be an overflow, not `StorageRequestBspsRequiredFulfilled`.

pallets/file-system/src/utils.rs
```diff
@@ -176,10 +203,104 @@ where
...
+        // Remove storage request if we reached the required number of bsps.
+        // TODO this should be modified to also take into account if the MSP confirmed storing the file once we add MSP functionality.
+        if file_metadata.bsps_confirmed == file_metadata.bsps_required {
+            // Remove storage request metadata.
+            <StorageRequests<T>>::remove(&location);
+
+            const REMOVE_LIMIT: u32 = u32::MAX;
+
+            // Remove storage request bsps
+            let _ = <StorageRequestBsps<T>>::clear_prefix(&location, REMOVE_LIMIT, None);
```
ffarall, line 277: We should handle this properly, and set a proper `REMOVE_LIMIT`, to avoid attacks because this function takes too long. `REMOVE_LIMIT` should be equal to the number of BSPs who volunteered for this file. We can keep a counter in the storage request for that if needed. And after setting the `REMOVE_LIMIT` to that value, if the returned value of `clear_prefix` is something other than `None`, then that's a big unexpected error. As it would mean that there is an inconsistency between the BSPs accounted for as volunteers, and the number of keys in `StorageRequestBsps` for that `location`

pallets/file-system/src/mock.rs
```diff
@@ -148,24 +154,61 @@ impl TrieVerifier for MockVerifier {
     }
 }
 
+type ThresholdType = u128;
+
+pub struct ThresholdDecayFunction;
+impl Get<ThresholdType> for ThresholdDecayFunction {
+    fn get() -> ThresholdType {
+        0
```
ffarall, line 162: What's up here? Shouldn't it be `2` or something like that?

ffarall (changes requested):

pallets/file-system/src/lib.rs
```diff
@@ -188,6 +218,48 @@ pub mod pallet {
...
+    #[pallet::storage]
+    #[pallet::getter(fn bsps_assignment_threshold)]
+    pub type BspsAssignmentThreshold<T: Config> = StorageValue<_, T::ThresholdType, ValueQuery>;
+
+    #[pallet::genesis_config]
+    pub struct GenesisConfig<T: Config> {
+        pub bsp_assignment_threshold: T::ThresholdType,
+    }
+
+    impl<T: Config> Default for GenesisConfig<T> {
```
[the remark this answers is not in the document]
snowmead: Genesis configs are normally in the pallet module. If you insist, I will move it to utils.

pallets/file-system/src/utils.rs
```diff
@@ -176,10 +203,104 @@ where
...
+        <T::Providers as storage_hub_traits::MutateProvidersInterface>::increase_data_used(
+            &who,
+            file_metadata.size,
+        )?;
+
+        Ok(())
+    }
+
     /// Revoke a storage request.
     pub(crate) fn do_revoke_storage_request(
```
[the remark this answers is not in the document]
snowmead: Anyone can create a storage request (i.e. user, msp and bsp). I think we should only charge a fee for revoking a storage request based on the number of proofs (confirmations) submitted by storage providers.

pallets/file-system/src/utils.rs
```diff
@@ -237,13 +361,13 @@ where
             Some(mut metadata) => {
                 // Remove BSP from storage request and challenge if has confirmed having stored this file.
                 if let Some(bsp) = <StorageRequestBsps<T>>::get(&location, &who) {
```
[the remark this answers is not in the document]
snowmead: Also the `bsps_required` in the storage request metadata is static number that we try to reach with `bsps_confirmed`, which simply tracks the number of BSPs that confirmed storing the data.

pallets/file-system/src/utils.rs
```diff
@@ -237,13 +361,13 @@ where
...
-                    // TODO: challenge the BSP to force update its storage
-
                     if bsp.confirmed {
                         metadata.bsps_confirmed =
                             metadata.bsps_confirmed.saturating_sub(1u32.into());
 
                         <StorageRequests<T>>::set(&location, Some(metadata));
+
+                        Self::challenge_and_reduce_data_used(&who, &file_key, size)?;
```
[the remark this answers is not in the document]
snowmead: Added a TODO for this.

pallets/file-system/src/utils.rs
snowmead: We already handle this by removing the BSP from the storage request so they are no longer considered to be a volunteer.

snowmead: We will only verify the proof and update the root of the provider if they are confirmed to be storing the file of a storage request or when there is no storage request at all.

pallets/file-system/src/utils.rs
snowmead: Oh! Thanks for spotting this. 🙏🏼

support/traits/src/lib.rs
snowmead: Ok perfect, to also avoid redundancy for common functions between both, `ReadProvidersInterface` inherits `ProvidersInterface`

pallets/file-system/src/utils.rs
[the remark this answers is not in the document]
ffarall: My mistake, at the time I was wrongly interpreting that this would be used by a BSP or MSP to cancel their sign up to a storage request.

ffarall (approved):

PR #25, opened 2024-03-26
feat(CI): :rocket: Integrated DockerHub to CI workflow

timbrinded:
This PR contains:
- Logic to skip building images when `skip-node-build` is set
- Tests to run against "latest" (which is tip of main) if build is skipped
- Images now pushed to docker hub
- Repo Rust toolchain file
- Updated Docs
- Some new scripts to launch networks

> [!IMPORTANT]  
> The zombienet tests don't actually run against storage-hub image since it seems our runtime spec is incompatible. Will revisit once our runtime is more developed (i.e. when we build `local_testnet` properly)


.github/workflows/parachain.yml
```diff
@@ -0,0 +1,158 @@
+# Storage Hub: Parachain CI/CD Workflow
+#
+# Overview:
+# 1. Setup: This attempts to determine two things:
+#     - If the parachain node needs to be rebuilt based on changes to src code,
+#       or if an explicit skip tag has been set in the PR.
+#     - The tag name to use for the Docker image based on the outcome of the first check.
+# 2. Build Image: Conditionally executes based on the outcome of the setup job. If a new
+#    build is determined nessecary, it proceeds to build a Docker image of the parachain node.
```
ffarall, line 9:
```suggestion
#    build is determined necessary, it proceeds to build a Docker image of the parachain node.
```

.github/workflows/parachain.yml
```diff
@@ -0,0 +1,158 @@
...
+# 3. Zombie Test: Executes integration tests using Zombienet. This job pulls the relevant Docker image
+#    based on the setup job's output and runs tests to ensure network functionality.
+#
+# Note: This workflow assumes the presence of repo secrets for Docker Hub authentication
+#
+# TODO:
+# - Add parachain test suite (when ready)
+# - Publish runtime to docker
```
ffarall, line 17: Isn't it publishing already?

ffarall (approved):
LGTM. Approved with minor comments.

.github/workflows/parachain.yml
timbrinded: Implicitly, yes, but i mean more like what moonbeam and other parachains do where they have a separate CI job to publish a node at a particular commit for a runtime release.

PR #33, opened 2024-04-05
feat: ✨ add randomness to SPs' IDs when signing up

TDemeco:
To have randomness in the SP ID generation we are introducing a new two-step sign up process: first, the user has to request to sign up as a SP (MSP or BSP) and afterwards, when the randomness requested is valid, it has to confirm the sign up.

This PR:
- Revamps the sign up extrinsics of the `providers` pallet to use this new process by changing the `xsp_sign_up` extrinsics to `request_xsp_sign_up` (which now creates a request to sign up instead of directly doing it) and adding the `confirm_sign_up` and `cancel_sign_up` extrinsics that allow a user to confirm its registration as a SP or cancel it respectively
- Adds a new type called `StorageProvider` which is an enum that can either be the metadata of a Main Storage Provider or of a Backup Storage Provider
- Adds a new storage map called `SignUpRequests` that is a map from an AccountId to a tuple of `(block number in which the request was made, type and metadata of the SP to register)` using the aforementioned new type, and a getter for that map
- Updates the mocked runtime of the `providers` pallet to include the new needed configuration parameters
- Adds a `Randomness` mock in the `providers` pallet's mocked runtime to allow for consistent testing
- Adds a TON of tests of the new sign up process, and revamps all the old ones to use this new logic

TODOs before merging:
- [x] Merge latest changes from main into this branch
- [x] Organice the testing suite by success/fail and MSP/BSP for more clarity
- [x] Update the mock runtimes of the pallets that include the `providers` pallet so they work
- [x] ~~Update the actual runtime with `pallet-babe` and `RandomnessFromOneEpochAgo` to use that randomness provider for the `providers` pallet~~ -> this task was replaced with the one below for this PR, will be done in another PR
- [x] Implement the mocked Randomness in the runtime for testing


snowmead (changes requested):
Few comments 👍🏼

snowmead:
In various places, the mutation closures contain a `let new_total_bsp_capacity = n.checked_sub(&bsp.capacity);` for example, which is unecessary. You can simply do `match n.checked_add(&bsp_info.capacity)`

```rust
TotalBspsCapacity::<T>::mutate(|n| {
            match n.checked_add(&bsp_info.capacity) {
                Some(new_total_bsp_capacity) => {
                    *n = new_total_bsp_capacity;
                    Ok(())
                }
                None => Err(DispatchError::Arithmetic(ArithmeticError::Overflow)),
            }
        })?;
```

pallets/providers/src/utils.rs
```diff
@@ -49,10 +50,22 @@ impl<T> Pallet<T>
...
 {
-    pub fn do_msp_sign_up(who: &T::AccountId, msp_info: &MainStorageProvider<T>) -> DispatchResult {
+    /// This function tests the randomness of the pallet. TODO: remove this, only for testing
+    pub fn test_randomness_output(who: &T::AccountId) -> (T::Hash, BlockNumberFor<T>) {
+        T::ProvidersRandomness::random(who.encode().as_ref())
+    }
+
+    /// This function holds the logic that checks if a user can request to sign up as a Main Storage Provider
+    /// and, if so, stores the request in the SignUpRequests mapping
+    pub fn do_request_msp_sign_up(
```
snowmead, line 60: You should check if the account does not already have a pending sign up request in `SignUpRequests` storage.

pallets/providers/src/utils.rs
```diff
@@ -116,35 +129,29 @@ where
...
 
-        // Increment the counter of Main Storage Providers registered
-        MspCount::<T>::set(new_amount_of_msps);
         Ok(())
     }
 
-    pub fn do_bsp_sign_up(
+    /// This function holds the logic that checks if a user can request to sign up as a Backup Storage Provider
+    /// and, if so, stores the request in the SignUpRequests mapping
+    pub fn do_request_bsp_sign_up(
```
snowmead, line 146: Same here ^

snowmead (changes requested):

pallets/providers/src/lib.rs
```diff
@@ -400,6 +436,61 @@ pub mod pallet {
...
+        ///
+        /// Notes:
+        /// - This extrinsic could be called by the user itself or by a third party
+        /// - The deposit that the user has to pay to register as a SP is held when the user requests to register as a SP
+        /// - If this extrinsic is successful, it will be free for the caller, to incentive state debloating
+        #[pallet::call_index(2)]
+        #[pallet::weight(Weight::from_parts(10_000, 0) + T::DbWeight::get().writes(1))]
+        pub fn confirm_sign_up(
+            origin: OriginFor<T>,
+            who: T::AccountId,
```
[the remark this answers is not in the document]
TDemeco: I wanted to allow third-parties to confirm sign ups for other accounts, just for more flexibility

pallets/providers/src/utils.rs
```diff
@@ -208,16 +215,156 @@ where
...
+        );
+
+        // Check that the current block number is not greater than the block number when the request was made plus the maximum amount of
+        // blocks that we allow the user to wait for valid randomness (should be at least more than an epoch if using BABE's RandomnessFromOneEpochAgo)
+        // We do this to ensure that a user cannot wait indefinitely for randomness that suits them
+        ensure!(
+            frame_system::Pallet::<T>::block_number()
+                < request_block + T::MaxBlocksForRandomness::get(),
+            Error::<T>::SignUpRequestExpired
+        );
```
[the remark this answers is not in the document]
TDemeco: Yep! Good catch, this logic was originally dependent on the block number that the randomness returned so it had to be down here, I'll move it

pallets/providers/src/utils.rs
TDemeco: My idea was that if a user has a pending sign up request and is trying to request another sign up, we just overwrite the previous one with the new one, but now that I'm thinking we would have to add the logic to check for it and change the held amount, so I think your solution is more elegant and simple. 👍🏻

pallets/providers/src/utils.rs
```diff
@@ -208,16 +215,156 @@ where
...
+        MainStorageProviders::<T>::insert(&msp_id, msp_info);
+
+        // Increment the counter of Main Storage Providers registered
+        MspCount::<T>::set(new_amount_of_msps);
+
+        // Remove the sign up request from the SignUpRequests mapping
+        SignUpRequests::<T>::remove(who);
+
+        // Emit the corresponding event
+        Self::deposit_event(Event::<T>::MspSignUpSuccess {
```
ffarall, line 321: In the convention we're using, events are emitted in the `lib.rs` file. I'm not against emitting them here though, but we should have a discussion about it and come to an agreement.

pallets/providers/src/utils.rs
```diff
@@ -239,6 +386,17 @@ where
 
         // Increment the counter of Backup Storage Providers registered
         BspCount::<T>::set(new_amount_of_bsps);
+
+        // Remove the sign up request from the SignUpRequests mapping
+        SignUpRequests::<T>::remove(who);
+
+        // Emit the corresponding event
+        Self::deposit_event(Event::<T>::BspSignUpSuccess {
```
ffarall, line 394: Same with this event

pallets/providers/src/lib.rs
```diff
@@ -303,23 +343,22 @@ pub mod pallet {
...
-        /// 5. Check that the data to be stored is greater than the minimum required by the runtime. todo!("Ask if this applies to MSPs")
+        /// 5. Check that the data to be stored is greater than the minimum required by the runtime.
         /// 6. Calculate how much deposit will the signer have to pay using the amount of data it wants to store
         /// 7. Check that the signer has enough funds to pay the deposit
         /// 8. Hold the deposit from the signer
-        /// 9. Update the MSP storage to add the signer as an MSP
-        /// 10. Increment the storage that holds total amount of MSPs currently in the system
-        /// 11. Emit an event confirming that the registration of the MSP has been successful
+        /// 9. Update the Sign Up Requests storage to add the signer as requesting to sign up as a MSP
+        /// 10. Emit an event confirming that the sign up request as MSP has been successful
```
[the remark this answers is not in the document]
ffarall: Honestly I like and approve both types of docs. I believe both styles are pretty verbose, and I like it like that, as it explains a good deal of the extrinsic. But also both need maintaining as there are changes in the pallets. If you want us to agree on a single style, I'm all for it, but I reiterate that both are good standards of documentation.

pallets/providers/src/lib.rs
```diff
@@ -338,10 +377,10 @@ pub mod pallet {
             };
 
             // Sign up the new MSP (if possible), updating storage
-            Self::do_msp_sign_up(&who, &msp_info)?;
+            Self::do_request_msp_sign_up(&who, &msp_info)?;
 
             // Emit the corresponding event
-            Self::deposit_event(Event::<T>::MspSignUpSuccess {
+            Self::deposit_event(Event::<T>::MspRequestSignUpSuccess {
```
ffarall, line 383: This is a bit confusing. In this transaction you do emit the event here.

pallets/providers/src/utils.rs
ffarall: I see why you do this. You have one single extrinsic that can emit one or the other event. Maybe this is a good argument to emit events here...

pallets/providers/src/lib.rs
```diff
@@ -400,6 +436,61 @@ pub mod pallet {
             Ok(().into())
         }
 
+        /// A dispatchable function that allows users to confirm their sign up as a Storage Provider, either MSP or BSP.
+        ///
+        /// This extrinsic will:
+        /// 1. Check that the extrinsic was signed
+        /// 2. Check that the account received has requested to register as a MSP
```
ffarall, line 443:
```suggestion
        /// 2. Check that the account received has requested to register as a SP
```

pallets/providers/src/lib.rs
```diff
@@ -400,6 +436,61 @@ pub mod pallet {
...
+        /// 3. Check that by registering this SP we would not go over the MaxMsps or MaxBsps limit
+        /// 4. Check that the current randomness is sufficiently fresh to be used as a salt for that request
+        /// 5. Check that the request has not expired
+        /// 6. Register the signer as a MSP with the data provided in the request
```
ffarall, line 447:
```suggestion
        /// 6. Register the signer as a MSP or BSP with the data provided in the request
```

pallets/providers/src/lib.rs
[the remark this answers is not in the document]
ffarall: I support the flexibility, but I'd suggest renaming `origin` to `who`, and `who` to `provider_account`, making `provider_account` an `Option`. Then if `provider_account` is `None`, we use `who` (the signer).

pallets/providers/src/utils.rs
ffarall: This actually exposes a kind of vulnerability. We do not allow for a same account to be a BSP and MSP at the same time. But you can sign up for both while not being confirmed, and then confirm them, becoming both. All the more reason to not allow SPs to re-sign up when they have a request ongoing.

pallets/providers/src/tests.rs
```diff
@@ -23,1040 +27,2509 @@ type DepositPerData = <Test as crate::Config>::DepositPerData;
...
+                    // Advance enough blocks for randomness to be valid
+                    run_to_block(
+                        frame_system::Pallet::<Test>::block_number()
+                            + BLOCKS_BEFORE_RANDOMNESS_VALID,
+                    );
+
+                    // Confirm the sign up of the account as a Main Storage Provider
+                    let confirm_result =
+                        StorageProviders::confirm_sign_up(RuntimeOrigin::signed(alice), alice);
+                    assert_eq!(confirm_result, Ok(Pays::No.into()));
```
ffarall, line 1257: Could think of checking the balances before and after, although I don't know if mocked runtimes consume gas 🤔

pallets/providers/src/tests.rs
```diff
@@ -23,1040 +27,2509 @@ type DepositPerData = <Test as crate::Config>::DepositPerData;
...
-                StorageProviders::msp_sign_up(
-                    RuntimeOrigin::signed(alice),
-                    storage_amount,
-                    multiaddresses.clone(),
-                    value_prop.clone()
-                ),
-                Error::<Test>::StorageTooLow
-            );
+        /// This module holds the failure cases for functions that test both Main Storage Providers and Backup Storage Providers
+        mod msp_and_bsp {
```
ffarall, line 1931: After adding the check for if an AccountId already has a registration request, add a failure check for it, with an account trying to first sign up as BSP, and then MSP, or viceversa.

ffarall (commented):
Very good step by step documentation of what you were doing, and documenting functions behaviour.

Very good coverage of testing!

Overall great code, just some minor comments.

Please make sure the runtime compiles before merging.

pallets/providers/src/utils.rs
snowmead: Ya like @ffarall points out, it leads to many caveats.

pallets/providers/src/lib.rs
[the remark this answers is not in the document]
TDemeco: Yeah, that makes it more clear. I'll work on that change

pallets/providers/src/utils.rs
TDemeco: Mmmm not really, since the storage that holds the requests is a mapping from an account to a single request. That means that an account can only have one sign up request active, so the user wouldn't be able to register as both a BSP and a MSP.
Still, I agree that it complicates the logic quite a bit just to allow users to not have to make an extra call (to `cancel_sign_up`) so it's not really worth it

pallets/providers/src/utils.rs
```diff
@@ -208,16 +215,156 @@ where
...
+    pub fn do_cancel_sign_up(who: &T::AccountId) -> DispatchResult {
+        // Check that the signer has requested to sign up as a Storage Provider
+        SignUpRequests::<T>::get(who).ok_or(Error::<T>::SignUpNotRequested)?;
+
+        // Remove the sign up request from the SignUpRequests mapping
+        SignUpRequests::<T>::remove(who);
+
+        // Return the deposit to the signer
+        // We return all held funds as there's no possibility of the user having another _valid_ hold with this pallet
+        T::NativeBalance::release_all(
```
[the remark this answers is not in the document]
TDemeco: I thought it made more sense to use `release_all` as it is only releasing all the funds held by this pallet (and does not interfere with any other external holds). The alternative would be to:
a) Keep track of the funds held (which adds a storage field and some extra logic to keep track of it and keep it updated)
b) Calculate the held funds on the fly considering the registered `capacity` of the SP (which adds extra logic and calculations that could go wrong, plus if the price per storage changes it just doesn't work

My guess is you were thinking about option a), so if you think we should go that route lets talk tomorrow about it during the meeting 👍🏻

pallets/providers/src/utils.rs
TDemeco: Yep! Lets discuss it tomorrow, but that's exactly the reason why. I could also split it into two extrinsics and can then emit events where we usually do. It wouldn't be as user-friendly but it's not that bad either

pallets/providers/src/tests.rs
TDemeco: They don't! It was actually a problem I ran into when developing a free transactions pallet (where the runtime would sponsor transactions for users). I learned then that the easiest way to check if a user would have paid for an extrinsic execution is to check in the returned`DispatchResultWithPostInfo` the `pays` field and making sure it is `Pays::No`

santikaplan:
Resolves SHT-149

pallets/providers/src/utils.rs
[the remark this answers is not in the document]
ffarall: I think the use of `release_all` is fine here. Doesn't pose a risk that I can see at least, and reduces complexity.

pallets/providers/src/utils.rs
ffarall: Good point, still checking I think is safer.

pallets/providers/src/utils.rs
ffarall: Ok, I think it is fine in this case.

ffarall (approved):

snowmead (approved):
LGTM 👍🏼

PR #34, opened 2024-04-08
chore: add documentation and unit tests

snowmead:


pallets/file-system/src/tests.rs
```diff
@@ -303,6 +303,32 @@ fn revoke_request_storage_success() {
     });
 }
 
+#[test]
+fn revoke_storage_request_not_owner_fail() {
```
ffarall, line 307: Shouldn't there be a test case for a non-existing storage request as well?

pallets/file-system/src/utils.rs
```diff
@@ -136,14 +136,24 @@ where
         Ok(())
     }
 
+    /// Volunteer to store a file.
+    ///
+    /// *Callable only by BSP accounts*
+    ///
+    /// A BSP can only volunteer for a storage request if it is eligible absed on the XOR of the `fingerprint` and the BSP's account ID and if it evaluates to a value
```
ffarall, line 143: absed?

pallets/file-system/src/tests.rs
```diff
@@ -366,6 +392,140 @@ fn bsp_volunteer_success() {
...
+                fingerprint,
+                multiaddresses.clone()
+            ),
+            Error::<Test>::BspAlreadyVolunteered
+        );
+    });
+}
+
+#[test]
+fn bsp_volunteer_threshold_too_high_fail() {
```
ffarall, line 481: The name of the error `ThresholdTooHigh` is a bit confusing. How about `AboveThreshold`?

pallets/file-system/src/tests.rs
```diff
@@ -454,6 +614,212 @@ fn bsp_confirm_storing_success() {
...
+            FileSystem::bsp_volunteer(
+                bsp_signed.clone(),
+                location.clone(),
+                fingerprint,
+                multiaddresses.clone()
+            ),
+            Error::<Test>::NotABsp
+        );
+
+        assert_noop!(
```
ffarall, line 799: I don't see why you're doing this twice

pallets/file-system/src/tests.rs
```diff
@@ -579,48 +945,294 @@ fn bsp_stop_storing_success() {
...
+
+#[test]
+fn compute_asymptotic_threshold_point_success() {
+    new_test_ext().execute_with(|| {
+        // Test the computation of the asymptotic threshold
+        let threshold = FileSystem::compute_asymptotic_threshold_point(1)
+            .expect("Threshold should be computable");
+
+        // Assert that the computed threshold is as expected
+        assert!(
```
ffarall, line 1179: This could be more thorough in the check, and actually check the result for a known input of numbers.

pallets/file-system/src/tests.rs
```diff
@@ -579,48 +945,294 @@ fn bsp_stop_storing_success() {
...
+
+        // Simulate the threshold decrease due to a new BSP sign up
+        FileSystem::subscribe_bsp_sign_up(&AccountId32::new([2; 32]))
+            .expect("BSP sign up should be successful");
 
-        // Simulate the threshold decrease due to a new bsp sign off
-        assert_ok!(
-            <FileSystem as SubscribeProvidersInterface>::subscribe_bsp_sign_off(&bsp_account_id)
+        // Verify that the threshold decreased
+        assert!(
```
ffarall, line 1196: Here also, ideally we can check the actual numerical value of the change.

pallets/file-system/src/tests.rs
```diff
@@ -579,48 +945,294 @@ fn bsp_stop_storing_success() {
...
+        let initial_threshold = compute_set_get_initial_threshold();
+
+        // Simulate the threshold increase due to a new BSP sign off
+        FileSystem::subscribe_bsp_sign_off(&AccountId32::new([2; 32]))
+            .expect("BSP sign off should be successful");
 
-        // Verify that the threshold increased and should be equal to the initial threshold
-        assert_eq!(FileSystem::bsps_assignment_threshold(), initial_threshold);
+        // Verify that the threshold increased
+        assert!(
```
ffarall, line 1213: Here also, ideally we can check the actual numerical value of the change.

ffarall (changes requested):
Very good docs and clear testing.

Very good overall, just some minor comments.

pallets/file-system/src/utils.rs
```diff
@@ -271,9 +291,6 @@ where
             Error::<T>::BspAlreadyConfirmed
```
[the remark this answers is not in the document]
snowmead: I believe this is unecessary since it is impossible due to the check above. I can't use `expect_or_error` anyways, because it does not return the value of the execution. it is simply validating a condition is true.

pallets/file-system/src/tests.rs
snowmead: I don't want to compute the threshold in the unit test because I feel like it is redundant. 
I added a an extra condition though that checks that the threshold should be greater or equal than the asymptote.
`threshold >= <Test as Config>::AssignmentThresholdAsymptote::get(),`

ffarall (approved):

PR #36, opened 2024-04-09
feat: run as storage provider cli

snowmead:


node/src/command.rs
```diff
@@ -14,10 +14,21 @@ use storage_hub_runtime::Block;
...
     service::new_partial,
 };
 
+/// Configuration for the provider.
+#[derive(Debug, Clone)]
+pub struct ProviderOptions {
+    /// Provider type.
+    pub provider_type: ProviderType,
+    /// P2P port
+    pub p2p_port: u16,
```
links234, line 27: Is this actually passed anywhere? I'm asking because I thought substrate node is in control of the P2p swarm and hence the port too.

node/src/command.rs
```diff
@@ -14,10 +14,21 @@ use storage_hub_runtime::Block;
...
+    /// Seed to generate deterministic peer id.
+    pub secret_key_seed: String,
```
links234, line 29: Isn't this a bad combination with `#[derive(Debug)]`? 😜

links234 (changes requested):
Minor change requests.

Also please check why `zombienet` does not like the new changes.

node/src/command.rs
snowmead: true

node/src/command.rs
snowmead: You are right, will remove

pallets/providers/Cargo.toml
```diff
@@ -40,23 +40,23 @@ pallet-balances = { workspace = true, features = ["std"] }
 [features]
 default = ["std"]
 runtime-benchmarks = [
-    "frame-benchmarking/runtime-benchmarks",
```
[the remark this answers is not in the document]
snowmead: I am not sure if there is a method for formatting .toml files automatically. I added rustfmt at least.

links234 (approved):

PR #38, opened 2024-04-11
feat: :sparkles: add pallet `randomness` to provide randomness for the runtime

TDemeco:
This PR adapts the `randomness` pallet from [Moonkit](https://github.com/Moonsong-Labs/moonkit/tree/main/pallets/randomness) to provide the required randomness functionality for the StorageHub runtime


pallets/randomness/src/tests.rs
```diff
@@ -0,0 +1,64 @@
...
+#[test]
+fn set_babe_randomness_is_mandatory() {
+    use frame_support::dispatch::{DispatchClass, GetDispatchInfo};
+
+    let info = crate::Call::<Test>::set_babe_randomness {}.get_dispatch_info();
+    assert_eq!(info.class, DispatchClass::Mandatory);
+}
+
+#[test]
+fn set_babe_randomness_works() {
```
snowmead, line 18: It would be nice to have a test in which you have an epoch that passes with new randomness being set to validate passing from one epoch to another(e.g. `babe_randomness_epoch_change`).

runtime/src/configs/mod.rs
```diff
@@ -312,30 +313,55 @@ impl pallet_collator_selection::Config for Runtime {
...
+            let benchmarking_babe_output = Hash::default();
+            return Some(benchmarking_babe_output);
+        }
+        relay_chain_state_proof()
+            .read_optional_entry(well_known_keys::ONE_EPOCH_AGO_RANDOMNESS)
+            .ok()
+            .flatten()
     }
 }
+const EPOCH_DURATION_IN_BLOCKS: BlockNumber = 100;
```
snowmead, line 358: Isn't this supposed to be based on the current relay chain (i.e. kusama, polkadot, etc.)?

snowmead (approved):
Couple of comments 👍🏼

pallets/randomness/src/tests.rs
TDemeco: I was thinking of improving the mocked data getter to have a test like this, I'll work on it 👍🏻

runtime/src/configs/mod.rs
TDemeco: That's the idea, but after researching Cumulus extensively, I couldn't find a way to get the `EPOCH_DURATION_IN_BLOCKS` constant (`babe.epochDuration`) from the relay chain, plus that way it wouldn't be a constant. I think the best we could do would be to extract the constant from the relay chain repo somehow

pallets/randomness/README.md
```diff
@@ -0,0 +1,52 @@
+# Randomness
+
+This pallet provides access to randomness using as source the **BABE epoch randomness**, which is produced by the relay chain per relay chain epoch. It is based on **all the VRF produced** by the relay chain validators **during** a complete **epoch**.(~600 blocks on Kusama, ~2400 blocks on Polkadot). At the beginning of a new Epoch, those VRFs are **mixed together** and **hashed** in order to produce a **pseudo-random word**.
+
+## Babe Epoch Randomness
```
a-moreira, line 5: I really like the explanations in this file, easy to understand 😄 .

pallets/randomness/README.md
```diff
@@ -0,0 +1,52 @@
+# Randomness
+
+This pallet provides access to randomness using as source the **BABE epoch randomness**, which is produced by the relay chain per relay chain epoch. It is based on **all the VRF produced** by the relay chain validators **during** a complete **epoch**.(~600 blocks on Kusama, ~2400 blocks on Polkadot). At the beginning of a new Epoch, those VRFs are **mixed together** and **hashed** in order to produce a **pseudo-random word**.
```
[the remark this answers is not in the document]
a-moreira: Oh, I got confused since we do something similar of concatenating then hashing, sorry 😬

pallets/randomness/README.md
TDemeco: Thanks! 😃

pallets/randomness/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+                    log::warn!(
+                        "Failed to fill BABE epoch randomness \
+							REQUIRE HOTFIX TO FILL EPOCH RANDOMNESS FOR EPOCH {:?}",
+                        relay_epoch_index
+                    );
+                }
+            }
+
+            // Update storage with the latest epoch for which randomness was processed for
+            <RelayEpoch<T>>::put(relay_epoch_index);
```
ffarall, line 133: Does this need to happen at every block? Or just when there is a change?

pallets/randomness/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+
+        fn is_inherent(call: &Self::Call) -> bool {
+            matches!(call, Call::set_babe_randomness { .. })
+        }
+    }
+
+    #[pallet::hooks]
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        /// This hook checks, on block finalization, that the required inherent was included and clears
+        /// storage to make it necesary to include it in future blocks as well
```
ffarall, line 172:
```suggestion
        /// storage to make it necessary to include it in future blocks as well
```

pallets/randomness/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+        fn on_finalize(_now: BlockNumberFor<T>) {
```
ffarall, line 173: IIRC, the weight used by `on_finalize` had to be defined in `on_initialize`, so that the block producer can know if it will be able to produce a block. I know this literally only does one rad, and modification of the chain's state, but since there is no `on_initialize` here, how does it work?

pallets/randomness/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+				<InherentIncluded<T>>::take().is_some(),
+				"Mandatory randomness inherent not included; InherentIncluded storage item is empty"
+			);
+        }
+    }
+
+    // Randomness trait
+    impl<T: Config> frame_support::traits::Randomness<T::Hash, BlockNumberFor<T>> for Pallet<T> {
+        /// Uses the BABE randomness of this epoch to generate a random seed that can be used
+        /// for commitments from the last epoch. The provided `subject` MUST have been commited
```
ffarall, line 186:
```suggestion
        /// for commitments from the last epoch. The provided `subject` MUST have been committed
```

pallets/randomness/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+                digest.extend_from_slice(babe_randomness.as_ref());
+                digest.extend_from_slice(hashed_subject.as_ref());
+                // Hash it
+                let randomness = T::Hashing::hash(digest.as_slice());
+                // Return the randomness for this subject and the latest block for which this randomness is useful
+                // `subject` commitments done after `latest_valid_block` are predictable, and as such MUST be discarded
+                (randomness, latest_valid_block)
+            } else {
+                // If there's no randomness available, return an empty randomness that's invalid for every block
+                // TODO: Check edge case (randomness requested on the first block of the runtime for example)
```
ffarall, line 206: If by first block you mean the genesis block, I think it is safe to say that whoever deploys this chain, will be in control of that, and shouldn't shoot themselves in the foot.

primitives/session-keys/src/inherent.rs
```diff
@@ -0,0 +1,76 @@
+// Copyright 2019-2022 Moonsong Labs
+// This file is part of Moonkit.
```
ffarall, line 2: Again, not part of Moonkit, although we could give them credit. Otherwise, we could just import this from Moonkit, if we want to get updates from them.

ffarall (approved):
Approved with some minor comments. If you can clear those doubts for me, we're good to go.

Cargo.toml
```diff
@@ -102,15 +105,17 @@ pallet-balances = { git = "https://github.com/paritytech/polkadot-sdk.git", tag
...
+polkadot-primitives = { git = "https://github.com/paritytech/polkadot-sdk.git", tag = "polkadot-v1.9.0", default-features = false }
 pallet-xcm = { git = "https://github.com/paritytech/polkadot-sdk.git", tag = "polkadot-v1.9.0", default-features = false }
 polkadot-parachain-primitives = { git = "https://github.com/paritytech/polkadot-sdk.git", tag = "polkadot-v1.9.0", default-features = false }
 polkadot-runtime-common = { git = "https://github.com/paritytech/polkadot-sdk.git", tag = "polkadot-v1.9.0", default-features = false }
 xcm = { package = "staging-xcm", git = "https://github.com/paritytech/polkadot-sdk.git", tag = "polkadot-v1.9.0", default-features = false }
 xcm-builder = { package = "staging-xcm-builder", git = "https://github.com/paritytech/polkadot-sdk.git", tag = "polkadot-v1.9.0", default-features = false }
 xcm-executor = { package = "staging-xcm-executor", git = "https://github.com/paritytech/polkadot-sdk.git", tag = "polkadot-v1.9.0", default-features = false }
+#kusama-runtime-constants = { git = "https://github.com/polkadot-fellows/runtimes.git", tag = "v1.2.1", default-features = false }
```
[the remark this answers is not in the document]
TDemeco: Done!

pallets/randomness/src/lib.rs
TDemeco: Yes, forgot the `on_initialize` hook that returns the `on_finalize` weight. Fixed (although we still have to benchmark `on_finalize`)

primitives/session-keys/src/inherent.rs
TDemeco: Its not exactly their pallet, I made some updates but forgot to remove that text. We should give them credit somewhere still as the base of the randomness pallet comes from them

PR #39, opened 2024-04-16
feat: ✨ Implement first version of Blockchain Service listening to events and sending transactions

ffarall:


node/src/services/blockchain/handler.rs
```diff
@@ -72,19 +207,351 @@ impl ActorEventLoop<BlockchainService> for BlockchainServiceEventLoop {
...
+        let events_in_block = self.get_events_storage_element(block_hash)?;
+
+        // Filter the events for the extrinsic.
+        // Each event record is composed of the `phase`, `event` and `topics` fields.
+        // We are interested in those events whose `phase` is equal to `ApplyExtrinsic` with the index of the extrinsic.
+        // For more information see: https://polkadot.js.org/docs/api/cookbook/blocks/#how-do-i-map-extrinsics-to-their-events
+        let events = events_in_block
+            .into_iter()
+            .filter(|ev| ev.phase == frame_system::Phase::ApplyExtrinsic(extrinsic_index as u32))
+            .collect();
```
snowmead, line 480: Wow… what a hassle. 😅 

We should probably index all storage hub related transactions and easily fetch them later instead of doing all this work. We could use this method as a last resort if the transaction we are searching for was not found in our index.

node/src/services/blockchain/mod.rs
```diff
@@ -1,18 +1,32 @@
...
+use sc_service::RpcHandlers;
+use sp_keystore::KeystorePtr;
+use sp_runtime::KeyTypeId;
 use storage_hub_infra::actor::{ActorHandle, ActorSpawner, TaskSpawner};
 
+use crate::service::ParachainClient;
+
 use self::handler::BlockchainService;
 
+pub const KEY_TYPE: KeyTypeId = KeyTypeId(*b"bcsv");
```
snowmead, line 17: What about `bcsvc`? Normally “service" is represented as `svc`. If you hadn’t told me before, I wouldn’t have known what `sv` is.

node/src/services/blockchain/handler.rs
```diff
@@ -72,19 +207,351 @@ impl ActorEventLoop<BlockchainService> for BlockchainServiceEventLoop {
...
+
+    /// Get the events storage element in a block.
+    fn get_events_storage_element(&self, block_hash: H256) -> Result<EventsVec> {
+        // Would be cool to be able to do this...
+        // let events_storage_key = frame_system::Events::<storage_hub_runtime::Runtime>::hashed_key();
+
+        // Get the events storage key.
+        let events_storage_key = [
+            Twox128::hash(b"System").to_vec(),
+            Twox128::hash(b"Events").to_vec(),
```
snowmead, line 537: Do you think we can turn these hashes into consts instead since it is highly improbably that they will ever change?

node/src/tasks/bsp_volunteer_mock.rs
```diff
@@ -0,0 +1,118 @@
...
+            .blockchain
+            .send_extrinsic(call)
+            .await;
+
+        // Wait for the transaction to be included in a block.
+        let mut block_hash = None;
+        // TODO: Consider adding a timeout.
+        while let Some(tx_result) = tx_watcher.recv().await {
+            // Checking if there is an update with an error in our transaction.
+            if tx_result.starts_with("Error") {
```
snowmead, line 61: Should we change the receiver channel to have a `Result` instead and automatically return `Err` variant if “Error” is in the transaction result since it will be done for all tasks and probably be repeated?

node/src/tasks/bsp_volunteer_mock.rs
```diff
@@ -0,0 +1,118 @@
...
+                    .blockchain
+                    .unwatch_extrinsic(subscription_id.to_owned())
+                    .await?;
+
+                // Breaking while loop.
+                // Even though we unwatch the transaction, and the loop should break, we still break manually
+                // in case we continue to receive updates. This should not happen, but it is a safety measure,
+                // and we already have what we need.
+                break;
+            }
```
snowmead, line 91: We could probably externalize this into a common function so all tasks can call this.

node/src/tasks/bsp_volunteer_mock.rs
```diff
@@ -0,0 +1,118 @@
...
+            .storage_hub_handler
+            .blockchain
+            .get_extrinsic_from_block(block_hash, tx_hash)
+            .await?;
+
+        // Check if the extrinsic was successful. In this mocked task we know this should fail if Alice is
+        // not a registered BSP.
+        let extrinsic_successful = ActorHandle::<BlockchainService>::extrinsic_successful(extrinsic_in_block.clone())
+            .expect("Extrinsic does not contain an ExtrinsicFailed nor ExtrinsicSuccess event, which is not possible; qed");
+        if !extrinsic_successful {
```
snowmead, line 108: Regarding my previous comment, we could have access to `dispatch_err` and `dispatch_info` if we add it in the return type of `extrinsic_successful`. It would be useful.

snowmead (changes requested):
Overall it’s really good! Really nice job! 🔥

node/src/service.rs
```diff
@@ -299,6 +302,41 @@ async fn start_dev_impl(
...
+            &task_spawner,
+            client.clone(),
+            Arc::new(rpc_handlers),
+            keystore.clone(),
+        )
+        .await;
+
+        let file_transfer_service_handle = file_transfer_service_handle.expect(
+            "File Transfer Service handle is expected to be present when the node is running as a Storage Provider. qed",
+        );
```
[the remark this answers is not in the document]
ffarall: Good call

node/src/service.rs
```diff
@@ -299,6 +302,41 @@ async fn start_dev_impl(
...
 
+    // Spawning the Blockchain Service if node is running as a Storage Provider.
+    if let Some(provider_options) = provider_options {
+        let task_spawner = TaskSpawner::new(task_manager.spawn_handle(), "generic");
+
+        // Initialise seed for signing transactions using blockchain service.
+        // In dev mode we use a well known dev account.
+        keystore
+            .sr25519_generate_new(KEY_TYPE, Some("//Alice"))
+            .expect("Creating key with account Alice should succeed.");
```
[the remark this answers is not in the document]
ffarall: Good suggestion, done.

node/src/services/blockchain/commands.rs
```diff
@@ -0,0 +1,116 @@
...
+use storage_hub_infra::actor::ActorHandle;
+
+use super::{handler::BlockchainService, types::Extrinsic};
+
+/// Commands that can be sent to the BlockchainService actor.
+#[derive(Debug)]
+pub enum BlockchainServiceCommand {
+    SendExtrinsic {
+        call: storage_hub_runtime::RuntimeCall,
+        callback: tokio::sync::oneshot::Sender<(tokio::sync::mpsc::Receiver<String>, H256)>,
```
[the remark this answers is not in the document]
ffarall: In simple terms, the Sender sends to the overlaying task, the Receiver that it gets from the RPC call, so that the overlaying task can get updates from the RPC receiver.

node/src/services/blockchain/commands.rs
```diff
@@ -0,0 +1,116 @@
...
+        self.send(message).await;
+        rx.await.expect("Failed to receive response from BlockchainService. Probably means BlockchainService has crashed.")
+    }
+
+    fn extrinsic_successful(extrinsic: Extrinsic) -> Result<bool> {
+        for ev in extrinsic.events {
+            match ev.event {
+                storage_hub_runtime::RuntimeEvent::System(
+                    frame_system::Event::ExtrinsicFailed {
+                        dispatch_error: _,
```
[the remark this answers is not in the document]
ffarall: Good call

node/src/tasks/bsp_volunteer_mock.rs
ffarall: If you see the logs now, you'll find the error, although since it is encoded, it won't be of much use as it is. But it's there.

node/src/services/blockchain/events.rs
```diff
@@ -8,14 +16,22 @@ pub struct ChallengeRequest {
...
-    pub who: String,
-    pub location: String,
-    pub fingerprint: String,
-    pub size: u64,
-    pub multiaddresses: Vec<String>,
+    pub who: AccountId32,
+    pub location: FileLocation,
+    pub fingerprint: H256,
+    pub size: StorageData,
+    pub multiaddresses: BoundedVec<MultiAddress, MaxDataServerMultiAddresses>,
```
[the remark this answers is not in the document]
ffarall: Yeah, I thought the same thing, but didn't want to keep increasing the scope of this PR. I'll write a task in our board to remember to make this change.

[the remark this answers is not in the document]
ffarall: Actually it is already mapped [here](https://www.notion.so/moonsonglabs/Change-storage-request-to-receive-peer-ID-instead-of-multiaddresses-of-user-14a0a721dd76484fba0f3d6ebab4f791?pvs=4)

node/src/services/blockchain/events.rs
```diff
@@ -8,14 +16,22 @@ pub struct ChallengeRequest {
...
-// TODO: use proper types
+/// New storage request event.
+///
+/// This event is emitted when a new storage request is created on-chain.
+/// It contains the following information:
+/// - `who`: the account ID of the requester.
+/// - `location`: the location of the file (as a file path).
+/// - `fingerprint`: the fingerprint of the file (root hash of the merklised file).
+/// - `size`: the size of the file.
+/// - `multiaddresses`: the multiaddresses from where the user would send the file.
```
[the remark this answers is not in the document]
ffarall: Good call, will do

node/src/services/blockchain/handler.rs
```diff
@@ -16,38 +16,163 @@
...
+                                Err(e) => {
+                                    error!(target: LOG_TARGET, "Failed to send receiver: {:?}", e);
+                                }
+                            }
+                        }
+                        Err(e) => {
+                            warn!(target: LOG_TARGET, "Failed to send extrinsic: {:?}", e);
+
+                            // Send the error to the callback.
+                            let (tx, rx) = tokio::sync::mpsc::channel(1);
```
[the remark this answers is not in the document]
ffarall: Good call, will do

node/src/tasks/bsp_volunteer_mock.rs
ffarall: Yes, good call, will do

ffarall: Good call, will do

node/src/services/blockchain/handler.rs
ffarall: Good suggestion, done

node/src/services/blockchain/handler.rs
ffarall: We will eventually do that in the indexer, maybe by then we can optimise this to use that, but I guess for now I'd keep it like this.

node/src/services/blockchain/handler.rs
```diff
@@ -72,19 +207,351 @@ impl ActorEventLoop<BlockchainService> for BlockchainServiceEventLoop {
...
+        let extra: SignedExtra = (
+        frame_system::CheckNonZeroSender::<storage_hub_runtime::Runtime>::new(),
+        frame_system::CheckSpecVersion::<storage_hub_runtime::Runtime>::new(),
+        frame_system::CheckTxVersion::<storage_hub_runtime::Runtime>::new(),
+        frame_system::CheckGenesis::<storage_hub_runtime::Runtime>::new(),
+        frame_system::CheckEra::<storage_hub_runtime::Runtime>::from(generic::Era::mortal(
+            period,
+            current_block,
+        )),
+        frame_system::CheckNonce::<storage_hub_runtime::Runtime>::from(nonce),
```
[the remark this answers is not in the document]
ffarall: Yes, absolutely true. I'll leave a TODO for that along tipping, and a task for that.

node/src/services/blockchain/handler.rs
```diff
@@ -72,19 +207,351 @@ impl ActorEventLoop<BlockchainService> for BlockchainServiceEventLoop {
...
+                            fingerprint,
+                            size,
+                            multiaddresses,
+                        });
+                    }
+                    // Ignore all other events.
+                    _ => {}
+                }
+            }
+        }
```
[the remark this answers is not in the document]
ffarall: Sounds good

node/src/services/blockchain/handler.rs
```diff
@@ -16,38 +16,163 @@
...
+                            }
+
+                            match callback.send((rx, H256::zero())) {
+                                Ok(_) => {
+                                    trace!(target: LOG_TARGET, "Receiver sent successfully");
+                                }
+                                Err(e) => {
+                                    error!(target: LOG_TARGET, "Failed to send error message through channel: {:?}", e);
+                                }
+                            }
```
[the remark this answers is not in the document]
ffarall: It was because the type the callback expected had to be a `Receiver`. But now it uses `Result`

node/src/services/blockchain/handler.rs
```diff
@@ -16,38 +16,163 @@
...
-#[derive(Debug)]
-pub enum BlockchainServiceCommand {}
+const LOG_TARGET: &str = "blockchain-service";
 
-#[derive(Debug)]
+/// The BlockchainService actor.
+///
+/// This actor is responsible for sending extrinsics to the runtime and handling block import notifications.
+/// For such purposes, it uses the [`ParachainClient`] to interact with the runtime, the [`RpcHandlers`] to send
+/// extrinsics, and the [`Keystore`] to sign the extrinsics.
```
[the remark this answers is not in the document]
ffarall: Yes. Done.

node/src/services/blockchain/mod.rs
ffarall: Good suggestion, but `KeyTypeId`s are always 4 chars in length.

node/src/tasks/bsp_volunteer_mock.rs
ffarall: Yeah, could be improved, but I agree it is better in another PR.

node/src/services/blockchain/mod.rs
snowmead: Ah I see, we can change it later if we come up with something better.

node/src/services/blockchain/handler.rs
```diff
@@ -94,13 +110,7 @@ impl Actor for BlockchainService {
...
-                            // Send the error to the callback.
-                            let (tx, rx) = tokio::sync::mpsc::channel(1);
-                            if tx.send(e.to_string()).await.is_err() {
-                                error!(target: LOG_TARGET, "Failed to send error message through channel");
-                            }
-
-                            match callback.send((rx, H256::zero())) {
+                            match callback.send(Err(e)) {
                                 Ok(_) => {
                                     trace!(target: LOG_TARGET, "Receiver sent successfully");
```
links234, line 120: Should update this trace message too.

links234 (approved):

node/src/service.rs
```diff
@@ -225,15 +231,16 @@ async fn start_dev_impl(
...
+    // This is done here because the File Transfer Service modifies the network configuration.
+    let mut file_transfer_service_handle = None;
+    if provider_options.is_some() {
+        let task_spawner = TaskSpawner::new(task_manager.spawn_handle(), "generic");
+
+        file_transfer_service_handle = Some(
+            spawn_file_transfer_service(&task_spawner, genesis_hash, &config, &mut net_config)
+                .await,
+        );
     }
```
TDemeco, line 244: Just out of curiosity: if we shadow the `file_transfer_service_handle` variable with its content when we spawn the Blockchain service (and we never use it for anything else beforehand), why is it needed to initiate it as a `Some()` value and then `.expect` in the Blockchain service initialization?

node/src/service.rs
```diff
@@ -302,6 +309,56 @@ async fn start_dev_impl(
         telemetry: telemetry.as_mut(),
     })?;
 
+    // Spawning the Blockchain Service if node is running as a Storage Provider.
+    if let Some(provider_options) = provider_options {
+        // File Transfer Service handle is expected to be present when the node is running as a Storage Provider.
+        let file_transfer_service_handle = file_transfer_service_handle.expect(
+            "File Transfer Service handle is expected to be present when the node is running as a Storage Provider. qed",
+        );
```
TDemeco, line 317: My previous comment refers to this, we never check `.is_some` or `.is_none` with `file_transfer_service_handle`, so why do we make it an `Option`?

node/src/service.rs
```diff
@@ -450,22 +507,29 @@ async fn start_node_impl(
...
+
+        file_transfer_service_handle = Some(
+            spawn_file_transfer_service(
+                &task_spawner,
+                genesis_hash,
+                &parachain_config,
+                &mut net_config,
+            )
+            .await,
+        );
```
TDemeco, line 532: Same here ^

node/src/service.rs
```diff
@@ -551,6 +615,57 @@ async fn start_node_impl(
         telemetry: telemetry.as_mut(),
     })?;
 
+    // Spawning the Blockchain Service if node is running as a Storage Provider.
+    if let Some(provider_options) = provider_options {
+        // File Transfer Service handle is expected to be present when the node is running as a Storage Provider.
+        let file_transfer_service_handle = file_transfer_service_handle.expect(
+            "File Transfer Service handle is expected to be present when the node is running as a Storage Provider. qed",
+        );
```
TDemeco, line 623: Same here ^

node/src/services/blockchain/commands.rs
```diff
@@ -0,0 +1,121 @@
...
+    },
+}
+
+/// Interface for interacting with the BlockchainService actor.
+pub trait BlockchainServiceInterface {
+    /// Send an extrinsic to the runtime.
+    async fn send_extrinsic(
+        &self,
+        call: impl Into<storage_hub_runtime::RuntimeCall>,
+    ) -> Result<(tokio::sync::mpsc::Receiver<String>, H256)>;
```
TDemeco, line 35: The return here is a string and a hash, but what do they represent? Maybe a type alias to make it clearer could work, or adding it to the docs of the function

node/src/services/blockchain/handler.rs
```diff
@@ -72,19 +222,344 @@ impl ActorEventLoop<BlockchainService> for BlockchainServiceEventLoop {
...
+    async fn send_extrinsic(
+        &mut self,
+        call: impl Into<storage_hub_runtime::RuntimeCall>,
+    ) -> Result<RpcExtrinsicOutput> {
+        debug!(target: LOG_TARGET, "Sending extrinsic to the runtime");
+
+        // Get the nonce for the caller and increment it for the next transaction.
+        // TODO: Handle initialisation of nonce when node is restarted.
+        let nonce = self.nonce_counter;
+        self.nonce_counter += 1;
```
TDemeco, line 340: Checked math here or is panicking what's expected in that case?

node/src/services/mod.rs
```diff
@@ -61,13 +58,8 @@ impl<S: StorageHubHandlerConfig> StorageHubHandler<S> {
...
     pub fn start_bsp_tasks(&self) {
-        ResolveRemoteUploadRequest::new(self.clone())
-            .subscribe_to(&self.task_spawner, &self.file_transfer)
-            .start();
-        NewStorageRequestHandler::new(self.clone())
-            .subscribe_to(&self.task_spawner, &self.blockchain)
-            .start();
-        AcceptedBspVolunteerHandler::new(self.clone())
+        log::info!("Starting BSP tasks");
+        BspVolunteerMockTask::new(self.clone())
```
TDemeco, line 62: Minor thing but I would leave a TODO here just in case, it will definitely not happen but if we are super SUPER distracted we might miss this mock and leave it there in prod 😬

node/src/tasks/bsp_volunteer_mock.rs
```diff
@@ -0,0 +1,127 @@
...
+    async fn handle_event(&self, event: NewStorageRequest) -> anyhow::Result<()> {
+        info!(
+            target: LOG_TARGET,
+            "Initiating BSP volunteer mock for location: {:?}, fingerprint: {:?}",
+            event.location,
+            event.fingerprint
+        );
+
+        // TODO: Here we would send the actual multiaddresses of this BSP.
+        let multiaddresses = BoundedVec::default();
```
TDemeco, line 48: Just a comment for the future when we implement this, since we now have a vector of multiaddresses on the `storage-providers` pallet, we will need logic to choose which one to use

node/src/tasks/bsp_volunteer_mock.rs
```diff
@@ -0,0 +1,127 @@
...
+        while let Some(tx_result) = tx_watcher.recv().await {
+            // Parse the JSONRPC string, now that we know it is not an error.
+            let json: serde_json::Value = serde_json::from_str(&tx_result)
+                .expect("The result, if not an error, can only be a JSONRPC string; qed");
+
+            debug!(target: LOG_TARGET, "Transaction information: {:?}", json);
+
+            // Checking if the transaction is included in a block.
+            // TODO: Consider if we might want to wait for "finalized".
+            if let Some(in_block) = json["params"]["result"]["inBlock"].as_str() {
```
TDemeco, line 76: Are there any other valid `result`s apart from `inBlock` that we should handle? I remember when I used PolkadotJS the submitted transaction could be included, finalized, etc, and all those were valid statuses that we had to handle, so I want to make sure that's not the case here

TDemeco (commented):
Left some comments! Great work, it's nice to see the services taking shape 🙌🏻

node/src/services/blockchain/events.rs
```diff
@@ -8,14 +16,27 @@ pub struct ChallengeRequest {
...
```
snowmead, line 27: I think you forgot to remove this.

node/src/services/blockchain/handler.rs
```diff
@@ -72,19 +222,344 @@ impl ActorEventLoop<BlockchainService> for BlockchainServiceEventLoop {
...
+            current_block,
+        )),
+        frame_system::CheckNonce::<storage_hub_runtime::Runtime>::from(nonce),
+        frame_system::CheckWeight::<storage_hub_runtime::Runtime>::new(),
+        pallet_transaction_payment::ChargeTransactionPayment::<storage_hub_runtime::Runtime>::from(
+            tip,
+        ),
+        cumulus_primitives_storage_weight_reclaim::StorageWeightReclaim::<
+            storage_hub_runtime::Runtime,
+        >::new(),
```
snowmead, line 418: Out of curiosity, how does it work if you submit an extrinsic without these `extra` things (don't know what to call them).

node/src/services/blockchain/handler.rs
```diff
@@ -72,19 +222,344 @@ impl ActorEventLoop<BlockchainService> for BlockchainServiceEventLoop {
...
+        );
+
+        // Sign the payload.
+        let signature = raw_payload
+            .using_encoded(|e| self.keystore.sr25519_sign(KEY_TYPE, &caller_pub_key, e))
+            .expect("The payload is always valid and should be possible to sign; qed")
+            .expect("They key type and public key are valid because we just extracted them from the keystore; qed");
+
+        // Construct the extrinsic.
+        UncheckedExtrinsic::new_signed(
```
snowmead, line 453: What are the properties of an `UncheckedExtrinsic`?

node/src/services/blockchain/types.rs
```diff
@@ -0,0 +1,47 @@
...
+    pub events: EventsVec,
+}
+
+/// ExtrinsicResult enum.
+///
+/// This enum represents the result of an extrinsic execution. It can be either a success or a failure. It contains the
+/// following variants:
+/// - `Success`: the extrinsic was executed successfully.
+/// - `Failure`: the extrinsic execution failed.
+
```
snowmead, line 38: There is a space between docs and struct.

```suggestion
```

snowmead (approved):
LGTM 👍🏼 

Just a couple clarifying questions and minor changes.

node/src/services/blockchain/handler.rs
ffarall: After a couple of type aliases, it resolves to this: https://github.com/paritytech/polkadot-sdk/blob/3c3d6fceb82372a3019b37117aa453d564b212de/substrate/primitives/runtime/src/generic/unchecked_extrinsic.rs#L69

PR #40, opened 2024-04-17
feat: ✨ implement `change_capacity` extrinsic on the `Providers` pallet

TDemeco:
In this PR we:
- Organize the errors of the `Providers` pallet to keep them tidy
- Implement the `change_capacity` extrinsic to allow SPs to change their "contracted" capacity
- Refactor the Main Storage Provider and Backup Storage Provider structures to add a timestamp of the last block where the capacity was changed, to use for the timelock of `change_capacity`
- Add errors for the `change_capacity` extrinsic
- Improved documentation for the functions of the `Providers` pallet
- Add success and failure tests for new functionality

### DIFF
--- a/pallets/providers/src/lib.rs
+++ b/pallets/providers/src/lib.rs
@@ -278,7 +278,7 @@ pub mod pallet {
         /// that BSP's account id.
         BspSignOffSuccess { who: T::AccountId },
 
-        /// Event emitted when a SP has changed is total data (stake) successfully. Provides information about
+        /// Event emitted when a SP has changed its capacity successfully. Provides information about
         /// that SP's account id, its old total data that could store, and the new total data.
         CapacityChanged {
             who: T::AccountId,
--- a/pallets/providers/src/lib.rs
+++ b/pallets/providers/src/lib.rs
@@ -291,44 +291,59 @@ pub mod pallet {
     /// The errors that can be thrown by this pallet to inform users about what went wrong
     #[pallet::error]
     pub enum Error<T> {
+        // Sign up errors:
         /// Error thrown when a user tries to sign up as a SP but is already registered as a MSP or BSP.
         AlreadyRegistered,
-        /// Error thrown when a user tries to sign up or change its stake to store less storage than the minimum required by the runtime.
-        StorageTooLow,
-        /// Error thrown when a user does not have enough balance to pay the deposit that it would incur by signing up as a SP or changing its total data (stake).
-        NotEnoughBalance,
-        /// Error thrown when the runtime cannot hold the required deposit from the account to register it as a SP
-        CannotHoldDeposit,
         /// Error thrown when a user tries to sign up as a BSP but the maximum amount of BSPs has been reached.
         MaxBspsReached,
         /// Error thrown when a user tries to sign up as a MSP but the maximum amount of MSPs has been reached.
         MaxMspsReached,
-        /// Error thrown when a user tries to sign off as a SP but is not registered as a MSP or BSP.
-        NotRegistered,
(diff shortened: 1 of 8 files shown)


pallets/providers/src/lib.rs
```diff
@@ -276,54 +280,70 @@ pub mod pallet {
 
         /// Event emitted when a SP has changed is total data (stake) successfully. Provides information about
         /// that SP's account id, its old total data that could store, and the new total data.
```
a-moreira, line 282: should this comment be updated to reflect the new nomenclature (total data/stake -> capacity)?

TDemeco: Yep! Missed that one, ty!

pallets/providers/src/utils.rs
```diff
@@ -499,13 +505,296 @@ where
...
+        // Check how much has the used already deposited for the current capacity
+        let current_deposit = T::NativeBalance::balance_on_hold(
+            &HoldReason::StorageProviderDeposit.into(),
+            account_id,
+        );
+
+        // Check if the new deposit is bigger or smaller than the current deposit
+        // Note: we do not check directly capacities as, for example, a bigger new_capacity could entail a smaller deposit
+        // because of changes in storage pricing, so we check the difference in deposits instead
+        if new_deposit > current_deposit {
```
snowmead, line 708: Can we extract this logic of the `if` and `else` into functions like `hold_balance` and `release_balance`? Both functions `do_change_capacity_bsp` and `do_change_capacity_msp` are the same.

snowmead (approved):
lgtm 👍🏼

pallets/providers/src/utils.rs
TDemeco: Done!

pallets/providers/src/utils.rs
```diff
@@ -499,12 +505,270 @@ where
...
+    /// and returning the old capacity if successful
+    pub fn do_change_capacity_bsp(
+        account_id: &T::AccountId,
+        bsp_id: BackupStorageProviderId<T>,
+        new_capacity: StorageData<T>,
+    ) -> Result<StorageData<T>, DispatchError> {
+        // Check that the BSP is registered and get its info
+        let mut bsp = BackupStorageProviders::<T>::get(&bsp_id).ok_or(Error::<T>::NotRegistered)?;
+
+        // Check that the new capacity is diferent from the current capacity
```
ffarall, line 620:
```suggestion
        // Check that the new capacity is different from the current capacity
```

pallets/providers/src/tests.rs
```diff
@@ -2894,6 +2899,816 @@ mod sign_off {
...
+    mod success {
+        use super::*;
+
+        /// This module holds the success cases for changing the capacity of Main Storage Providers
+        mod msp {
+            use super::*;
+
+            #[test]
+            fn msp_increase_change_capacity_works() {
+                ExtBuilder::build().execute_with(|| {
```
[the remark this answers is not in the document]
TDemeco: I have an issue with this kind of tests: these parameters are runtime constants and, as such, the only way to modify them is through a runtime upgrade which is not easily simulatable during a single test run. I could work around this in a few different ways, but all of them include modifying the actual pallet to accommodate testing these scenarios, do you think it is worth it? If so, I'll work on doing that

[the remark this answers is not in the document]
ffarall: I've put a task in the backlog to create an integration test to test a runtime upgrade that includes such a change. We'll leave it like this for the time being.

ffarall (approved):

PR #41, opened 2024-04-17
feat: 🏗️  Init File & Forest storage utilities in StorageHub client

snowmead:


support/infra/src/storage.rs
```diff
@@ -1,32 +1,67 @@
...
+    pub proven: Proven<Chunk>,
+    /// The compact proof.
+    pub proof: CompactProof,
+    /// The root hash of the trie.
+    pub root_hash: Key,
+}
 
 /// Storage interface to be implemented by the storage providers.
-pub trait Storage: Clone + Send + Sync + 'static {
+pub trait FileStorage: 'static {
```
ffarall, line 28: Why are all these structs and and traits defined here and not in forest-manager or wherever else they're used? Do they need to be shared with others somehow?

node/src/service.rs
```diff
@@ -641,10 +644,22 @@ async fn start_sh_services(
 
     let blockchain_service_handle = spawn_blockchain_service(&task_spawner).await;
 
-    let sh_handler = StorageHubHandler::new(
+    let file_storage = Arc::new(RwLock::new(InMemoryFileStorage::new()));
```
ffarall, line 647: Bare in mind that these function is removed in the Blockchain Service branch, since initialisation of StorageHub services has to be split in two. Anyway, it can be handled in that branch, since most likely this PR will be merged first. But just remember to keep an eye on this.

node/src/service.rs
```diff
@@ -641,10 +644,22 @@ async fn start_sh_services(
...
+    let forest_storage = Arc::new(RwLock::new(InMemoryForestStorage::new()));
```
ffarall, line 648: Remind me to have a discussion on tokio's `RwLock` using a *write-first* approach.

node/src/services/mod.rs
```diff
@@ -1,34 +1,61 @@
...
 
 use crate::tasks::{
     AcceptedBspVolunteerHandler, NewStorageRequestHandler, ResolveRemoteUploadRequest,
 };
 
 use self::{blockchain::handler::BlockchainService, file_transfer::FileTransferService};
 
-#[derive(Debug, Clone)]
-pub struct StorageHubHandler {
+pub trait StorageHubHandlerConfig: Send + 'static {
```
ffarall, line 19:
```suggestion
pub trait StorageHubHandlerConfig: Send + 'static + Clone {
```

ffarall: By adding this constraint, you can use `#[derive(Clone)]` in the task definitions like `ResolveRemoteUploadRequest`, and not have to implement `Clone` manually. Less boilerplate code, and I believe it is safe to ask this constraint from `StorageHubHandlerConfig`.

client/forest-manager/src/in_memory.rs
```diff
@@ -0,0 +1,30 @@
+pub struct InMemoryForestStorage {}
+
+impl InMemoryForestStorage {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+impl storage_hub_infra::storage::ForestStorage for InMemoryForestStorage {
```
ffarall, line 9: Wouldn't it be more natural if the `ForestStorage` trait was defined in the `forest-manager` crate? Could be in `lib.rs` or in some specific file for it.

client/file-manager/src/in_memory.rs
```diff
@@ -0,0 +1,47 @@
+pub struct InMemoryFileStorage {}
+
+impl InMemoryFileStorage {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+impl storage_hub_infra::storage::FileStorage for InMemoryFileStorage {
```
ffarall, line 9: Same here, what do you think of defining `FileStorage` trait in this crate?

ffarall (commented):
Mostly approved, just some questions that wanted to ask before confirming and merging. Good work!

node/src/services/mod.rs
links234: If we add `Clone` bound to `StorageHubHandlerConfig`, it will pass the constraint to `FileStorage` and `ForestStorage`. We can change it like this but we have to make sure we never actually `clone` those types, which imo is worse thank manually implementing this.
Also the boilerplate code can be fixed later (with our own macro).

support/infra/src/storage.rs
links234: Initially I added a mock Storage trait here and went from there. It makes more sense for each to be next to the implementations as well.

client/forest-manager/src/traits.rs
```diff
@@ -0,0 +1,24 @@
+use sp_core::H256;
+use sp_trie::CompactProof;
+use storage_hub_infra::types::{Key, Metadata, Proven};
+
+pub struct ForestProof {
```
ffarall, line 5: This seems reasonable now, although bare in mind it might change once we actually implement it, for convenience.

ffarall (approved):

PR #45, opened 2024-04-23
feat: ✨ Dev mode sealing - [Instant, Manual, Interval]

snowmead:
This PR adds the ability to specify the type of sealing in `--dev` mode:

Options:
- `--sealing manual`: manually execute RPC call to create blocks
- `--sealing instant`: blocks are instantly created (not finalized, but we could add an option later)
- `--sealing 6000`: blocks are created on a millisecond interval (in this case 6 seconds)

### DIFF
--- a/node/src/cli.rs
+++ b/node/src/cli.rs
@@ -1,4 +1,4 @@
-use std::path::PathBuf;
+use std::{path::PathBuf, str::FromStr};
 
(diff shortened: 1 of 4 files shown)


a-moreira (approved):
👍

ffarall (approved):

PR #46, opened 2024-04-23
feat: implement `UserSendsFile` task

a-moreira:
This PR implements a "user task" to handle `AcceptedBspVolunteer` events from the runtime, and connect to peers to send chunk proofs.
It also adds a `User` as a `ProviderType`, temporarily.

- [x] Doc comments.


node/src/tasks/user_submitted_file.rs
```diff
@@ -0,0 +1,111 @@
...
+        let file_metadata = Metadata {
+            owner: event.owner.to_string(),
+            size: event.size.into(),
+            // TODO: use `FileLocation` instead of `String`.
+            location: format!("{:?}", event.location.into_inner()),
+            fingerprint: event.fingerprint,
+        };
+        let chunk_count = file_metadata.chunk_count();
+
+        // TODO(Arthur): double check this, I'm assuming Blake2 as the Trie hash function.
```
links234, line 60: We are not bound to use the same hash function as the Trie here but I agree we should double check.

node/src/tasks/user_submitted_file.rs
```diff
@@ -0,0 +1,111 @@
...
+        let file_key = Blake2Hasher::hash(&serde_json::to_vec(&file_metadata)?.to_owned());
```
links234, line 61: What do you think of having this logic moved in a method like `Metadata::get_file_key()`?

node/src/tasks/user_submitted_file.rs
```diff
@@ -0,0 +1,111 @@
...
+                PeerId::try_from_multiaddr(&multiaddr).expect("Multiaddr without PeerId")
+            })
+            .collect::<Vec<PeerId>>();
+
+        for peer_id in peer_ids {
+            for chunk_id in 0..chunk_count {
+                // Acquire the write lock for the file storage.
+                let file_storage_lock: tokio::sync::RwLockWriteGuard<
+                    <SHC as StorageHubHandlerConfig>::FileStorage,
+                > = self.storage_hub_handler.file_storage.write().await;
```
links234, line 81: Why do we need write lock here?

node/src/tasks/user_submitted_file.rs
```diff
@@ -0,0 +1,111 @@
...
+
+        let peer_ids = multiaddresses
+            .iter()
+            .map(|multiaddr| {
+                let multiaddr_str = multiaddr
+                    .split("/")
+                    .last()
+                    .expect("Multiaddress of Bsp should not be empty.");
+                let multiaddr =
+                    Multiaddr::from_str(multiaddr_str).expect("Failed to convert into Multiaddr");
```
links234, line 71: Probably we would like to continue with any BSPs we can instead of failing everything? Or is the assumption that this is verified by the runtime before emitting the event too?

node/src/services/file_transfer/handler.rs
```diff
@@ -146,12 +146,52 @@ impl Actor for FileTransferService {
...
+                    }
+                }
+                FileTransferServiceCommand::RegisterNewFile {
+                    peer_id,
+                    file_key,
+                    callback,
+                } => {
+                    let result = match self.peer_file_registry.insert((peer_id, file_key)) {
+                        true => Ok(()),
+                        false => Err(RequestError::RegisterNewFileFailure),
```
links234, line 168: I think it's better to be specific with what the error means. Do you think of `PeerFileAlreadyRegistered`?

node/src/services/file_transfer/handler.rs
```diff
@@ -146,12 +146,52 @@ impl Actor for FileTransferService {
...
-                FileTransferServiceCommand::UnregisterFile { peer_id, file_key } => {
-                    self.peer_file_registry.remove(&(peer_id, file_key));
+                FileTransferServiceCommand::UnregisterFile {
+                    peer_id,
+                    file_key,
+                    callback,
+                } => {
+                    let result = match self.peer_file_registry.remove(&(peer_id, file_key)) {
+                        true => Ok(()),
+                        false => Err(RequestError::UnregisterFileFailure),
```
links234, line 186: Same here. What do you think about `PeerFileNotFound`?

node/src/tasks/user_submitted_file.rs
```diff
@@ -0,0 +1,111 @@
...
+            event.owner,
+            event.location,
+        );
+
+        // TODO: use `Multiaddr` instead of `String`.
+        let multiaddresses = event.multiaddresses;
+        let file_metadata = Metadata {
+            owner: event.owner.to_string(),
+            size: event.size.into(),
+            // TODO: use `FileLocation` instead of `String`.
```
links234, line 54: I see we have `FileLocation`, if we are not missing anything major from it, let's also update it here too.

links234 (changes requested):
Good overall, left some questions/comments.

node/src/tasks/user_submitted_file.rs
snowmead: `file_metadata.key()`?

node/src/tasks/user_submitted_file.rs
snowmead: I wouldn’t fail here. Instead continue until the end and if there are no `peer_ids` at the end of this, I would fallback to querying the storage providers for their multiaddresses, and find any that might have been added after they volunteered. It is highly unlikely that there will be a change, but it could happen.

node/src/tasks/user_submitted_file.rs
```diff
@@ -0,0 +1,111 @@
...
+
+                let upload_response = self
+                    .storage_hub_handler
+                    .file_transfer
+                    .upload_request(peer_id, file_key, proof)
+                    .await;
+
+                match upload_response {
+                    Ok(_) => {
+                        info!(target: LOG_TARGET, "Successfully uploaded chunk id {:?} to peer {:?}", chunk_id, peer_id)
```
snowmead, line 98: I would make this debug! since there would be quite a lot of chunks to send 😅

node/src/tasks/user_submitted_file.rs
a-moreira: We don't, my bad. We just need to read.

node/src/tasks/user_submitted_file.rs
a-moreira: Yes, good idea.

node/src/services/file_transfer/handler.rs
a-moreira: Yes, good idea.

node/src/tasks/user_submitted_file.rs
a-moreira: You are right, I removed the failures and now just proceed with getting whatever PeerIds are available.

node/src/tasks/user_submitted_file.rs
a-moreira: Ok. The `FileLocation<T>` type comes from the runtime, while the `Metadata` type is defined in the `infra` crate. Probably not a good idea to make that crate dependent on the runtime, so I created a new type mirroring the one from the runtime.

links234 (approved):

ffarall (changes requested):
Overall good PR, the code was clean, tidy, and easy to follow. Just some minor comments, but almost ready to merge.

node/src/tasks/user_submitted_file.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        for peer_id in peer_ids {
+            for chunk_id in 0..chunk_count {
+                let proof = self
+                    .storage_hub_handler
+                    .file_storage
+                    .read()
+                    .await
+                    .generate_proof(&file_key, &chunk_id)
+                    .expect("File is not in storage, or proof does not exist.");
```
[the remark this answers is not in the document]
links234: For this flow (which requires you to merkleize the file and hence have the file in store first) this should never fail. If it fails it means that something bad happened and it's not really recoverable from here.

node/src/tasks/user_submitted_file.rs
```diff
@@ -0,0 +1,64 @@
...
+
+use log::info;
+
+use sc_network::PeerId;
+use storage_hub_infra::event_bus::EventHandler;
+
+const LOG_TARGET: &str = "user-submitted-file-task";
+
+/// Handles the events related to users submitting files to be stored.
+pub struct UserSubmittedFileTask<SHC: StorageHubHandlerConfig> {
```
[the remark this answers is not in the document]
a-moreira: Makes sense! I agree :-)

node/src/tasks/user_submitted_file.rs
```diff
@@ -0,0 +1,102 @@
...
+use log::{debug, error, info};
+
+use sc_network::PeerId;
+
+use storage_hub_infra::event_bus::EventHandler;
+use storage_hub_infra::types::Metadata;
+
+const LOG_TARGET: &str = "user-submitted-file-task";
+
+/// Handles the events related to users submitting files to be stored.
```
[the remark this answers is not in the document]
a-moreira: You're right, thanks. I wrote this right after starting to implement this and didn't review later.

ffarall (approved):
Great work!

PR #47, opened 2024-04-24
ci: :sparkles: Add rust-test ci

timbrinded:
Added new CI for our rust tests.

- Uses `nextest` because it's just better than `cargo test`
- Uses archive and partitions so that it runs across multiple runners
- Is very slow, but that's unsurprising considering the specs of GHA basic tier runners


snowmead (approved):
beast!

ffarall (approved):

PR #48, opened 2024-04-25
Feat ✨: in memory FileStorage client

links234:
This PR updates the `FileStorage` trait and implements it as an in memory store.

### DIFF
--- a/client/common/Cargo.toml
+++ b/client/common/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "forest-manager"
+name = "common"
 version = "0.1.0"
 description = ""
 homepage = { workspace = true }
@@ -22,5 +22,4 @@ trie-db = { workspace = true }
 sp-core = { workspace = true }
 sp-trie = { workspace = true, default-features = true }
 
-common = { workspace = true }
 storage-hub-infra = { workspace = true }
--- a/client/common/Cargo.toml
+++ b/client/common/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "common"
 version = "0.1.0"
-description = ""
+description = "Common types, traits and utilities shared by other client crates."
 homepage = { workspace = true }
 license = { workspace = true }
 authors = { workspace = true }
--- a/client/file-manager/src/in_memory.rs
+++ b/client/file-manager/src/in_memory.rs
@@ -1,40 +1,170 @@
-use storage_hub_infra::types::{Chunk, FileProof, Key, Metadata};
+use std::collections::HashMap;
 
-use crate::traits::FileStorage;
+use common::types::HashT;
+use sp_core::H256;
+use storage_hub_infra::types::{Chunk, ChunkId, FileProof, Key, Leaf, Metadata};
 
-pub struct InMemoryFileStorage {}
+use sp_trie::{recorder::Recorder, MemoryDB, Trie, TrieDBBuilder, TrieLayout, TrieMut};
+use trie_db::TrieDBMutBuilder;
 
-impl InMemoryFileStorage {
-    pub fn new() -> Self {
-        Self {}
+use crate::traits::{FileStorage, FileStorageError};
+
+pub struct FileData<T: TrieLayout + 'static> {
+    root: HashT<T>,
+    memdb: MemoryDB<T::Hash>,
+}
+
+impl<T: TrieLayout + 'static> FileData<T> {
+    pub fn get() {}
+}
+
+impl<T: TrieLayout + 'static> FileData<T> {
+    fn new() -> Self {
+        Self {
+            root: Default::default(),
+            memdb: MemoryDB::default(),
+        }
+    }
+
+    pub fn get_root(&self) -> H256 {
+        H256::from_slice(
+            self.root
+                .as_ref()
+                .try_into()
+                .expect("trie hash should be 32 bytes"),
+        )
     }
 }
 
-impl FileStorage for InMemoryFileStorage {
-    type Key = Key;
-    type Value = Chunk;
+pub struct InMemoryFileStorage<T: TrieLayout + 'static> {
+    pub metadata: HashMap<Key, Metadata>,
+    pub file_data: HashMap<Key, FileData<T>>,
+}
 
-    fn generate_proof(&self, _challenged_key: &Self::Key) -> FileProof<Self::Key> {
-        unimplemented!()
+impl<T: TrieLayout> InMemoryFileStorage<T> {
+    pub fn new() -> Self {
+        Self {
+            metadata: HashMap::new(),
+            file_data: HashMap::new(),
+        }
+    }
+}
+
+impl<T: TrieLayout + 'static> FileStorage for InMemoryFileStorage<T> {
+    fn generate_proof(
+        &self,
+        file_key: &Key,
+        chunk_id: &ChunkId,
+    ) -> Result<FileProof, FileStorageError> {
+        let metadata = self.metadata.get(file_key);
+
+        if metadata.is_none() {
+            return Err(FileStorageError::FileDoesNotExists);
+        }
+        let metadata = metadata.unwrap();
+
+        let file_data = self.file_data.get(file_key);
+        if file_data.is_none() {
+            panic!(
+                "Invariant broken! Metadata for file key {:?} found but no assosiated trie",
+                file_key
+            );
+        }
+        let file_data = file_data.unwrap();
+
+        if file_data.get_root() != metadata.fingerprint {
+            return Err(FileStorageError::FingerprintAndStoredFileMismatch);
+        }
+
+        let recorder: Recorder<T::Hash> = Recorder::default();
+
+        // A `TrieRecorder` is needed to create a proof of the "visited" leafs, by the end of this process.
+        let mut trie_recorder = recorder.as_trie_recorder(file_data.root);
+
+        let trie = TrieDBBuilder::<T>::new(&file_data.memdb, &file_data.root)
+            .with_recorder(&mut trie_recorder)
+            .build();
+
+        let chunk: Option<Vec<u8>> = trie
+            .get(&chunk_id.get_be_key())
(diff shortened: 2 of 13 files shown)


client/common/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "common"
+version = "0.1.0"
+description = ""
```
a-moreira, line 4: nit: It would be great to have a short description of the package :-)

client/file-manager/src/traits.rs
```diff
@@ -1,25 +1,49 @@
...
+    /// File chunk does not exists.
+    FileChunkDoesNotExists,
+    /// Failed to insert the file chunk.
+    FailedToInsertFileChunk,
+    /// Failed to get file chunk.
+    FailedToGetFileChunk,
+    /// Failed to generate proof.
+    FailedToGenerateCompactProof,
+    /// The requested file does not exists.
+    FileDoesNotExists,
```
a-moreira, line 16:
```suggestion
    FileDoesNotExist,
```

client/file-manager/src/traits.rs
```diff
@@ -1,25 +1,49 @@
-use storage_hub_infra::types::{FileProof, Metadata};
+use storage_hub_infra::types::{Chunk, ChunkId, FileProof, Key, Metadata};
+
+#[derive(Debug)]
+pub enum FileStorageError {
+    /// File chunk already exists.
+    FileChunkAlreadyExists,
+    /// File chunk does not exists.
+    FileChunkDoesNotExists,
```
a-moreira, line 8:
```suggestion
    FileChunkDoesNotExist,
```

client/file-manager/src/in_memory.rs
```diff
@@ -1,40 +1,170 @@
...
+        let metadata = metadata.unwrap();
+
+        let file_data = self.file_data.get(file_key);
+        if file_data.is_none() {
+            panic!(
+                "Invariant broken! Metadata for file key {:?} found but no assosiated trie",
+                file_key
+            );
+        }
+        let file_data = file_data.unwrap();
```
a-moreira, line 73: I think there is a more idiomatic way of expressing this snippet in Rust:
```suggestion
        let metadata = self.metadata.get(file_key).ok_or(FileStorageError::FileDoesNotExists)?;

        let file_data = match self.file_data.get(file_key) {
            Some(file_data) => file_data,
            None => panic!("Invariant broken! Metadata for file key {:?} found but no associated trie", file_key)
        };
```
But looks right nonetheless :-)

support/infra/src/types.rs
```diff
@@ -19,26 +19,50 @@ pub struct Metadata {
...
+    }
+
+    pub fn chunk_ids(&self) -> impl Iterator<Item = ChunkId> {
+        (0..self.chunk_count()).map(|n| ChunkId(n))
+    }
+}
+
+/// Typed u64 representing the index of a file [`Chunk`]. Indexed from 0.
+#[derive(Clone)]
+pub struct ChunkId(pub u64);
```
a-moreira, line 41: If not really necessary, I'd rather keep the inner type private and create a method to access it like `pub fn inner(&self) -> u64` in the `ChunkId` impl.

a-moreira: Or maybe we can just use it as a `newtype` with `type ChunkId = u64` since the only method in this Impl simply calls `to_be_bytes()`

a-moreira (approved):
Looks good to me! 😁
Just left a few comments about style and code organization.

client/file-manager/src/in_memory.rs
links234: Thanks, this look cleaner!

client/file-manager/src/in_memory.rs
```diff
@@ -1,40 +1,167 @@
...
+impl<T: TrieLayout + 'static> FileData<T> {
+    fn new() -> Self {
+        Self {
+            root: Default::default(),
+            memdb: MemoryDB::default(),
+        }
+    }
+
+    pub fn get_root(&self) -> H256 {
+        H256::from_slice(
```
ffarall, line 30: Why making `root` be something generic like `HashT<T>` but then making `get_root` return directly `H256`? In other words, why not just making `root: H256`, or going all the way with the generalisation?

client/file-manager/src/in_memory.rs
```diff
@@ -1,40 +1,167 @@
...
+            self.root
+                .as_ref()
+                .try_into()
+                .expect("trie hash should be 32 bytes"),
+        )
+    }
+}
+
+pub struct InMemoryFileStorage<T: TrieLayout + 'static> {
+    pub metadata: HashMap<Key, Metadata>,
```
ffarall, line 40: These `HashMap`s will probably be accessed MANY times by concurrent threads. How about using [DashMap](https://github.com/xacrimon/dashmap)? It would also allow us to avoid using `RwLock` on the whole `InMemoryFileStorage`.

snowmead (approved):
LGTM 👍🏼 just a few comments.

snowmead:
Can you please add tests?

client/file-manager/src/in_memory.rs
```diff
@@ -1,40 +1,167 @@
...
+
+        let mut trie =
+            TrieDBMutBuilder::<T>::new(&mut file_data.memdb, &mut file_data.root).build();
+
+        // Check that we don't have a chunk already stored.
+        if trie
+            .contains(&chunk_id.to_be_bytes())
+            .map_err(|_| FileStorageError::FailedToGetFileChunk)?
+        {
+            return Err(FileStorageError::FileChunkAlreadyExists);
```
ffarall, line 158: So this doesn't allow you to re-write a chunk? I guess for our scope no it's ok like this though.

support/infra/src/types.rs
```diff
@@ -19,26 +19,43 @@ pub struct Metadata {
...
+    }
+}
+
+/// Typed u64 representing the index of a file [`Chunk`]. Indexed from 0.
+pub type ChunkId = u64;
+
 // TODO: this is currently a placeholder in order to define Storage interface.
 /// Typed chunk of a file. This is what is stored in the leaf of the stored Merkle tree.
-pub type Chunk = [u8; FILE_CHUNK_SIZE];
+pub type Chunk = Vec<u8>;
```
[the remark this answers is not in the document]
links234: Last chunk is smaller size. Do you like the alternative to pad it with 0's more?

client/file-manager/src/in_memory.rs
```diff
@@ -1,40 +1,167 @@
...
+        let file_data = match self.file_data.get(file_key) {
+            Some(file_data) => file_data,
+            None => panic!(
+                "Invariant broken! Metadata for file key {:?} found but no associated trie",
+                file_key
+            ),
+        };
+
+        if file_data.get_root() != metadata.fingerprint {
+            return Err(FileStorageError::FingerprintAndStoredFileMismatch);
```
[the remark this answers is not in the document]
links234: We should panic only on broken invariants that the client needs to maintain (i.e. if there is an entry in metadata, there must be an entry in the file_data, even if it does not contain any chunk yet).

In this case we check if the file is complete or not (which can happen when someone requests a read and we don't have the whole file yet).

Let me split it properly.

client/file-manager/src/in_memory.rs
```diff
@@ -1,40 +1,167 @@
...
-impl InMemoryFileStorage {
+use crate::traits::{FileStorage, FileStorageError};
+
+pub struct FileData<T: TrieLayout + 'static> {
+    root: HashT<T>,
+    memdb: MemoryDB<T::Hash>,
+}
+
+impl<T: TrieLayout + 'static> FileData<T> {
+    pub fn get() {}
```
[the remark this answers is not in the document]
links234: Oops, let me delete it.

client/file-manager/src/in_memory.rs
links234: In the end we need to convert to the concrete type that the Runtime/other clients expect, which is `H256`. We have it as "generic" `HashT<T>` because that is what `TrieLayout` wants.

client/file-manager/src/in_memory.rs
```diff
@@ -1,40 +1,167 @@
...
+        // A `TrieRecorder` is needed to create a proof of the "visited" leafs, by the end of this process.
+        let mut trie_recorder = recorder.as_trie_recorder(file_data.root);
+
+        let trie = TrieDBBuilder::<T>::new(&file_data.memdb, &file_data.root)
+            .with_recorder(&mut trie_recorder)
+            .build();
+
+        let chunk: Option<Vec<u8>> = trie
+            .get(&chunk_id.to_be_bytes())
+            .map_err(|_| FileStorageError::FailedToGetFileChunk)?;
```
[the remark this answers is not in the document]
links234: `.transpose()` converts `Result<Option<T>>` into `Option<Result<T>>`. Here I want to "unwrap" the Result and and the Option and treat them differently. Also the `Result::Err` should have precedence over the `None`, so this is the right order.

support/infra/src/types.rs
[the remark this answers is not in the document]
ffarall: Using Vec seems reasonable here.

client/file-manager/src/in_memory.rs
ffarall: Yeah, I've added a [task](https://www.notion.so/moonsonglabs/Use-DashMap-for-metadata-and-file_data-of-InMemoryFileStorage-940eac6e1211482599c85e6c0e3031bb?pvs=4) to our backlog.

client/file-manager/src/in_memory.rs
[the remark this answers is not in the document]
ffarall: Discussed in call. Result: will be split into two errors. For `debug_assert` will be used in the overlaying task to panic in case this error shows up. It is technically possible that this happens if the overlaying tasks don't handle corrupt file deletions properly. But it shouldn't happen.

ffarall (approved):

PR #50, opened 2024-04-30
feat: ✨ add parent randomness to the `randomness` pallet

TDemeco:
This PR adds functionality to the `randomness` pallet to allow other pallets to obtain not only the `OneEpochAgoRandomness` that was already implemented, but also the `ParentBlockRandomness` for pallets that need more frequently changing randomness. 

Note: this new randomness is predictable by the block builder, so use carefully.


ffarall (approved):
LGTM

PR #51, opened 2024-05-01
feat: ✨ RocksDB Forest Storage

snowmead:
This PR:

- [x] Adds a new `ForestStorage` implementation - `RocksDbForestStorage`
- [x] Adds a helper method to `StorageHubHandler` to add a file to forest storage
- [x] Returns `LookupKey` on `insert`
- [x] Add known key `ROOT` to RocksDB storage to persist root key
- [x] `RocksDbForestStorage` automatically creates a trie in RocksDB if none exists (i.e. root does not exists)
- [x] Add helper function to create `RocksDbForestStorage` instance and bootstrap the storage to call this from node service
- [x] Storage layer cli option `--storage-layer [memory, rocksdb]`
- [x] Tests & Documentation
- [x] Moved `support/infra/src/types.rs` to `client/common/src/types.rs`
- [x] Generify all keys for file and forest storages to use `Hasher::Out`
- [x] Improve forest storage error model
- [x] SCALE encode/decode metadata keys
- [x] Changed all hashers to use `BlakeTwo256`
- [x] Removed associated types from Forest storage - replaced with concrete types
- [x] Add `thiserror` to Forest storage `Error`
- [x] Create file key from `Metadata` by first SCALE encoding every field individually and then concatenating them all to hash them


client/forest-manager/src/tests.rs
```diff
@@ -0,0 +1,68 @@
+use common::types::HashT;
```
ffarall, line 1: Maybe we should rename it to `test_utils.rs` or something like that? It's not too intuitive to have a file called `tests.rs` that actually has no tests.

client/forest-manager/src/prove.rs
```diff
@@ -46,122 +46,57 @@ pub(crate) fn prove<T: TrieLayout, F: ForestStorage>(
...
                 .map_err(|_| ForestStorageErrors::FailedToConstructProvenLeaves)?)
         }
-        (Some((key, value)), None) if *challenged_file_key.as_ref() > *key => {
-            // Scenario 3: After the last leaf
-            let leaf = Leaf {
-                key: key.into(),
-                data: deserialize_value(&value)?,
-            };
-            Ok(Proven::new_neighbour_keys(Some(leaf), None)
+        // Scenario 3: Before the first leaf
```
ffarall, line 64: I wouldn't want to merge this PR with this work around to an inconsistent behaviour of trie-db. Let's leave only the match clause of what would be the expected behaviour, and leave ignored tests to make sure this behaviour gets fixed in the future.

client/forest-manager/src/rocksdb.rs
```diff
@@ -0,0 +1,611 @@
...
+                    overlay: Default::default(),
+                    root,
+                };
+
+                // Create a new trie
+                let trie =
+                    TrieDBMutBuilder::<T>::new(rocksdb_forest_storage.as_hash_db_mut(), &mut root)
+                        .build();
+
+                // Drop the `trie` to free `rockdb_forest_storage` and `root`.
```
ffarall, line 151:
```suggestion
                // Drop the `trie` to free `rocksdb_forest_storage` and `root`.
```

client/forest-manager/src/rocksdb.rs
```diff
@@ -0,0 +1,611 @@
...
+    fn create_metadata(owner: &str, location: Vec<u8>, size: u64) -> Metadata {
+        Metadata {
+            owner: owner.to_string(),
+            location,
+            size,
+            fingerprint: Default::default(),
+        }
+    }
+
+    fn create_and_insert_metadata<T>(
```
ffarall, line 472: Maybe a one-line comment on what this is, like in the others?

client/forest-manager/src/rocksdb.rs
```diff
@@ -0,0 +1,611 @@
...
+    fn get(&self, key: &H::Out, prefix: Prefix) -> Result<Option<DBValue>, String> {
+        let prefixed_key = prefixed_key::<H>(key, prefix);
+        self.db.get(0, &prefixed_key).map_err(|e| {
+            warn!(target: LOG_TARGET, "Failed to read from DB: {}", e);
+            format!("Failed to read from DB: {}", e)
+        })
+    }
+}
+
+/// Backend storage used by the [`RocksDBForestStorage`].
```
ffarall, line 64:
```suggestion
/// Trait that a storage backend need to implement to be used by the [`RocksDBForestStorage`].
```

ffarall: Otherwise it is confusing with the `StorageDb` struct, which is commented to be the "Storage backend for RocksDB".

client/forest-manager/src/rocksdb.rs
```diff
@@ -0,0 +1,611 @@
...
+                    ForestStorageErrors::FailedToParseRoot
+                })
+            })
+            .transpose()?;
+
+        Ok(root)
+    }
+}
+
+/// Forest storage backend using RocksDB.
```
ffarall, line 105: Confusing docs. `StorageDb` is commented to be the "Storage backend for RocksDB", and now this is the "Forest storage backend using RocksDB"? At this point I would ask what is a storage backend? 😅

client/forest-manager/src/rocksdb.rs
```diff
@@ -0,0 +1,611 @@
...
+            warn!(target: LOG_TARGET, "Failed to read from DB: {}", e);
+            format!("Failed to read from DB: {}", e)
+        })
+    }
+}
+
+/// Backend storage used by the [`RocksDBForestStorage`].
+pub trait Backend<T: TrieLayout>: Storage<HashT<T>>
+where
+    <T::Hash as Hasher>::Out: TryFrom<Vec<u8>>,
```
ffarall, line 67: You're gonna have to change this to `TryFrom<'a &[u8]>` to use the `BlakeTwo256` hasher. Take a look at PR #58 for a guideline on how to do that.

client/forest-manager/src/rocksdb.rs
```diff
@@ -0,0 +1,611 @@
...
+        let proof = forest_storage.generate_proof(&vec![lookup_key]).unwrap();
+
+        assert_eq!(proof.proven.len(), 1);
+        assert!(
+            matches!(proof.proven.first().expect("Proven leaves should have proven 1 challenge"), Proven::ExactKey(leaf) if leaf.key.as_ref() == lookup_key)
+        );
+    }
+
+    #[test]
+    fn test_generate_proof_neighboor_keys() {
```
ffarall, line 548:
```suggestion
    fn test_generate_proof_neighbor_keys() {
```

client/forest-manager/src/rocksdb.rs
```diff
@@ -0,0 +1,611 @@
...
+        challenge[0] = challenge[0] + 1;
+
+        let proof = forest_storage.generate_proof(&vec![challenge]).unwrap();
+
+        assert_eq!(proof.proven.len(), 1);
+        assert!(
+            matches!(proof.proven.first().expect("Proven leaves should have proven 1 challenge"), Proven::NeighbourKeys((Some(leaf), None)) if leaf.key.as_ref() == largest)
+        );
+    }
+}
```
ffarall, line 611: I would add a test with more than 16 data insertions to be absolutely sure that there are intermediate nodes in the trie, and test how this codabase handles intermediate nodes. If it updates and clears them accordingly, for example.

ffarall (changes requested):

client/forest-manager/Cargo.toml
```diff
@@ -24,3 +29,6 @@ sp-trie = { workspace = true, default-features = true }
 
 common = { workspace = true }
 storage-hub-infra = { workspace = true }
+
+[features]
+test_rocksdb = []
```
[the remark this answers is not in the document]
snowmead: Forgot to remove.

client/forest-manager/src/prove.rs
snowmead: Agreed.

client/forest-manager/src/rocksdb.rs
snowmead: What about `/// RocksDB based [`ForestStorage`] implementation.`?

client/forest-manager/src/rocksdb.rs
snowmead: I use `[u8; 32]` to avoid the lifetime which would be added everywhere with `'a &[u8]`, unless you object to it.

client/forest-manager/src/rocksdb.rs
snowmead: Agreed

node/src/services/mod.rs
```diff
@@ -57,6 +135,30 @@ impl<S: StorageHubHandlerConfig> StorageHubHandler<S> {
         }
     }
 
+    /// Add file to the forest storage.
+    pub async fn _add_file_to_forest(
```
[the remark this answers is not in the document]
snowmead: Added this so it is reusable and also gives the tools for people in the team to know how to insert files into the forest storage. I will remove.

client/common/src/types.rs
```diff
@@ -1,4 +1,153 @@
...
+}
+
+#[derive(Clone, Serialize, Deserialize)]
+pub struct FileProof {
+    /// The file chunk (and id) that was proven.
+    pub proven: Leaf<ChunkId, Chunk>,
+    /// The compact proof.
+    pub proof: SerializableCompactProof,
+    /// The root hash of the trie, also known as the fingerprint of the file.
+    pub root: H256,
```
links234, line 152: `H256` would be changed to our new type `Hash` as well, but this one can be changed in another PR since it's out of scope.

client/file-manager/src/traits.rs
```diff
@@ -30,29 +31,34 @@ pub enum FileStorageWriteStatus {
...
-    fn generate_proof(&self, key: &Key, chunk_id: &ChunkId) -> Result<FileProof, FileStorageError>;
+    fn generate_proof(
+        &self,
+        key: &HasherOutT<T>,
+        chunk_id: &ChunkId,
+    ) -> Result<FileProof, FileStorageError>;
 
     /// Remove a file from storage.
-    fn delete_file(&mut self, key: &Key);
+    fn delete_file(&mut self, key: &HasherOutT<T>);
```
links234, line 44: What is the reason to change from concrete type `Key` to `HasherOutT<T>`?

client/file-manager/src/traits.rs
```diff
@@ -30,29 +31,34 @@ pub enum FileStorageWriteStatus {
 }
 
 /// Storage interface to be implemented by the storage providers.
-pub trait FileStorage: 'static {
+pub trait FileStorage<T: TrieLayout>: 'static {
```
links234, line 34: I don't see why the **trait** should be aware of the `TrieLayout`.

client/forest-manager/src/rocksdb.rs
```diff
@@ -0,0 +1,627 @@
...
+    pub const ROOT: &[u8] = b":root";
+}
+
+pub(crate) fn other_io_error(err: String) -> io::Error {
+    io::Error::new(io::ErrorKind::Other, err)
+}
+
+/// Open the database on disk, creating it if it doesn't exist.
+fn open_creating_rocksdb(db_path: String) -> io::Result<Database> {
+    let root = PathBuf::from("/tmp/");
```
links234, line 33: I don't think we want it to be saved in `/tmp/`. Maybe add a TODO?

client/forest-manager/src/rocksdb.rs
```diff
@@ -0,0 +1,627 @@
...
+mod well_known_keys {
+    pub const ROOT: &[u8] = b":root";
+}
+
+pub(crate) fn other_io_error(err: String) -> io::Error {
+    io::Error::new(io::ErrorKind::Other, err)
+}
+
+/// Open the database on disk, creating it if it doesn't exist.
+fn open_creating_rocksdb(db_path: String) -> io::Result<Database> {
```
links234, line 32:
```suggestion
fn open_or_create_rocksdb(db_path: String) -> io::Result<Database> {
```

node/src/services/blockchain/events.rs
```diff
@@ -39,7 +40,7 @@ impl EventBusMessage for NewStorageRequest {}
 pub struct AcceptedBspVolunteer {
     pub who: AccountId32,
     pub location: FileLocation,
-    pub fingerprint: H256,
+    pub fingerprint: Key,
```
links234, line 43: While the underlying data is the same (a `Vec<u8>` representing an hash) the `fingerprint` is definitely not a `Key`. I think we need to rename `Key` to `FileKey` to make it clear.

node/src/services/handler.rs
```diff
@@ -0,0 +1,175 @@
...
+pub struct StorageHubHandler<S: StorageHubHandlerConfig> {
+    /// The task spawner for spawning asynchronous tasks.
+    pub task_spawner: TaskSpawner,
+    /// The actor handle for the file transfer service.
+    pub file_transfer: ActorHandle<FileTransferService>,
+    /// The actor handle for the blockchain service.
+    pub blockchain: ActorHandle<BlockchainService>,
+    /// The file storage layer which stores all files in chunks.
+    pub file_storage: Arc<RwLock<S::FileStorage>>,
+    /// The forest storage layer which tracks all files stored in the file storage layer.
```
links234, line 126:
```suggestion
    /// The forest storage layer which tracks all complete files stored in the file storage layer.
```

node/src/services/handler.rs
```diff
@@ -0,0 +1,175 @@
...
+    ///
+    /// This layer tracks all of the files stored in [`FileStorage`](StorageHubHandlerConfig::FileStorage).
+    type ForestStorage: ForestStorage<Self::TrieLayout> + Send + Sync;
+}
+
+pub struct InMemoryStorageHubConfig<T: TrieLayout>(std::marker::PhantomData<T>);
+
+impl<T: TrieLayout + Send + 'static> StorageHubHandlerConfig for InMemoryStorageHubConfig<T>
+where
+    <<T as TrieLayout>::Hash as sp_core::Hasher>::Out: TryFrom<[u8; 32]>,
```
links234, line 38: If we are binding the hash size to 32 bytes, why not using `H256` directly for `Key` and `ForestProof`?

links234 (changes requested):
Core functionality looks good. I think there is still some confusion between hashes, file keys and how to handle them which would be good to clarify before merging.

client/common/src/types.rs
```diff
@@ -1,4 +1,153 @@
...
+    ) -> Result<Self, &'static str> {
+        match (left, right) {
+            (None, None) => Err("Both left and right leaves cannot be None"),
+            (left, right) => Ok(Proven::NeighbourKeys((left, right))),
+        }
+    }
+}
+
+/// Proof of file key(s) in the forest trie.
+pub struct ForestProof<T: TrieLayout> {
```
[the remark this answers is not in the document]
snowmead: Removed

client/forest-manager/src/lib.rs
```diff
@@ -1,5 +1,11 @@
...
 pub(crate) mod prove;
+pub mod rocksdb;
 pub mod traits;
-pub mod types;
 pub(crate) mod utils;
+
+#[cfg(test)]
+mod test_utils;
+
+const LOG_TARGET: &str = "forest_storage";
```
a-moreira, line 11: Nit: just following the standard on the rest of the project:
```suggestion
const LOG_TARGET: &str = "forest-storage";
```

node/src/services/blockchain/events.rs
a-moreira: From what we discussed last week, it seemed like a good idea to use a straightforward `[u8; 32]` for the `fingerprint` type.

client/file-manager/src/traits.rs
links234: Nevermind, this is actually the way!

client/common/src/types.rs
```diff
@@ -1,4 +1,153 @@
...
+
+// TODO: this is currently a placeholder in order to define Storage interface.
+/// Metadata contains information about a file.
+/// Most importantly, the fingerprint which is the root Merkle hash of the file.
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq)]
+pub struct Metadata {
+    pub owner: String,
+    pub location: FileLocation,
+    pub size: u64,
+    pub fingerprint: Key,
```
a-moreira, line 35: Same as Alexandru's comment below, I think setting `pub type Fingerprint = [u8; 32]` for the fingerprint would be better since it's not a file key and we know it's 32 bytes.

node/src/services/blockchain/events.rs
snowmead: After some discussion, we decided on created a `struct Fingerprint(H256)` for the `fingerprint`.

client/common/src/types.rs
snowmead: responded [here](https://github.com/Moonsong-Labs/storage-hub/pull/51#discussion_r1598712585)

client/common/src/types.rs
```diff
@@ -1,4 +1,192 @@
...
+/// FileKey is the identifier for a file.
+/// Computed as the hash of the FileMetadata.
+pub type Key = Vec<u8>;
+
+// TODO: this is currently a placeholder in order to define Storage interface.
+/// This type mirrors the `FileLocation<T>` type from the runtime, which is a BoundedVec.
+type FileLocation = Vec<u8>;
+
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq, Default)]
+pub struct Fingerprint(H256);
```
ffarall, line 28: Fingerprint is a wrapper over the `H256` type, which is essentially a wrapper over `[u8; 32]`. Why not making `Fingerprint = Vec<u8>` or `Fingerprint = [u8; 32]`?

client/common/src/types.rs
```diff
@@ -1,4 +1,192 @@
...
+
+// TODO: this is currently a placeholder in order to define Storage interface.
+/// Metadata contains information about a file.
+/// Most importantly, the fingerprint which is the root Merkle hash of the file.
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq)]
+pub struct Metadata {
+    pub owner: String,
+    pub location: FileLocation,
+    pub size: u64,
+    pub fingerprint: Fingerprint,
```
ffarall, line 63: This further convinces me that `Fingerprint = Vec<u8>` or similar. `Metadata` is the struct's whose fields we're going to SCALE-encode to get the `file_key`. The way it is right now, if you SCALE-encode a `Fingerprint`, you'll get additional bytes because it is a struct, with one field (`H256`), which also has one field (the bytes we care about). You'd have to extract the inner fields twice before SCALE-encoding. If `Fingerprint = Vec<u8>` it's easier.

client/common/src/types.rs
```diff
@@ -1,4 +1,153 @@
...
+}
+
+/// Proof of file key(s) in the forest trie.
+pub struct ForestProof<T: TrieLayout> {
+    /// The file key that was proven.
+    pub proven: Vec<Proven<HasherOutT<T>, Metadata>>,
+    /// The compact proof.
+    pub proof: CompactProof,
+    /// The root hash of the trie.
+    pub root: HasherOutT<T>,
```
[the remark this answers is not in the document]
ffarall: If it doesn't cause too much rust problems, I think this is fine. Nonetheless, to be coherent with my comment on `Fingerprint`, I think this comment from @links234 is reasonable.

client/common/src/types.rs
ffarall: Same as with `ForestProof`, also agree it's out of scope.

client/forest-manager/src/prove.rs
```diff
@@ -23,143 +27,67 @@ use crate::{traits::ForestStorage, types::ForestStorageErrors, utils::deserializ
 /// # Errors
 /// This function can return an error in cases where it fails to read or seek within the trie,
 /// or when deserialization of a leaf's value fails.
-pub(crate) fn prove<T: TrieLayout, F: ForestStorage>(
+pub(crate) fn prove<T: TrieLayout>(
     trie: &trie_db::TrieDB<'_, '_, T>,
-    challenged_file_key: &F::LookupKey,
-) -> Result<Proven<F::RawKey, F::Value>, ForestStorageErrors> {
+    challenged_file_key: &<T::Hash as Hasher>::Out,
+) -> Result<Proven<HasherOutT<T>, ()>, ErrorT<T>>
```
ffarall, line 33: The value would be the `Metadata` here, right? Why are you not returning it anymore? Wouldn't it be useful for whoever is trying to prove something? And it's not that much of an added effort.

ffarall (changes requested):
Awesome and BIG work. The refactoring looks much better now. Just a few comments before we can merge.

links234 (approved):

client/common/src/types.rs
snowmead: After some discussion with @links234 we decided to do it this way to ensure that only `Fingerprint` are compared to others of the same type. Meaning you wouldn't be able to compare a root (e.g. `&[u8]` or `H::Out`) with a `Fingerprint` unless you call `into()` from a slice, making it less likely to make an unintentional comparison.

And putting `H256` here conveys more meaning since it is a hash not just some raw bytes that could be represent anything.

client/common/src/types.rs
snowmead: Yes that is true if you encode the entire struct as a whole, but after we discussed in other channels, we decdided to encode every field in a specific order. So [here](https://github.com/Moonsong-Labs/storage-hub/pull/51/files#diff-42bb238a63ffd1e7e923b338cb833e62f00cf021ff6f69b51557a169789cfc7aR89-R102) I unpack the hash from the `fingerprint` before encoding.

client/forest-manager/src/prove.rs
snowmead: We don't need to insert the metadata as a value of the file key leaf since the only role that forest storage has to fulfilling is to track and prove file key inclusion. When a BSP requires metadata of a file, they will get it from the file storage. It also reduces the storage capacity taken by the forest storage and avoids redundant data being stored in both file and forest storage.

client/common/src/types.rs
ffarall: Conclusion: let's make `Fingerprint([u8; 32])`

client/forest-manager/src/prove.rs
ffarall: Not needed to have metadata here, it would be duplicate information, and with just the file keys gotten from this function, a task could fetch the metadata for those file keys from the file-manager client

ffarall (approved):

PR #55, opened 2024-05-02
feat: ✨ Add return verified keys to `CommitmentVerifier` trait

ffarall:


support/primitives/src/lib.rs
```diff
@@ -104,25 +113,68 @@ impl<H: Hasher> CommitmentVerifier for TrieVerifier<H> {
             match (prev_leaf, next_leaf) {
                 // Scenario 1 (valid): `next_leaf` is the challenged leaf which is included in the proof.
                 // The challenge is the leaf itself (i.e. the challenge exists in the trie).
-                (_, Some((next_key, _))) if next_key == challenge.as_ref().to_vec() => continue,
+                (_, Some((next_key, _))) if next_key == challenge.as_ref().to_vec() => {
+                    let next_key = match next_key.try_into() {
```
snowmead, line 117: you can use `.map_err()` instead of `match`

snowmead (approved):
Nice 👍🏼

PR #56, opened 2024-05-03
feat: :sparkles: Abstract `TrieVerifier` over `TrieLayout` type

ffarall:


support/primitives/src/lib.rs
```diff
@@ -13,20 +12,20 @@ mod tests;
 
 /// A struct that implements the `CommitmentVerifier` trait, where the commitment
 /// is a Merkle Patricia Trie root hash.
-pub struct TrieVerifier<H: Hasher>
+pub struct TrieVerifier<L: TrieLayout>
```
snowmead, line 15: super picky but can you make it `T` instead of `L`?

support/primitives/src/lib.rs
```diff
@@ -13,20 +12,20 @@ mod tests;
...
 where
-    <H as sp_core::Hasher>::Out: TryFrom<Vec<u8>>,
+    <<L as TrieLayout>::Hash as sp_core::Hasher>::Out: TryFrom<Vec<u8>>,
```
snowmead, line 17: You can do `T::Hash` since `T` is already constrained to `TrieLayout`.

snowmead (approved):
LGTM 👍🏼

PR #58, opened 2024-05-07
feat: ✨ Add `submit_proof` extrinsic and incorporate verifiers to runtime

ffarall:


support/primitives/src/lib.rs
```diff
@@ -12,17 +11,17 @@ mod tests;
...
 {
     pub _phantom: core::marker::PhantomData<T>,
 }
 
 /// Implement the `CommitmentVerifier` trait for the `TrieVerifier` struct.
-impl<T: TrieLayout> CommitmentVerifier for TrieVerifier<T>
+impl<T: TrieLayout, const H_LENGTH: usize> CommitmentVerifier for TrieVerifier<T, H_LENGTH>
 where
-    <T::Hash as sp_core::Hasher>::Out: TryFrom<Vec<u8>>,
+    <T::Hash as sp_core::Hasher>::Out: for<'a> TryFrom<&'a [u8; H_LENGTH]>,
```
snowmead, line 24: What is the reason behind this?

pallets/proofs-dealer/src/types.rs
```diff
@@ -50,10 +106,29 @@ pub type ProvidersPalletFor<T> = <T as crate::Config>::ProvidersPallet;
...
 /// Syntactic sugar for the type of Balance used in the NativeBalances pallet.
 pub type BalanceFor<T> = <<T as crate::Config>::NativeBalance as fungible::Inspect<
     <T as frame_system::Config>::AccountId,
 >>::Balance;
+
+/// Syntactic sugar for the type of RandomnessProvider type used in the proofs pallet.
+pub type RandomnessProviderFor<T> = <T as crate::Config>::RandomnessProvider;
+
+/// Syntactic sugar for the Randomness Output type used in the proofs pallet.
+pub type RandomnessOutputFor<T> = <T as frame_system::Config>::Hash;
```
snowmead, line 134: I don't mind this, but I would make this be an alias over the `RandomnessProvider` to ensure the hash corresponds directly to what the provider should be outputting.

pallets/proofs-dealer/src/utils.rs
```diff
@@ -49,22 +88,134 @@ where
...
         );
 
-        // TODO: Check that the proof corresponds to a correct challenge block.
+        // Check that key_proofs is not empty.
+        ensure!(!key_proofs.is_empty(), Error::<T>::EmptyKeyProofs);
 
-        // TODO: Verify proof.
+        // The check for whether forest_proof and each key_proof is not empty is handled by the corresponding
+        // verifiers for each. We do not preemptively check for this here, since the `CommitmentVerifier::Proof`
+        // type is not required to have an `is_empty` method.
```
snowmead, line 117: Is this a lingering comment?

pallets/proofs-dealer/src/utils.rs
```diff
@@ -49,22 +88,134 @@ where
...
+        // Get root for submitter.
+        // If a submitter is a registered Provider, it must have a root, so this shouldn't happen.
+        // However, since the implementation of that is not up to this pallet, we need to check.
+        let root = ProvidersPalletFor::<T>::get_root(submitter.clone())
+            .ok_or(Error::<T>::ProviderRootNotFound)?;
+
+        // Check if root is non-zero.
+        // A zero root means that the Provider is not providing any service yet, so he shouldn't be
+        // submitting any proofs.
+        ensure!(root == Self::default_forest_root(), Error::<T>::ZeroRoot);
```
snowmead, line 128: Based on previous comment: here we could just check if `root` is `None`.

pallets/proofs-dealer/src/utils.rs
```diff
@@ -49,22 +88,134 @@ where
...
+            };
+
+        // Get stake for submitter.
+        // If a submitter is a registered Provider, it must have a stake, so this shouldn't happen.
+        // However, since the implementation of that is not up to this pallet, we need to check.
+        let stake = ProvidersPalletFor::<T>::get_stake(submitter.clone())
+            .ok_or(Error::<T>::ProviderStakeNotFound)?;
+
+        // Check that the stake is non-zero.
+        ensure!(stake > BalanceFor::<T>::zero(), Error::<T>::ZeroStake);
```
snowmead, line 144: If it is zero, we should clean up all relevant data of the provider. Then any consecutive calls from the same provider would fail since they would no longer be one.

pallets/proofs-dealer/src/utils.rs
```diff
@@ -49,22 +88,134 @@ where
...
+        // Verify each key proof.
+        for key_proven in forest_keys_proven {
+            // Check that there is a key proof for each key proven.
+            let key_proof = key_proofs
+                .get(&key_proven)
+                .ok_or(Error::<T>::KeyProofNotFound)?;
+
+            // Generate the challenges for the key.
+            let challenges =
+                Self::generate_challenges_from_seed(seed, submitter, key_proof.challenge_count);
```
snowmead, line 211: Where does the `challenge_count` come from?

pallets/file-system/src/lib.rs
```diff
@@ -466,13 +464,20 @@ pub mod pallet {
             origin: OriginFor<T>,
             location: FileLocation<T>,
             root: FileKey<T>,
-            proof: Proof<T>,
+            forest_proof: ForestProof<T>,
+            key_proof: KeyProof<T>,
```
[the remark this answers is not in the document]
ffarall: The reason why I purposely separated this was because the "merged" `Proof<T>` type in the proofs dealer pallet, is used for a WHOLE proof. Basically, the key proofs in a `Proof<T>` correspond to the keys proven in the forest proof of the `Proof<T>`. In this case, the `forest_proof` will have a proof of non-inclusion of the file key being confirmed, i.e. the two neighbouring keys. `key_proof`, on the other hand, will have a proof of the file being added. Basically they don't correspond to each other, and that's why I have them separate.

pallets/file-system/src/utils.rs
```diff
@@ -318,9 +319,25 @@ where
             }
         }
 
-        // Check that the proof is valid.
-        <T::ProofDealer as storage_hub_traits::ProofsDealerInterface>::verify_proof(
-            &bsp, &root, &proof,
+        // TODO: Initialise challenges properly constructing the key for this particular file.
```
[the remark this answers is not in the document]
ffarall: Actually they won't. The `key_proof` will have proofs for chunks of the file being confirmed, while the `forest_proof` will have proofs of the two neighbouring existing keys to the one being confirmed.

pallets/file-system/src/utils.rs
```diff
@@ -318,9 +319,25 @@ where
...
+
+        // TODO: Generate challenges for the key proof properly.
+        let challenges = vec![];
+
+        // Check that the key proof is valid.
+        <T::ProofDealer as storage_hub_traits::ProofsDealerInterface>::verify_key_proof(
+            &file_key,
+            &challenges,
+            &key_proof,
         )?;
```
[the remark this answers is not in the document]
ffarall: The thing though is that in `do_submit_proof`, the keys proven in `forest_proof` have to match the `key_proofs` provided. That is not the case here, hence the separation.

pallets/proofs-dealer/src/lib.rs
```diff
@@ -13,27 +13,25 @@ mod mock;
 #[cfg(test)]
 mod tests;
```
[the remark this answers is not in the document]
ffarall: Yeah, agreed and good idea that of the diagrams. I'll keep it in mind and most likely do it in a separate PR, after adding the challenges hooks.

pallets/proofs-dealer/src/lib.rs
```diff
@@ -71,22 +88,29 @@ pub mod pallet {
             + MaxEncodedLen
             + FullCodec;
 
-        /// The type used to verify Merkle Patricia Trie proofs.
-        /// Something that implements the `CommitmentVerifier` trait.
-        type KeyVerifier: CommitmentVerifier;
+        /// The hashing system (algorithm) being used for the Merkle Patricia Forests (e.g. Blake2).
+        type MerkleHashing: Hash<Output = Self::MerkleHash> + TypeInfo;
```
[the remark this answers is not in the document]
ffarall: Good point, I'll rename it to `MerkleHasher` just to be consistent with the other type `MerkleHash`.

pallets/proofs-dealer/src/lib.rs
```diff
@@ -229,11 +265,69 @@ pub mod pallet {
         /// until some of the challenges in the queue are dispatched.
         PriorityChallengesQueueOverflow,
 
-        /// The proof submitter is not a registered Provider.
-        NotProvider,
-
         /// The fee for submitting a challenge could not be charged.
         FeeChargeFailed,
+
+        /// `submit_proof` extrinsic errors
```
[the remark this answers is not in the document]
ffarall: Not really, I separated the errors into errors emitted by each extrinsic, or "general errors". Welcome to suggestions if that was not clear 😂

pallets/proofs-dealer/src/types.rs
ffarall: I agree, but in this case, the `RandomnessProvider` is something that implements the `Randomness` trait and it doesn't have an associated type for it's output. The output is specified as a type parameter, and I do that [here](https://github.com/Moonsong-Labs/storage-hub/blob/1df731e9c35ce3c6fe4d1bcf115ae950356a66cb/pallets/proofs-dealer/src/lib.rs#L71), specifying that it should be a `Hash`.

pallets/proofs-dealer/src/types.rs
```diff
@@ -1,17 +1,60 @@
...
+    fmt::{Debug, Formatter, Result},
+};
+use storage_hub_traits::{CommitmentVerifier, ProvidersInterface};
+
+/// Type that encapsulates the proof a Provider submits.
+///
+/// The proof consists of a forest proof and a set of key proofs.
+/// A good proof would have a forest proof that proves that some keys belong to a
+/// Merkle Patricia Forest of a Provider, and the corresponding key proofs for those keys.
+#[derive(Encode, Decode, TypeInfo, PartialEq, Eq, Clone)]
```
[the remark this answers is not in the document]
ffarall: It is needed to use this type in an `Event` if I'm not mistaken.

pallets/proofs-dealer/src/utils.rs
ffarall: Nice catch! My bad. The `challenges_count` should actually be a function of the file size. I'll fix it right away.

ffarall: Actually, I was wrong, this has to be here. In the `challenges_count` field, the Storage Provider specifies how many chunks are proven in this proof. That many chunks should be related to the size of the file, but the Proofs Dealer pallet is abstracted over the `KeyVerifierProofsFor<T>` type, and therefore is not aware of the `size` of files being proven. In fact it is not even aware that what's being proved are files. So this variable is used to generate the challenges, and then it is up to the `KeyVerifier` to check that the `challenges_count` is accurate to the `size` of the file.

pallets/proofs-dealer/src/utils.rs
ffarall: Yes, but that is not the role of the Proofs Dealer pallet. In fact, this should be impossible, there shouldn't be a Storage Provider with zero stake. However, given that that is not up to the Proofs Dealer pallet to ensure, that's why I have this check anyways.

pallets/proofs-dealer/src/utils.rs
ffarall: Yes, once we make the change so that providers with no files have a `None` root, this should be changed along with that.

pallets/proofs-dealer/src/utils.rs
ffarall: No, this is a comment for why I don't start by checking if the proofs are empty. Because if I could, I would do it here, to avoid all the rest of the comptuation.

support/primitives/src/lib.rs
ffarall: If I remember correctly, it is because `H256` doesn't implement `TryFrom<Vec<u8>>`, but it implements `TryFrom<[u8; 32]>`

snowmead (approved):
Beast!

PR #59, opened 2024-05-08
feat: ✨ implement the `payment-streams` pallet

TDemeco:
This PR:
- Adds the `payment-streams` pallet, which is in charge of tracking and handling payments from users to Providers for the provided services they are using (in our case, storage). This pallet has two options to interact with it:
    - Through extrinsics that are gated for sudo-only access, useful for testing but should probably be removed before the runtime goes live.
    - Through an interface that can be used by other pallets, which allows them to create, update and delete fixed and dynamic rate payment streams. The intended usage is with this interface, as for example when a BSP volunteers to store a file the `file-system` pallet can either create a new dynamic rate stream between that user and BSP (if non existed previously) or update the stream with the storage amount provided considering this added storage (if one was already there)
- Tests almost all successful scenarios and expected failure scenarios for this pallet.
- Adds it to the runtime to allow for e2e tests.


pallets/payment-streams/src/utils.rs
```diff
@@ -0,0 +1,490 @@
...
+
+        // Check that the user is not flagged as without funds
+        ensure!(
+            !UsersWithoutFunds::<T>::contains_key(user_account),
+            Error::<T>::UserWithoutFunds
+        );
+
+        // Check if the user is already registered and, if not, try to hold the deposit
+        let mut user_payment_streams_count = RegisteredUsers::<T>::get(user_account);
+        if user_payment_streams_count == 0 {
```
ffarall, line 81: Shouldn't we hold an amount based on the number of payment streams the user has?

pallets/payment-streams/src/utils.rs
```diff
@@ -0,0 +1,490 @@
...
+    T: pallet::Config,
+{
+    /// This function holds the logic that checks if a payment stream can be created and, if so, stores the payment stream in the PaymentStreams mapping
+    /// and holds the necessary balance from the sender if it's its first payment stream
+    ///
+    /// Note: Maybe we should add a check to make sure the user has enough balance to pay for at least X amount of blocks?
+    pub fn do_create_payment_stream(
+        bsp_account: &T::AccountId,
+        user_account: &T::AccountId,
+        rate: BalanceOf<T>,
```
ffarall, line 59: Shouldn't we check for a `rate == 0`?

pallets/payment-streams/src/utils.rs
```diff
@@ -0,0 +1,490 @@
...
+            Self::deposit_event(Event::<T>::PaymentStreamCharged {
+                user_account: user_account.clone(),
+                backup_storage_provider_id: bsp_id,
+                amount: amount_charged,
+            });
+        }
+
+        // Check that the user is not flagged as without funds (after charging, to make sure it makes sense to update the payment stream)
+        ensure!(
+            !UsersWithoutFunds::<T>::contains_key(user_account),
```
ffarall, line 175: If the user is flagged as a consequence of the charging you did before, this is going to revert that. So in practice you wouldn't charge the user, nor flagged him, because the whole transaction would revert.

pallets/payment-streams/src/utils.rs
```diff
@@ -0,0 +1,490 @@
...
+            Error::<T>::UserWithoutFunds
+        );
+
+        // Update the payment stream in the PaymentStreams mapping
+        PaymentStreams::<T>::mutate(bsp_id, user_account, |payment_stream| {
+            match payment_stream {
```
ffarall, line 181: You can use an `expect_or_err!` here

pallets/payment-streams/src/utils.rs
```diff
@@ -0,0 +1,490 @@
...
+
+            // Remove the user from the UsersWithoutFunds mapping
+            UsersWithoutFunds::<T>::remove(user_account);
+        }
+
+        Ok(())
+    }
+
+    /// This function holds the logic that checks if a payment stream can be charged and, if so, charges the payment stream from the user's balance.
+    /// The charge is calculated as: rate * time_passed where time_passed is the time between the last valid proof submitted and the last charged proof of this payment stream.
```
ffarall, line 259:
```suggestion
    /// The charge is calculated as: `rate * time_passed` where `time_passed` is the time between the last valid proof submitted and the last charged proof of this payment stream.
```

pallets/payment-streams/src/utils.rs
```diff
@@ -0,0 +1,490 @@
...
+        ensure!(
+            PaymentStreams::<T>::contains_key(bsp_id, user_account),
+            Error::<T>::PaymentStreamNotFound
+        );
+
+        // Get the information of the payment stream
+        let payment_stream = PaymentStreams::<T>::get(bsp_id, user_account)
+            .ok_or(Error::<T>::PaymentStreamNotFound)?;
+
+        // Note: No need to check if a new proof has been submitted since the last charge as the only consecuence of that is charging 0 to the user,
```
ffarall, line 284:
```suggestion
        // Note: No need to check if a new proof has been submitted since the last charge as the only consequence of that is charging 0 to the user,
```

pallets/payment-streams/src/utils.rs
```diff
@@ -0,0 +1,490 @@
...
+                who: user_account.clone(),
+            });
+
+            // Return 0 as the amount that has been charged
+            Ok(Zero::zero())
+        } else {
+            // If the user does have enough funds to pay for its storage:
+
+            // Clear the user from the UsersWithoutFunds mapping
+            // TODO: Design a more robust way of handling out-of-funds users
```
ffarall, line 323: This seems like a good approach, what were you suggesting here?

pallets/payment-streams/src/utils.rs
```diff
@@ -0,0 +1,490 @@
...
+            UsersWithoutFunds::<T>::remove(user_account);
```
ffarall, line 324: Shoulnd't you check first if the user is in the mapping?

pallets/payment-streams/src/utils.rs
```diff
@@ -0,0 +1,490 @@
...
+            T::NativeBalance::transfer(
+                user_account,
+                bsp_account,
+                amount_to_charge,
+                Preservation::Preserve,
+            )?;
+
+            // Set the last charge to the block number of the last valid proof submitted
+            PaymentStreams::<T>::mutate(bsp_id, user_account, |payment_stream| {
+                match payment_stream {
```
ffarall, line 336: Why not `expect_or_err!` here?

pallets/payment-streams/src/utils.rs
```diff
@@ -0,0 +1,490 @@
...
+        if amount_charged > Zero::zero() {
+            // We emit a payment charged event only if the user had to pay for its storage before deleting the payment stream
+            Self::deposit_event(Event::<T>::PaymentStreamCharged {
+                user_account: user_account.clone(),
+                backup_storage_provider_id: bsp_id,
+                amount: amount_charged,
+            });
+        }
+
+        // TODO: What do we do when a user is flagged as without funds? Does the provider assume the loss and we remove the payment stream?
```
ffarall, line 225: Yes, if the provider is executing this, then it means that he's taking the loss and it will stop storing the file for the user.

pallets/payment-streams/src/utils.rs
```diff
@@ -0,0 +1,490 @@
...
+        // Check that the user is not flagged as without funds (after charging, to make sure it makes sense to delete the payment stream)
+        ensure!(
```
ffarall, line 227: Same here, this would revert everything you just did.

pallets/payment-streams/src/utils.rs
```diff
@@ -0,0 +1,490 @@
...
+
+        // Decrease the user's payment streams count
+        let mut user_payment_streams_count = RegisteredUsers::<T>::get(user_account);
+        user_payment_streams_count = user_payment_streams_count
+            .checked_sub(1)
+            .ok_or(ArithmeticError::Underflow)?;
+        RegisteredUsers::<T>::insert(user_account, user_payment_streams_count);
+
+        // If the user has no more payment streams, release the deposit
+        if user_payment_streams_count == 0 {
```
ffarall, line 243: Here too, why not using a deposit amount that is proportional to the payment streams a user has?

pallets/payment-streams/src/utils.rs
```diff
@@ -0,0 +1,490 @@
...
+            // Release the deposit from the user
+            T::NativeBalance::release_all(
+                &HoldReason::PaymentStreamStorageDeposit.into(),
+                &user_account,
+                Precision::Exact,
+            )?;
+
+            // Remove the user from the UsersWithoutFunds mapping
+            UsersWithoutFunds::<T>::remove(user_account);
```
ffarall, line 252: If the user suceeded in paying before, this would already be handled

pallets/payment-streams/src/utils.rs
```diff
@@ -0,0 +1,490 @@
...
+
+        // Ensure that the new last valid proof block is greater than the last valid proof block of the payment stream
+        ensure!(
+            last_valid_proof_block > payment_stream.last_valid_proof,
+            Error::<T>::InvalidLastValidProofBlockNumber
+        );
+
+        // Update the last valid proof block of the payment stream
+        PaymentStreams::<T>::mutate(bsp_id, user_account, |payment_stream| {
+            match payment_stream {
```
ffarall, line 462: Why not `expect_or_err!` here?

pallets/payment-streams/src/tests.rs
```diff
@@ -0,0 +1,1512 @@
...
+            // Try to update the rate of a payment stream that does not exist
+            assert_noop!(
+                PaymentStreams::update_payment_stream(RuntimeOrigin::root(), alice, bob, 10),
+                Error::<Test>::PaymentStreamNotFound
+            );
+        });
+    }
+
+    #[test]
+    fn update_payment_stream_fails_if_user_is_flagged_as_without_funds() {
```
ffarall, line 395: I would add a test case where the user runs out of funds BECAUSE of attempting to change a payment stream and there is an attempt to charge him. With that test, you should spot that actually the user doesn't get charged, the transaction fails, and it is not flagged as without funds.

pallets/payment-streams/src/utils.rs
```diff
@@ -0,0 +1,490 @@
...
+        let user_balance = T::NativeBalance::reducible_balance(
+            &user_account,
+            Preservation::Preserve,
+            Fortitude::Polite,
+        );
+
+        // If the user does not have enough balance to pay for its storage:
+        if user_balance < amount_to_charge {
+            // Flag it in the UsersWithoutFunds mapping and emit the UserWithoutFunds event
+            UsersWithoutFunds::<T>::insert(user_account, ());
```
ffarall, line 312: In this case, shouldn't we charge whatever he has left? Not only it is more fair, but also the way it is right now, the user might not have enough funds to pay to an SP he owes 1M DOT, but another SP with whom he also has a payment stream, might be able to charge him, let's say the 1k DOT he has left.

pallets/payment-streams/src/utils.rs
```diff
@@ -0,0 +1,490 @@
...
+        )
+        .ok_or(Error::<T>::NotABackupStorageProvider)?;
+
+        // Get the information of the payment stream to update
+        let payment_stream = PaymentStreams::<T>::get(bsp_id, user_account)
+            .ok_or(Error::<T>::PaymentStreamNotFound)?;
+
+        // Ensure that the new last valid proof block is greater than the last valid proof block of the payment stream
+        ensure!(
+            last_valid_proof_block > payment_stream.last_valid_proof,
```
ffarall, line 456: Should we also check that it's greater than `last_charged_proof`? We could check it with `expect_or_err!` as `last_checked_proof` should always be less or equal than `last_valid_proof`.

ffarall (changes requested):
Awesome work, very clean and to the point.

General comments:
- We can abstract this pallet over any provider, not just BSPs.
- This implementation doesn't account for changes in storage price over time. We can achieve it with a `price_index` variable.

pallets/payment-streams/src/lib.rs
```diff
@@ -0,0 +1,440 @@
...
+    pub struct Pallet<T>(_);
+
+    // Storage:
+
+    /// The mapping from a Backup Storage Provider to its provided users to their payment streams.
+    ///
+    /// This is used to get the payment stream of a user for a specific Backup Storage Provider.
+    ///
+    /// This storage is updated in:
+    /// - [add_payment_stream](crate::dispatchables::add_payment_stream), which adds a new entry to the map.
```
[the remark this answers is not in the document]
TDemeco: I think they are not that bad to maintain but we can delete them if we find we are changing this pallet too regularly.

pallets/payment-streams/src/lib.rs
```diff
@@ -0,0 +1,440 @@
...
+        /// Because this pallet emits events, it depends on the runtime's definition of an event.
+        type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+
+        /// Type to access the Balances pallet (using the fungible trait from frame_support)
+        type NativeBalance: Inspect<Self::AccountId>
+            + Mutate<Self::AccountId>
+            + hold::Inspect<Self::AccountId, Reason = Self::RuntimeHoldReason>
+            // , Reason = Self::HoldReason> We will probably have to hold deposits
+            + hold::Mutate<Self::AccountId, Reason = Self::RuntimeHoldReason>
+            + freeze::Inspect<Self::AccountId>
```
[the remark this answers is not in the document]
TDemeco: Nope! Fixed

pallets/payment-streams/src/lib.rs
```diff
@@ -0,0 +1,440 @@
...
+    /// The mapping from a user to if it has been flagged for not having enough funds to pay for its storage.
+    ///
+    /// This is used to flag users that do not have enough funds to pay for their storage, so other Backup Storage Providers
+    /// can stop providing storage to them.
+    ///
+    /// This storage is updated in:
+    /// - [charge_payment](crate::dispatchables::charge_payment), which emits a `UserWithoutFunds` event and sets the user's entry in this map if it does not
+    /// have enough funds, and clears the entry if it was set and the user has enough funds.
+    #[pallet::storage]
+    pub type UsersWithoutFunds<T: Config> = StorageMap<_, Blake2_128Concat, T::AccountId, ()>;
```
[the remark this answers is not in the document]
TDemeco: Nope! Should it be? Haven't finalized the "user without funds" logic just yet, I think that's probably for another PR

pallets/payment-streams/src/lib.rs
```diff
@@ -0,0 +1,440 @@
...
+
+    /// This enum holds the HoldReasons for this pallet, allowing the runtime to identify each held balance with different reasons separately
+    ///
+    /// This allows us to hold tokens and be able to identify in the future that those held tokens were
+    /// held because of this pallet
+    #[pallet::composite_enum]
+    pub enum HoldReason {
+        /// Deposit that a user has to pay to open payment streams (TODO: This should go on user registration I believe, not in this pallet)
+        PaymentStreamStorageDeposit,
+        // TODO: Only for testing, remove this for production
```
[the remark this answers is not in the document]
TDemeco: Done!

pallets/payment-streams/src/types.rs
```diff
@@ -0,0 +1,27 @@
...
+}
+
+// Type aliases:
+
+/// BalanceOf is the balance type of the runtime.
+pub type BalanceOf<T> =
+    <<T as Config>::NativeBalance as Inspect<<T as frame_system::Config>::AccountId>>::Balance;
+
+/// BackupStorageProviderId is the type that represents an ID of a Backup Storage Provider, uniquely linked with an AccountId
+pub type BackupStorageProviderId<T> = <T as frame_system::Config>::Hash;
```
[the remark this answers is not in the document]
TDemeco: Fixed!

pallets/payment-streams/src/utils.rs
```diff
@@ -0,0 +1,490 @@
...
+}
+
+impl<T> Pallet<T>
+where
+    T: pallet::Config,
+{
+    /// This function holds the logic that checks if a payment stream can be created and, if so, stores the payment stream in the PaymentStreams mapping
+    /// and holds the necessary balance from the sender if it's its first payment stream
+    ///
+    /// Note: Maybe we should add a check to make sure the user has enough balance to pay for at least X amount of blocks?
```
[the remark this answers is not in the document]
TDemeco: Yep, changed

pallets/payment-streams/src/utils.rs
TDemeco: Even better, we hold it separately for each stream according to the details of each one now!

pallets/payment-streams/src/utils.rs
TDemeco: Yep, missed that. Fixed!

pallets/payment-streams/src/utils.rs
TDemeco: Absolutely, don't know how I missed the revertion. Ty

pallets/payment-streams/src/utils.rs
TDemeco: Yeah, I was mostly thinking about what happens when a user gets set as out-of-funds, and if there's anything there that should be checked before clearing the user from the out-of-funds map. I'll revisit this more in-depth in a future PR

pallets/payment-streams/src/utils.rs
TDemeco: Should I? If it's not on the map, the `remove` method should do nothing, so we don't lose anything by not checking first if we have to remove it either way

pallets/payment-streams/src/utils.rs
TDemeco: Yep, nice catch. Fixed

pallets/payment-streams/src/utils.rs
TDemeco: Ok! I'll work on all the out-of-funds logic in another PR, this is already quite big as it is

pallets/payment-streams/src/utils.rs
TDemeco: Yep. I'll work on the out-of-funds logic in another PR. Agree with everything you just said here

ffarall (approved):

primitives/traits/src/lib.rs
```diff
@@ -60,6 +60,9 @@ pub trait ProvidersInterface {
     /// Get the ProviderId from AccountId, if it is a registered Provider.
     fn get_provider_id(who: Self::AccountId) -> Option<Self::ProviderId>;
 
+    /// Get the payment account of a registered Provider.
+    fn get_provider_payment_account(who: Self::ProviderId) -> Option<Self::AccountId>;
```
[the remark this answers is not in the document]
TDemeco: Done 👍🏻

PR #61, opened 2024-05-10
feat: ✨ add payment accounts to SPs to receive payment from streams

TDemeco:
This PR:
- Adds a field to the Storage Provider structures called `payment_account`, which is the address where the storage provider wants to receive payment for its services.
- Changes naming on the common interfaces between pallets to better reflect what the types and methods actually represent (`Provider` to `ProviderId`, `get_provider` to `get_provider_id`).
- Updates all pallets to use this new naming.
- Adds a new method to the `ProvidersInterface` called `get_provider_payment_account`, which receives a `ProviderId` and returns the payment account of that provider (if it exists). Used in the `payment-streams` pallet.
- Updates the `payment-stream` pallet to use provider IDs directly instead of account IDs that are registered as providers, and adds the necessary checks for that.

Note: a possible improvement, since we are now using generic provider IDs, would be to revamp naming on the `payment-streams` pallet to never mention "Backup Storage Providers" and to refer to them as simply "Providers", reflecting on the generic aspect of it.

### DIFF
--- a/pallets/payment-streams/src/lib.rs
+++ b/pallets/payment-streams/src/lib.rs
@@ -24,7 +24,7 @@ mod tests;
 pub use pallet::*;
 use scale_info::prelude::vec::Vec;
 pub use scale_info::Type;
-use types::{BackupStorageProviderId, BalanceOf, PaymentStream};
+use types::*;
 
 #[frame_support::pallet]
 pub mod pallet {
@@ -35,6 +35,7 @@ pub mod pallet {
     };
     use frame_system::pallet_prelude::{BlockNumberFor, *};
     use sp_runtime::traits::Convert;
+    use storage_hub_traits::ProvidersInterface;
 
     /// Configure the pallet by specifying the parameters and types on which it depends.
     #[pallet::config]
@@ -52,12 +53,7 @@ pub mod pallet {
             + freeze::Mutate<Self::AccountId>;
 
         /// The trait for reading storage provider data.
-        type Providers: storage_hub_traits::ProvidersInterface<
-            Balance = Self::NativeBalance,
-            AccountId = Self::AccountId,
-            Provider = <Self as frame_system::Config>::Hash,
-            MerkleHash = <Self as frame_system::Config>::Hash,
-        >;
+        type ProvidersPallet: ProvidersInterface<AccountId = Self::AccountId>;
 
         /// The overarching hold reason
         type RuntimeHoldReason: From<HoldReason>;
@@ -90,7 +86,7 @@ pub mod pallet {
     pub type PaymentStreams<T: Config> = StorageDoubleMap<
         _,
         Blake2_128Concat,
-        BackupStorageProviderId<T>,
+        ProviderIdFor<T>,
         Blake2_128Concat,
         T::AccountId,
         PaymentStream<T>,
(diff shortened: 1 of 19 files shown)


ffarall (approved):
Approved but do not merge before merging other PRs this depends on. Also to manage the conflicts here.

pallets/payment-streams/src/tests.rs
```diff
@@ -99,21 +100,28 @@ mod create_stream {
...
             register_account_as_bsp(alice, 100);
+            let alice_bsp_id =
+                <StorageProviders as ProvidersInterface>::get_provider_id(alice).unwrap();
 
             // Create a payment stream from Bob to Alice of 10 units per block
             let rate: BalanceOf<Test> = 10;
             assert_ok!(PaymentStreams::create_payment_stream(
                 RuntimeOrigin::root(),
-                alice,
+                alice_bsp_id,
```
snowmead, line 112: So now every BSP has to keep track of their provider id in order to execute these extrinsics? I think we should be using the base account id explicitely for extrinsics and storage related queries that would not necessitate users or providers having to keep  a mapping of this locally or having to pre query storage to get the provider id before executing something.

pallets/payment-streams/src/lib.rs
```diff
@@ -229,20 +227,15 @@ pub mod pallet {
         #[pallet::weight(Weight::from_parts(10_000, 0) + T::DbWeight::get().writes(1))]
         pub fn create_payment_stream(
```
snowmead, line 228: Why don't we remove this extrinsic and utilize unit tests to test `do_create_payment_stream` and eventually integration tests? I found it confusing that this was here and it confirmed my suspicion after seeing the docs for this stating it was for testing purposes only.

pallets/payment-streams/src/utils.rs
```diff
@@ -54,15 +55,15 @@ where
...
         rate: BalanceOf<T>,
     ) -> DispatchResult {
-        // Get the BSP ID of the BSP account
-        let bsp_id = <T::Providers as storage_hub_traits::ProvidersInterface>::get_provider(
-            bsp_account.clone(),
-        )
-        .ok_or(Error::<T>::NotABackupStorageProvider)?;
+        // Check that the given ID belongs to an actual provider
+        ensure!(
+            <T::ProvidersPallet as ProvidersInterface>::is_provider(*bsp_id),
```
snowmead, line 64: `is_provider` does not guarantee that the id is a BSP, it could be an MSP. I would also add a check using `is_bsp`. And would do this everywhere where we want to ensure the id is in fact a BSP.

snowmead (approved):
Looks good, just left a few comments.

pallets/payment-streams/src/tests.rs
TDemeco: Nope! It is a bit confusing so I wrote it on the docs of each one, but the extrinsics of this pallet (except `charge_payment_stream`) are not to be actually used in prod (the `PaymentStreamsInterface` should be used by another pallet instead), instead they are designed to help with manual testing in scenarios such as when trying the chain in polkadot{.js}.

This is why they have as a parameter the `sp_id` instead of the `sp_account`, because for the functionality of creating, updating or deleting a payment stream to be executed, the file system pallet should call the payment streams pallet with its interface when a SP's request to store is accepted or a user executes a file deletion

pallets/payment-streams/src/lib.rs
TDemeco: This is not only for rust tests but also for manual tests, but I do agree that considering that these extrinsics should be removed before deploying in production it does not make a lot of sense to use them in the unit tests, so I will change them to use the `PaymentStreamsInterface` functions instead

pallets/payment-streams/src/utils.rs
TDemeco: This was actually generic for a storage provider but the naming and docs did not reflect that. The latest commit now fixes that, ty 👍🏻

PR #65, opened 2024-05-13
Feat: BSP file transfer task

links234:
This PR adds:
 - `BspUploadRequestHandler` which listens to file transfer upload events 
 - File transfer service filters unregistered upload requests
 - File transfer service unregister command now only needs the file key
 - Update `BspVolunteerMock` to optimistically register file key and peers in file transfer service
 - As a drive-by refactor:
   - Rename `Key` to `FileKey`
   - Rename `Metadata` to `FileMetadata`


snowmead (approved):
lgtm 👍🏼

node/src/tasks/bsp_upload_request.rs
```diff
@@ -0,0 +1,149 @@
...
+impl<SHC: StorageHubHandlerConfig> EventHandler<RemoteUploadRequest>
+    for BspUploadRequestHandler<SHC>
+{
+    async fn handle_event(&self, event: RemoteUploadRequest) -> anyhow::Result<()> {
+        if !event.chunk_with_proof.verify() {
+            error!(
+                target: LOG_TARGET,
+                "Received invalid proof for chunk: {} (file: {}))", event.chunk_with_proof.proven.key, event.file_key
+            );
+            // TODO: Record this for further reputation actions
```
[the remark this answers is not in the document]
ffarall: Here we should definitely remove the user from the registry in the file transfer service.

node/src/tasks/bsp_upload_request.rs
```diff
@@ -0,0 +1,158 @@
...
+    HasherOutT<SHC::TrieLayout>: TryFrom<[u8; 32]>,
+{
+    async fn handle_event(&self, event: RemoteUploadRequest) -> anyhow::Result<()> {
+        if !event.chunk_with_proof.verify() {
+            error!(
+                target: LOG_TARGET,
+                "Received invalid proof for chunk: {} (file: {:?}))", event.chunk_with_proof.proven.key, event.file_key
+            );
+            // TODO: Record this for further reputation actions
+            return Ok(());
```
ffarall, line 50: Why returning `Ok` instead of `Error`?

node/src/tasks/bsp_upload_request.rs
```diff
@@ -0,0 +1,158 @@
...
+    handler::{StorageHubHandler, StorageHubHandlerConfig},
+};
+
+const LOG_TARGET: &str = "bsp-upload-request-handler";
+
+pub struct BspUploadRequestHandler<SHC: StorageHubHandlerConfig> {
+    storage_hub_handler: StorageHubHandler<SHC>,
+}
+
+impl<SHC: StorageHubHandlerConfig> Clone for BspUploadRequestHandler<SHC> {
```
ffarall, line 22: Why not making one single task file called `bsp_receive_file` whith two handlers, one for `RemoteUploadRequest` and one for `NewStorageRequest`. I think it would make it easier to follow the whole flow of that overarching task.

node/src/tasks/bsp_upload_request.rs
```diff
@@ -0,0 +1,158 @@
...
+                "Received invalid proof for chunk: {} (file: {:?}))", event.chunk_with_proof.proven.key, event.file_key
+            );
+            // TODO: Record this for further reputation actions
+            return Ok(());
+        }
+
+        let file_key: HasherOutT<SHC::TrieLayout> = TryFrom::try_from(*event.file_key.as_ref())
+            .map_err(|_| anyhow::anyhow!("File key and HasherOutT mismatch!"))?;
+
+        let mut write_file_storage = self.storage_hub_handler.file_storage.write().await;
```
ffarall, line 56: We could use a bit more comments in this function. For example here, we could say that we're capturing the `RwLock`, and that's why we then want to drop it fast.

node/src/tasks/bsp_upload_request.rs
```diff
@@ -0,0 +1,158 @@
...
+            &event.chunk_with_proof.proven.data,
+        );
+        drop(write_file_storage);
+
+        match write_chunk_result {
+            Ok(outcome) => match outcome {
+                FileStorageWriteOutcome::FileComplete => self.on_file_complete(&file_key).await,
+                FileStorageWriteOutcome::FileIncomplete => {}
+            },
+            Err(error) => match error {
```
ffarall, line 69: In any of this cases, shouldn't we also unregister the user and file from the file transfer service?

node/src/tasks/bsp_upload_request.rs
```diff
@@ -0,0 +1,158 @@
...
+        // Save [`FileMetadata`] of the newly stored file in the forest storage.
+        let mut write_forest_storage = self.storage_hub_handler.forest_storage.write().await;
+        let file_key = write_forest_storage
+            .insert_metadata(&metadata)
+            .expect("Failed to insert metadata.");
+        let read_forest_storage = write_forest_storage.downgrade();
+        let _forest_proof = read_forest_storage
+            .generate_proof(vec![file_key])
+            .expect("Failed to generate forest proof.");
+        drop(read_forest_storage);
```
ffarall, line 133: Here also, it could be documented why we do this.

node/src/services/handler.rs
```diff
@@ -21,6 +25,7 @@ pub trait StorageHubHandlerConfig: StorageHubHandlerInitializer + Send + 'static
     ///
     /// This is primarily used for constructing
     type TrieLayout: TrieLayout;
+
```
ffarall, line 28: Should this be here? Seems inconsistent with the other spacing used in other fields.

node/src/services/file_transfer/handler.rs
```diff
@@ -50,10 +53,18 @@ use super::{
 const LOG_TARGET: &str = "file-transfer-service";
 
 pub struct FileTransferService {
+    /// Protocol name used by substrate network for the file transfer service.
     protocol_name: ProtocolName,
+    /// Receiver for incoming requests.
     request_receiver: async_channel::Receiver<IncomingRequest>,
+    /// Substrate network service that gives access to P2p operations.
```
ffarall, line 60:
```suggestion
    /// Substrate network service that gives access to p2p operations.
```

ffarall (changes requested):

node/src/tasks/bsp_volunteer_mock.rs
```diff
@@ -43,6 +53,24 @@ impl<SHC: StorageHubHandlerConfig> EventHandler<NewStorageRequest> for BspVolunt
             event.fingerprint
         );
 
+        let result = self.inner_handle_event(event).await;
+        if result.is_err() {
+            if let Some(file_key) = &self.file_key_cleanup {
+                let _ = self
```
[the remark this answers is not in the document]
links234: The error can only be `FileNotRegistered` which is normal to expect since we might have 2 parallel requests fail and try to remove the file key.
I will log it for now as an warning.

node/src/tasks/bsp_volunteer_mock.rs
```diff
@@ -62,13 +90,40 @@ impl<SHC: StorageHubHandlerConfig> EventHandler<NewStorageRequest> for BspVolunt
...
+        };
+
+        // Get the file key.
+        let file_key: FileKey = metadata
+            .key::<<SHC::TrieLayout as TrieLayout>::Hash>()
+            .as_ref()
+            .try_into()?;
+
+        // Optimistically register the file for upload in the file transfer service.
+        // This solves the race condition between the user and the BSP.
```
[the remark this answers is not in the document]
links234: Yes. If the BSP waits for the blockchain service to emit `AcceptedBspVolunteer` (same as the user), there is no guarantee that the BSP will register it before the user.

node/src/tasks/bsp_upload_request.rs
links234: The only ones that are not an internal error are `FileDoesNotExists` or `FileChunkAlreadyExists`.

If the file doesn't exists it means that either:
1. There is a bug and we didn't add any metadata of the file before starting to receive the file
2. The file was deleted by another parallel request

If the file chunk already exists .. should be unregister (and delete file) directly or just update reputation and decide later?

node/src/tasks/bsp_upload_request.rs
```diff
@@ -0,0 +1,149 @@
...
+            return Ok(());
+        }
+
+        let mut write_file_storage = self.storage_hub_handler.file_storage.write().await;
+        let write_chunk_result = write_file_storage.write_chunk(
+            &event.file_key,
+            &event.chunk_with_proof.proven.key,
+            &event.chunk_with_proof.proven.data,
+        );
+        drop(write_file_storage);
```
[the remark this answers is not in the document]
links234: I'm not sure how that works/what you mean exactly. Let's discuss it.

node/src/tasks/bsp_upload_request.rs
links234: Left more like an todo, assuming no need to take action now but after some reputation threshold. Let me change it since we are not doing that for now.

node/src/tasks/bsp_upload_file_task.rs
```diff
@@ -0,0 +1,339 @@
...
+                    )));
+                }
+            },
+        }
+
+        Ok(())
+    }
+}
+
+impl<SHC: StorageHubHandlerConfig> EventHandler<NewStorageRequest> for BspUploadFileTask<SHC> {
```
ffarall, line 129: I would move this more to the top of this file, as this is the first event that is handled by the flow of this task.

ffarall: Also, it could be worth to comment what part of the flow of this task this event is handling.

node/src/tasks/bsp_upload_file_task.rs
```diff
@@ -0,0 +1,339 @@
...
+impl<SHC: StorageHubHandlerConfig> BspUploadFileTask<SHC> {
+    pub fn new(storage_hub_handler: StorageHubHandler<SHC>) -> Self {
+        Self {
+            storage_hub_handler,
+            file_key_cleanup: None,
+        }
+    }
+}
+
+impl<SHC: StorageHubHandlerConfig> EventHandler<RemoteUploadRequest> for BspUploadFileTask<SHC>
```
ffarall, line 50: Could be worth to comment what part of the flow of this task this handler is doing.

node/src/tasks/bsp_upload_file_task.rs
```diff
@@ -0,0 +1,339 @@
...
+            .storage_hub_handler
+            .file_transfer
+            .unregister_file(file_key.as_ref().into())
+            .await;
+
+        // Delete the file from the file storage.
+        let mut write_file_storage = self.storage_hub_handler.file_storage.write().await;
+        write_file_storage.delete_file(&file_key);
+
+        // TODO: Send transaction to runtime to unvolunteer the file.
```
ffarall, line 218: I'd have to be convinced on why we would need such transaction 🤔

node/src/tasks/bsp_upload_file_task.rs
```diff
@@ -0,0 +1,339 @@
...
+        info!(
+            target: LOG_TARGET,
+            "Initiating BSP volunteer mock for location: {:?}, fingerprint: {:?}",
+            event.location,
+            event.fingerprint
+        );
+
+        let result = self.handle_new_storage_request_event(event).await;
+        if result.is_err() {
+            if let Some(file_key) = &self.file_key_cleanup {
```
ffarall, line 140: What is this `file_key_cleanup`?

ffarall (approved):
Approved with suggestion of adding comments and questions

PR #66, opened 2024-05-15
feat: ✨ Upgrade trie-db 0.29.1

snowmead:
There was a bug with the double ended iteration implemented in `trie-db` which broke the behaviour expected for `next_back()` to yield either `None` or the previous leaf node after `seek(challenge)`. 

The fix was merged in https://github.com/paritytech/trie/pull/215 with a minor version upgrade from  `trie-db@0.29.0` to `trie-db@0.29.1`. Since it is a minor version, we can apply it to our dependencies without requiring `sp-trie` to upgrade it's sub dependency on `trie-db`.

Changes:
- Upgrades `trie-db` to `0.29.1`
- Removed ignored unit tests that were failing before due to bug in previous `trie-db` version
- Remove false check for `prev_leaf` which can be `None` now based on fixed double ended trie iteration behaviour
- Fix broken test to match `prev_leaf` to be `None` instead of being equal to `next_leaf`

### DIFF
--- a/support/primitives/src/tests.rs
+++ b/support/primitives/src/tests.rs
@@ -348,10 +348,7 @@ fn commitment_verifier_challenge_key_before_first_key_success() {
 
         let challenged_key_vec = challenge_key.to_vec();
 
-        // Assert that challenge_key is below next_leaf and that prev_leaf is equal to next_leaf.
-        // This is due to some inconsistent behaviour in the iterator, that when you seek to a key
-        // that is less than the first key, it will return the first key as the next_back leaf,
-        // even if it is not lower than the challenge key.
+        // Assert that challenge_key is below next_leaf and that prev_leaf is None.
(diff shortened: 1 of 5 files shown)


links234 (approved):

support/primitives/src/tests.rs
```diff
@@ -352,7 +352,7 @@ fn commitment_verifier_challenge_key_before_first_key_success() {
         // This is due to some inconsistent behaviour in the iterator, that when you seek to a key
         // that is less than the first key, it will return the first key as the next_back leaf,
         // even if it is not lower than the challenge key.
-        assert!(prev_leaf == next_leaf && challenged_key_vec < next_leaf.unwrap().unwrap().0);
+        assert!(prev_leaf.is_none() && challenged_key_vec < next_leaf.unwrap().unwrap().0);
```
ffarall, line 355: Change doc comment to:
```
Assert that challenge_key is below next_leaf and that prev_leaf is None.
```

ffarall (approved):

PR #71, opened 2024-05-22
refactor: ♻️  Implement builder pattern in `StorageHubHandler` and abstract `start_node_impl` over storage layer

ffarall:


a-moreira (approved):

node/src/cli.rs
```diff
@@ -61,7 +61,7 @@ pub enum ProviderType {
 #[derive(ValueEnum, Clone, Debug)]
 pub enum StorageLayer {
     /// RocksDB
-    Rocksdb,
+    RocksDB,
```
snowmead, line 64: Reason why I initially named this with lowercase "db" is because the `ValueEnum` macro separates the string version of the command based on camel case. Not sure if it will work but before I remember naming it `RocksDb` and it would run with `rocks-db`.

node/src/service.rs
```diff
@@ -665,6 +646,31 @@ async fn start_node_impl(
         );
     }
 
+    // If node is running as a Storage Provider, start building the StorageHubHandler using the StorageHubBuilder.
+    let mut sh_builder = None;
+    if provider_options.is_some() {
```
snowmead, line 651: We could extract everything in this if statement into a function, since it is the same for both the node and dev modes.

node/src/services/builder.rs
```diff
@@ -0,0 +1,166 @@
...
+
+use crate::service::{ParachainClient, ParachainNetworkService};
+
+use super::{
+    blockchain::{spawn_blockchain_service, BlockchainService},
+    file_transfer::{spawn_file_transfer_service, FileTransferService},
+    handler::StorageHubHandler,
+};
+
+/// Builds the `StorageHubHandler` by adding each component separately.
```
snowmead, line 24:
```suggestion
/// Builds the [`StorageHubHandler`] by adding each component separately.
```

I love linking in docs...

node/src/services/builder.rs
```diff
@@ -0,0 +1,166 @@
...
+            file_transfer: None,
+            blockchain: None,
+            file_storage: None,
+            forest_storage: None,
+            provider_pub_key: None,
+            _marker: Default::default(),
+        }
+    }
+
+    pub async fn with_file_transfer(
```
snowmead, line 55: I would add some small docs for each of the `with_*` functions just to give some clarity on what this would actually be doing in the background.

node/src/services/builder.rs
```diff
@@ -0,0 +1,166 @@
...
+impl<T> StorageLayerBuilder
+    for StorageHubBuilder<T, InMemoryFileStorage<T>, RocksDBForestStorage<T>>
+where
+    T: TrieLayout + Send + Sync,
+    HasherOutT<T>: TryFrom<[u8; 32]>,
+{
+    fn setup_storage_layer(&mut self) -> &mut Self {
+        let provider_pub_key = self
+            .provider_pub_key
+            .expect("Provider public key not set before building the storage layer.");
```
snowmead, line 154: This is not necessary, but with builder patterns you could instead have a function called `as_provider` that returns another struct builder called `StorageHubProviderBuilder` and this where the struct would be implemented. 

So the `setup_storage_layer` is only callable once you have the appropriate builder, and from there you can execute `with_provider_pub_key` which would also be moved to this builder. Finally execute `build()` which would execute `setup_storage_layer()` for you since being a provider requires a storage layer no matter what.

snowmead: It also separates the layers of services we expose with the builders to be available only for specific needs.

snowmead (approved):
Nice work @a-moreira @ffarall!

node/src/cli.rs
ffarall: You're right, from running `--help` we see that:
```
Type of StorageHub provider [default: memory] [possible values: rocks-db, memory]
```
Anyway, I don't think this is bad.

node/src/services/builder.rs
ffarall: I see your point. However setting up the storage layer is one of the first things we do, not the last. So that conversion of two different `Builders` would kind of be wasted I think.

PR #72, opened 2024-05-22
feat: ✨ RPC implementation for file upload

a-moreira:
Allows file uploading through Substrate's JSON RPC.


links234 (approved):
LGTM!

client/rpc/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            let mut t = TrieDBMutBuilder::<T>::new(&mut memdb, &mut root).build();
+            for (chunk_id, chunk) in file_chunks.iter().enumerate() {
+                let chunk_id = chunk_id.to_be_bytes();
+                t.insert(&chunk_id, chunk).expect("error");
+            }
+        }
+
+        // Acquire FileStorage write lock.
+        let mut file_storage_lock = self.file_storage.write().await;
```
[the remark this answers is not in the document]
a-moreira: you're right, thanks!

client/rpc/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+        let file_metadata = FileMetadata {
+            size: file_size,
+            fingerprint: root.as_ref().into(),
+            owner: owner.to_string(),
+            location: location.clone().into(),
+        };
+        let file_key = file_metadata.key::<T::Hash>();
+
+        // Acquire FileStorage write lock.
+        let mut file_storage_lock = self.file_storage.write().await;
```
links234, line 126: What do you think of having this named `write_file_storage` instead?

client/rpc/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+        owner: AccountId32,
+    ) -> RpcResult<FileMetadata> {
+        let mut file = File::open(PathBuf::from(file_path.clone())).map_err(into_rpc_error)?;
+        let mut file_chunks = Vec::new();
+
+        // Read file in chunks of [`FILE_CHUNK_SIZE`] into buffer then push buffer into a vector.
+        // Loops until EOF or until some error that is NOT `ErrorKind::Interrupted` is found.
+        // If `ErrorKind::Interrupted` is found, the operation is simply retried, as per
+        // https://doc.rust-lang.org/std/io/trait.Read.html#errors-1
+        loop {
```
links234, line 81: You should add chunks to the `file_data_trie` right after reading, this way (if you have a RocksDb implementation) you are not holding the whole file in memory.

client/rpc/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+/// Interface generated by the `rpc` macro from our `FileStorageApi` trait.
+#[async_trait]
+impl<FL, T> FileStorageApiServer for FileStorageRpc<FL, T>
+where
+    FL: Send + Sync + FileStorage<T>,
+    T: Send + Sync + TrieLayout + 'static,
+{
+    async fn upload_file(
```
links234, line 68: I think we either call this 'load_file_in_store' or also add some mechanism to watch for this file.

I think right now the `UserSendsFileTask` processes all events and will just panic if the file is not found in `file_storage`.

links234 (changes requested):
Looks good, I left some suggestions.

client/rpc/src/lib.rs
a-moreira: Good catch, thanks.

client/rpc/src/lib.rs
a-moreira: Found a name in between: `file_storage_write_lock` 😄

links234 (approved):

client/rpc/src/lib.rs
a-moreira: Makes sense, I'll rename and add a TODO here

ffarall (changes requested):
RPC method is not showing in PolkadotJS to test it out. Please fix.

client/rpc/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+    T: Send + Sync + TrieLayout + 'static,
+{
+    async fn load_file_in_storage(
+        &self,
+        file_path: String,
+        location: String,
+        owner: AccountId32,
+    ) -> RpcResult<FileMetadata> {
+        // Open file in the local file system.
+        let mut file = File::open(PathBuf::from(file_path.clone())).map_err(into_rpc_error)?;
```
snowmead, line 78: I believe we should store the files with the names being the `file_key`. This would avoid any possible collisions with other uploaded files. It would be more ergonomic to use `file_key` across all of our query methods.

client/rpc/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+            location: location.clone().into(),
+        };
+        let file_key = file_metadata.key::<T::Hash>();
+
+        // Acquire FileStorage write lock.
+        let mut file_storage_write_lock = self.file_storage.write().await;
+
+        // Finally store file in File Storage.
+        file_storage_write_lock
+            .insert_file_with_data(file_key, file_metadata.clone(), file_data_trie)
```
snowmead, line 136: So I only realized what this method was actually trying to do until I reached the end. I think we should rename this method to `upload_file_to_storage`.

snowmead (approved):
LGTM 👍🏼 Just had a couple of comments.

a-moreira:
@ffarall The RPC method is readily available when running the node as a provider, we can test it with JS or a simple CURL request (below).
However, using custom RPCs on the hosted PolkadotJS UI is not possible.  For that we would need to run our own instance and add the custom types there before booting it.

To test without running our own PolkadotJS UI instance:
```
curl --location 'localhost:9944/' --header 'Content-Type: application/json' --data '{
    "jsonrpc":"2.0",
    "method":"filestorage_loadFileInStorage",
    "params":["./README.md", "Test Location", "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY"],
    "id":1
}'
```
Or
```
import { ApiPromise, WsProvider } from '@polkadot/api';

async function main () {
	const api = await ApiPromise.create({
		provider: new WsProvider('ws://localhost:9944'),
	  rpc: {
	    filestorage: {
	      loadFileInStorage: {
	        description: 'Load file in local storage',
	        params: [
	          {
	            name: 'file_path',
	            type: 'String'
	          },
	          {
	            name: 'location',
	            type: 'String'
	          },
	          {
	            name: 'owner',
	            type: 'AccountId32'
	          }
	        ],
	        type: 'FileMetadata'
	      },
			}
		},
		types: {
			FileMetadata: {
				owner: 'AccountId32',
				location: 'Vec<u8>',
				size: 'u64',
				fingerprint: 'Vec<u8>'
			}
		}
	});

	const owner = '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY';
	const file_path = './README.md';
	const location = 'Test Location';

	const result = await api.rpc.filestorage.loadFileInStorage(file_path, location, owner);

	console.log(result);

}

main()
	.catch(console.error)
	.finally(() => process.exit(0));
```

client/rpc/src/lib.rs
ffarall: Once we incorporate MSPs and the concept of buckets, a `file_path` should be constructed from the `bucket-id/arbitrary/path`. At that point, if a file path is the same, as far as StorageHub is concerned, the user is replacing a file.

ffarall: For a BSP however, it could be useful to have a mapping from `file_key` to `file_location`. But that is out of scope for this "user side" RPC method.

ffarall (approved):
Very good step by step documentation of the code.

PR #74, opened 2024-05-23
feat: :sparkles: Add Full Network setup scripts and basic tests

timbrinded:
This is a long time coming, but we finally have a **one liner to launch a full network**.

In addition we have a separate script to configure that network to onboard SH-BSP as a provider via: `bun zombie:setup:native`

This is by no means "done", but this branch includes changes we would probably want to be merged so we can start building out the test suites in earnest.

---

Full list of changes:
- **NEW**: `fullNetworkSetup.ts` setup script
- **NEW**: `modifyPlainSpec.ts` for creating plainspecs with overrides (not yet usable though)
- **CHANGED**: Entrypoint for docker images
- **CHANGED**: Package/deps upgrades
- **CHANGED**: Test CI to additionally run zombie tests against full native network


snowmead (approved):
🔥

test/scripts/fullNetworkSetup.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  process.stdout.write(`Requesting sign up for ${accounts["sh-BSP"].sr25519.id} ...`);
+  await storageApi.tx.Providers.request_bsp_sign_up({
+    capacity: 5000000,
+    multiaddresses: [new Binary(uint8Array)],
+  }).signAndSubmit(accounts["sh-BSP"].sr25519.signer); // troubleshoot why this is not working
+  process.stdout.write("✅\n");
+
+  console.log("Waiting for randomness (9 blocks)...");
+  await waitForChain(shClient, { blocks: 9, timeoutMs: 120_000 });
```
snowmead, line 125: I assume we would be able to query BABE to get the exact amount of blocks required to wait before randomness is valid.

test/scripts/fullNetworkSetup.ts
```diff
@@ -0,0 +1,149 @@
...
+  // Enrolling BSP
+  const string = "0x8e6a748e6d787260f47f61df1e2cac065db8c1d41428eb178102177876071c6b";
+  const buffer = Buffer.from(string, "utf8");
+  const uint8Array = new Uint8Array(buffer);
+
+  process.stdout.write(`Requesting sign up for ${accounts["sh-BSP"].sr25519.id} ...`);
+  await storageApi.tx.Providers.request_bsp_sign_up({
+    capacity: 5000000,
+    multiaddresses: [new Binary(uint8Array)],
+  }).signAndSubmit(accounts["sh-BSP"].sr25519.signer); // troubleshoot why this is not working
```
snowmead, line 121: nitpick: would we can have `"sh-BSP"` be a const since it is used everywhere.

docker/storage-hub-node.Dockerfile
```diff
@@ -23,13 +23,13 @@ RUN apt-get update && \
...
+COPY --chown=storage-hub:storage-hub build/storage-hub-node /storage-hub/storage-hub-node
+RUN chmod uog+x /storage-hub/storage-hub-node
 
 # ws_port
 EXPOSE 9333 9944 30333 30334
 
-CMD ["storage-hub-node"]
+VOLUME ["/data"]
+
+ENTRYPOINT ["/storage-hub/storage-hub-node"]
```
snowmead, line 35: nice 👍🏼

test/scripts/fullNetworkSetup.ts
timbrinded: done

test/util/polkadot-api.ts
```diff
@@ -1,48 +1,96 @@
...
+        "best"
+      ).subscribe((value) => {
+        valueCount++;
+
+        if (!value) {
+          reject(new Error("Randomness value is undefined"));
+          return;
+        }
+
+        if (valueCount === 2) {
```
ffarall, line 29: Why do you need to wait for `valueCount === 2`? Is it because you're waiting 2 epochs for randomness? In our parachain we're just waiting for one epoch.

test/util/polkadot-api.ts
```diff
@@ -1,48 +1,96 @@
...
+          clearTimeout(timeout);
+          subscription.unsubscribe();
+          resolve(value[1]);
```
ffarall, line 32: What is the first item of `value` (`value[0]`)?

test/util/polkadot-api.ts
```diff
@@ -1,48 +1,96 @@
...
+  client: PolkadotClient,
+  options?: {
+    timeoutMs?: number;
+    blocks?: number;
+  }
+) => {
+  process.stdout.write(
+    `Waiting a maximum of ${
+      options?.timeoutMs || 60_000 / 1000
+    } seconds for ${await client._request("system_chain", [])} chain to be ready...`
```
ffarall, line 62: Any idea why this is giving me the error `Property '_request' does not exist on type 'PolkadotClient'.`

test/util/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from "./keyring";
 export * from "./polkadot-api";
+export * from "./constants";
```
ffarall, line 3: I see this file has no content... 🤔

test/scripts/fullNetworkSetup.ts
```diff
@@ -0,0 +1,148 @@
...
+  // relayWs: "wss://rococo-rpc.polkadot.io",
+  shWs: "ws://127.0.0.1:32000",
+});
+
+async function main() {
+  await waitForChain(relayClient);
+
+  // Check if executor parameters are set
+
+  const { executor_params } = await relayApi.query.Configuration.ActiveConfig.getValue();
```
ffarall, line 44: Here and in every other line that uses `relayApi`, I'm getting an error like this:
```
'storageApi.query.System.Account' is of type 'unknown'.
Property 'System' comes from an index signature, so it must be accessed with ['System'].
```

I presume it may be something related to my TS config, but just flagging it in case.

docker/latest-node-compose.yml
```diff
@@ -5,7 +5,6 @@ services:
       - "9944:9944"
     command:
       [
-        "storage-hub-node",
```
ffarall, line 8: How's this working without the command of the binary?

test/configs/fullNetworkNative.toml
```diff
@@ -1,32 +1,44 @@
...
 
 [[parachains.collators]]
 name = "sh-BSP"
 validator = false
+ws_port = 35000
 command = "../target/release/storage-hub-node"
 args = [
     "--no-hardware-benchmarks",
     "-lparachain=debug",
+    "--log bsp-volunteer-mock-task=debug",
```
ffarall, line 41: This mocked task is deprecated now on main

ffarall (approved):
Amazing job! Just a couple of comments that are more doubts for myself to clear out than anything else.

Aside from that:
1. When running `bun install` I get the `bun.lockb` file updated. I don't think that should be tha case, so flagging it.
2. Same with `storagehub.scale` when running `bun scalegen`.

docker/latest-node-compose.yml
timbrinded: Docker image entrypoint should point directly to the binary so you dont need to specify cmd explicitly.

This compose file wont work until PR is merged though as latest images (from `main` branch) aren't using the new entry point.

test/util/index.ts
timbrinded: That's correct, i had hard coded some values there but i no longer needed it. We'll need constants though saved somewhere so keeping it here as a reminder.

test/util/polkadot-api.ts
timbrinded: `_request` was added in a very recent version of `papi`. It might be you are executing against an old pkg ver

test/util/polkadot-api.ts
timbrinded: when you open a StorageValue subscription it returns an async `Observable` hook; to wait for the next epoch you need to have received two instances of data. This is because establishing a connection will immediately retrieve the current value and watch for updates.

test/util/polkadot-api.ts
timbrinded: `value` is `[FixedSizeBinary<32>, number] | undefined` ; which is papi's version of the rust type: Option<(H256,u32)>. So in context of the storage this would be presumably be the randomness.

test/configs/fullNetworkNative.toml
timbrinded: done

test/scripts/fullNetworkSetup.ts
timbrinded: When you merged the latest main, that results in a different binary to execute against - which changes the SCALE file. New scale file implies a new type interface, so everything breaks.

Latest commit should fix this which includes updated scale files 🤞

test/util/polkadot-api.ts
ffarall: Oh, I imagine this is the representation of what in our runtime code is returned by calling `random`, which is like you said `(H256, u32)` (with the `Option` 'cause you're looking at storage directly). With that in mind, why are you returning the `blockHeight` (second value of the tuple) in `waitForRandomness`, instead of the actual randomness (first value of the tuple)?

timbrinded: Fair enough, will just return both.

PR #75, opened 2024-05-23
Feat: blockchain transaction timeout

links234:
This PR refactors duplicate transaction handling code into a `SubmittedTransaction` which can be awaited for successful inclusion. On top of this a timeout is implemented.

### DIFF
--- a/node/src/services/blockchain/transaction.rs
+++ b/node/src/services/blockchain/transaction.rs
@@ -1,0 +1,92 @@
+use std::str::FromStr;
+
+use anyhow::anyhow;
+use log::*;
+use sp_core::H256;
+use storage_hub_infra::actor::ActorHandle;
+use tokio::sync::mpsc::Receiver;
+
+use crate::services::blockchain::{commands::BlockchainServiceInterface, types::ExtrinsicResult};
+
+use super::{types::ExtrinsicHash, BlockchainService};
+
+const LOG_TARGET: &str = "blockchain-transaction";
+
+#[derive(Debug)]
+pub struct SubmittedTransaction {
+    watcher: Receiver<String>,
+    hash: ExtrinsicHash,
+}
+
+impl SubmittedTransaction {
+    pub fn new(watcher: Receiver<String>, hash: H256) -> Self {
+        Self { watcher, hash }
+    }
+
+    pub async fn watch_for_success(
+        &mut self,
+        blockchain: &ActorHandle<BlockchainService>,
+    ) -> anyhow::Result<()> {
+        // Wait for the transaction to be included in a block.
+        let mut block_hash = None;
+        // TODO: Consider adding a timeout.
+        while let Some(result) = self.watcher.recv().await {
+            // Parse the JSONRPC string, now that we know it is not an error.
+            let json: serde_json::Value = serde_json::from_str(&result).map_err(|_| {
+                anyhow!("The result, if not an error, can only be a JSONRPC string; qed")
+            })?;
+
+            debug!(target: LOG_TARGET, "Transaction information: {:?}", json);
+
+            // Checking if the transaction is included in a block.
+            // TODO: Consider if we might want to wait for "finalized".
+            // TODO: Handle other lifetime extrinsic edge cases. See https://github.com/paritytech/polkadot-sdk/blob/master/substrate/client/transaction-pool/api/src/lib.rs#L131
+            if let Some(in_block) = json["params"]["result"]["inBlock"].as_str() {
+                block_hash = Some(H256::from_str(in_block)?);
+                let subscription_id = json["params"]["subscription"]
+                    .as_number()
+                    .ok_or_else(|| anyhow!("Subscription should exist and be a number; qed"))?;
+
+                // Unwatch extrinsic to release tx_watcher.
+                blockchain
+                    .unwatch_extrinsic(subscription_id.to_owned())
+                    .await?;
+
+                // Breaking while loop.
+                // Even though we unwatch the transaction, and the loop should break, we still break manually
+                // in case we continue to receive updates. This should not happen, but it is a safety measure,
+                // and we already have what we need.
+                break;
+            }
+        }
+
+        // Get the extrinsic from the block, with its events.
+        let block_hash = block_hash.ok_or_else(
+            || anyhow!("Block hash should exist after waiting for extrinsic to be included in a block; qed")
+        )?;
+        let extrinsic_in_block = blockchain
+            .get_extrinsic_from_block(block_hash, self.hash)
+            .await?;
+
+        // Check if the extrinsic was successful. In this mocked task we know this should fail if Alice is
+        // not a registered BSP.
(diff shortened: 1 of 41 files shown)


node/src/services/blockchain/transaction.rs
```diff
@@ -0,0 +1,150 @@
...
+use super::{types::ExtrinsicHash, BlockchainService};
+
+const LOG_TARGET: &str = "blockchain-transaction";
+
+#[derive(Debug)]
+pub struct SubmittedTransaction {
+    watcher: Receiver<String>,
+    hash: ExtrinsicHash,
+    timeout: Option<Duration>,
+}
```
a-moreira, line 23: It'd be good to have some doc comments, specially explaining that this `watcher` is the receiver side of a channel coming from the `RpcHandlers` response. 
The reason I'm asking is that there are a lot of indirections due to multiple channels of communication, and it's docs here would reduce the overhead of understanding what is coming from where.

node/src/services/blockchain/transaction.rs
```diff
@@ -0,0 +1,150 @@
...
+        loop {
+            // Get the elapsed time since submit.
+            let elapsed = start_time.elapsed();
+            // Calculate the remaining time to wait.
+            let remaining = match self.timeout {
+                Some(timeout) => {
+                    // Check if the timeout has been reached.
+                    if elapsed > timeout {
+                        return Err(anyhow!(
+                            "Timeout waiting for transaction to be included in a block"
```
a-moreira, line 58: Maybe add the tx hash to this error message to help with identifying the context.

ffarall: Agreed, it'd be useful to add it for all error messages.

node/src/services/blockchain/transaction.rs
```diff
@@ -0,0 +1,150 @@
...
+                        None => {
+                            // No timeout set, continue waiting.
+                            warn!(target: LOG_TARGET, "No timeout set and {:?} elapsed, continuing to wait for transaction to be included in a block.", NO_TIMEOUT_WARN);
+
+                            continue;
+                        }
+                    }
+                }
+            };
+            // Parse the JSONRPC string, now that we know it is not an error.
```
ffarall, line 92: This comment was left here, and now it is a bit misleading. 
```suggestion
            // Parse the JSONRPC string. The strings sent by the RPC wacher should be valid JSONRPC strings.
```

node/src/services/blockchain/transaction.rs
```diff
@@ -0,0 +1,150 @@
...
+                let subscription_id = json["params"]["subscription"]
+                    .as_number()
+                    .ok_or_else(|| anyhow!("Subscription should exist and be a number; qed"))?;
+
+                // Unwatch extrinsic to release tx_watcher.
+                blockchain
+                    .unwatch_extrinsic(subscription_id.to_owned())
+                    .await?;
+
+                // Breaking while loop.
```
ffarall, line 113:
```suggestion
                // Breaking loop.
```

node/src/services/blockchain/transaction.rs
```diff
@@ -0,0 +1,150 @@
...
+        // Get the extrinsic from the block, with its events.
+        let block_hash = block_hash.ok_or_else(
+            || anyhow!("Block hash should exist after waiting for extrinsic to be included in a block; qed")
+        )?;
+        let extrinsic_in_block = blockchain
+            .get_extrinsic_from_block(block_hash, self.hash)
+            .await?;
+
+        // Check if the extrinsic was successful. In this mocked task we know this should fail if Alice is
+        // not a registered BSP.
```
ffarall, line 130:
```suggestion
        // Check if the extrinsic was successful.
```

node/src/services/blockchain/transaction.rs
```diff
@@ -0,0 +1,150 @@
...
+            timeout: None,
+        }
+    }
+
+    pub fn with_timeout(mut self, timeout: Duration) -> Self {
+        self.timeout = Some(timeout);
+        self
+    }
+
+    pub async fn watch_for_success(
```
ffarall, line 41:
```suggestion
    /// Handles the lifecycle of a submitted transaction.
    ///
    /// Waits for the transaction to be included in a block.
    /// If the transaction is not included in a block within the specified timeout, it will be
    /// considered failed and an error will be returned.
    pub async fn watch_for_success(
```

ffarall (approved):
Great PR, great way of extracting this functionality

snowmead (approved):
Beast!

node/src/tasks/bsp_upload_file.rs
```diff
@@ -260,69 +242,21 @@ where
...
-        )?;
-        let extrinsic_in_block = self
-            .storage_hub_handler
+        self.storage_hub_handler
             .blockchain
-            .get_extrinsic_from_block(block_hash, tx_hash)
+            .send_extrinsic(call)
+            .await?
+            .with_timeout(Duration::from_secs(60))
+            .watch_for_success(&self.storage_hub_handler.blockchain)
```
snowmead, line 257: Why do you have to pass in blockchain handle when it is already accessible from `self` within the `watch_for_success`?

node/src/services/blockchain/transaction.rs
```diff
@@ -0,0 +1,150 @@
...
+const LOG_TARGET: &str = "blockchain-transaction";
+
+#[derive(Debug)]
+pub struct SubmittedTransaction {
+    watcher: Receiver<String>,
+    hash: ExtrinsicHash,
+    timeout: Option<Duration>,
+}
+
+const NO_TIMEOUT_WARN: Duration = Duration::from_secs(60);
```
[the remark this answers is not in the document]
a-moreira: AFAIU 60s is arbitrary. `timeout()` necessarily expects some value and this constant is a default one, so that after it passes, in case there is no timeout set (which would yield error), we just warn "ok, I waited 60s, I'm gonna try again" without waiting indefinitely at that point.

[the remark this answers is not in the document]
a-moreira: That is, we _will_ wait indefinitely if there is no timeout, but there will be warnings saying what's going on.

a-moreira (approved):
Great 🎉 ! Just incorporate Facu's (or yours) doc comments

node/src/services/blockchain/transaction.rs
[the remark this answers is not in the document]
a-moreira: @links234 maybe call the constant `NO_TIMEOUT_INTERVAL_WARNING`? If you don't think it's a mouthful 😆

node/src/services/blockchain/transaction.rs
links234: It is actually a `loop` not a `while` anymore.

node/src/tasks/bsp_upload_file.rs
links234: I didn't want to pollute `SubmittedTransaction`.

PR #76, opened 2024-05-23
feat: fetch latest nonce from chain state when initialising the BlockchainService

a-moreira:
Solves the issue of restarting the nonce after node initialisation.

### DIFF
--- a/node/src/services/blockchain/handler.rs
+++ b/node/src/services/blockchain/handler.rs
@@ -346,6 +346,7 @@ impl BlockchainService {
         debug!(target: LOG_TARGET, "Sending extrinsic to the runtime");
 
         // Get the nonce for the caller and increment it for the next transaction.
+        // TODO: Handle nonce overflow.
         let nonce = self.nonce_counter;
 
         // Construct the extrinsic.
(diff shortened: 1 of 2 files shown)


links234 (approved):
Very nice!

node/src/services/blockchain/handler.rs
```diff
@@ -272,12 +274,21 @@ impl BlockchainService {
         rpc_handlers: Arc<RpcHandlers>,
         keystore: KeystorePtr,
     ) -> Self {
+        // Get best nonce (not pending) for one of the pub keys (in this case the last one)
+        // in our vector of keys in the Keystore.
+        let current_block_hash = client.chain_info().best_hash;
```
ffarall, line 279: Although better than before, I believe this won't solve our problems definitively. This code executes before the node syncs the chain with its peers. Meaning that if there was a transaction for this `pub_key` after this node's last block recorded (which when initialising from scratch is 0), this won't return the correct nonce.

ffarall (changes requested):

node/src/services/blockchain/handler.rs
a-moreira: This is a good point. I changed it so we start with 0 and update the nonce after each new block notification is identified by the service.

ffarall (approved):
Great job!

PR #83, opened 2024-05-29
feat: Make Blockchain Service emit `AcceptedBspVolunteer` events

a-moreira:
- Makes BlockchainService emit `AcceptedBspVolunteer` events when the analogous event comes from the runtime.

- Removes unnecessary `.clone()`s (automatically generated by `cargo clippy --fix`).


node/src/services/blockchain/handler.rs
```diff
@@ -330,6 +335,42 @@ impl BlockchainService {
...
+                                multiaddresses,
+                                owner,
+                                size,
+                            },
+                        ) => {
+                            // We try to convert the types coming from the runtime into our expected types.
+                            let fingerprint: Fingerprint = fingerprint.as_bytes().into();
+                            // In this case, the Multiaddresses come as a BoundedVec of BoundedVecs of bytes,
+                            // and we need to convert them. It only fails if the bytes do not represent a valid multiaddress.
+                            // TODO: We should make sure Multiaddresses are valid from inside the runtime.
```
ffarall, line 352: We have a TODO for this because we considered it, but it is not easy to do that check in a runtime with no std apparently.

node/src/services/blockchain/handler.rs
```diff
@@ -330,6 +335,42 @@ impl BlockchainService {
...
+                            let multiaddresses: Vec<Multiaddr> = multiaddresses
+                                .into_iter()
+                                .map({
+                                    |bounded_vec| {
+                                        bounded_vec
+                                            .into_inner()
+                                            .try_into()
+                                            .expect("Malformed Multiaddress.")
```
ffarall, line 360: Do not `.expect` here. If it fails, it will crash the blockchain service (kind of a big deal). Instead, just handle the error by ignoring the event. If the BSP volunteering does not have a valid multiaddress, it's his fault.

node/src/services/blockchain/handler.rs
```diff
@@ -330,6 +335,42 @@ impl BlockchainService {
...
+                                    }
+                                })
+                                .collect();
+
+                            self.emit(AcceptedBspVolunteer {
```
ffarall, line 365: Shouldn't you filter by `owner` the events before emitting the internal event?

ffarall: You could add a condition to the `match` clause.

ffarall (changes requested):

node/src/services/blockchain/handler.rs
a-moreira: Yes, sorry, I shouldn't have opened the PR for review, it was supposed to be a draft. It's fixed now 😄

node/src/services/blockchain/handler.rs
a-moreira: You are right, we discussed this before. I'm now logging an error and just returning.

ffarall (approved):

PR #85, opened 2024-05-30
refactor: share types between BSP and runtime - part 1

links234:
In this PR:
1. Add `FileMetadata` in `FileKeyVerifier` crate and remove duplicate code
2. Share `FileMetadata`, `Fingerprint` and `Hash` between runtime/`FileKeyVerifier` and BSP
3. Fix RocksDB Forest Manager test


primitives/file-key-verifier/src/lib.rs
```diff
@@ -29,15 +29,135 @@ pub struct FileKeyVerifier<
...
+    }
+
+    pub fn chunks_to_check(&self) -> u64 {
+        self.size / SIZE_TO_CHALLENGES + (self.size % SIZE_TO_CHALLENGES != 0) as u64
+    }
+
+    pub fn chunks_count(&self) -> u64 {
+        self.size / CHUNK_SIZE + (self.size % CHUNK_SIZE != 0) as u64
+    }
+}
```
ffarall, line 135: Same here, I'm inclined to moving this to the bottom.

node/Cargo.toml
```diff
@@ -30,17 +30,20 @@ tokio = { workspace = true }
...
 shc-common = { workspace = true }
 shc-rpc = { workspace = true }
+shp-file-key-verifier = { workspace = true }
+storage-hub-traits = { workspace = true }
 
 clap = { workspace = true }
 log = { workspace = true }
 codec = { workspace = true }
-serde = { workspace = true }
+serde = { workspace = true, default-features = true }
```
ffarall, line 46: You removed the `default-features = false` in the root `Cargo.toml`. If you do that, wouldn't this be redundant?

client/common/src/types.rs
```diff
@@ -62,96 +59,15 @@ impl AsRef<[u8; 32]> for FileKey {
...
-
-    /// Decode metadata from the [SCALE](https://wentelteefje.github.io/parity-scale-codec-page/) encoded metadata bytes.
-    pub fn from_scale_encoded(bytes: Vec<u8>) -> Result<Self, codec::Error> {
-        decode_from_bytes(bytes.into())
-    }
-}
+pub type Hash = shp_file_key_verifier::Hash<H_LENGTH>;
+pub type Fingerprint = shp_file_key_verifier::Fingerprint<H_LENGTH>;
+pub type FileMetadata =
+    shp_file_key_verifier::FileMetadata<H_LENGTH, FILE_CHUNK_SIZE, FILE_SIZE_TO_CHALLENGES>;
```
ffarall, line 70: This types could use one or two lines of docs.

client/common/src/types.rs
```diff
@@ -62,96 +59,15 @@ impl AsRef<[u8; 32]> for FileKey {
...
-}
+pub type KeyVerifier = <Runtime as pallet_proofs_dealer::Config>::KeyVerifier;
+pub type FileKeyProof = <KeyVerifier as CommitmentVerifier>::Proof;
 
-impl Into<Hash> for Fingerprint {
-    fn into(self) -> Hash {
-        self.0
-    }
-}
+pub const H_LENGTH: usize = BlakeTwo256::LENGTH;
```
ffarall, line 65: How about defining this in `runtime/src/configs/mod.rs` and importing from there to here? In case we change hasher for instance (will hardly be the case, but since we're at it...)

ffarall (approved):
Approved with comment on docs and code organisation.

Please merge from main and solve conflicts here.

primitives/file-key-verifier/Cargo.toml
```diff
@@ -18,6 +18,7 @@ targets = ["x86_64-unknown-linux-gnu"]
 codec = { workspace = true }
 num-bigint = { workspace = true }
 scale-info = { workspace = true }
+serde = { workspace = true }
```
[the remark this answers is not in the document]
links234: Shouldn't it be fine if it compiles? In workspace, `serde` is `default-features=false` and I manually enable them where I need and not in runtime.

node/Cargo.toml
links234: I removed it in root so that we can import it in runtime with `no-std`.

PR #90, opened 2024-06-04
feat: Apply delta to Forest proof partial trie

snowmead:
The `apply_delta` function in the new `ProofDeltaApplier` interface used by the runtime to apply mutations (delta) to a partial trie constructed from a proof. 

This enables the runtime to compute a Provider's new Forest root by either adding or removing file keys to the partial trie. The runtime will require Providers to construct a proof of non-inclusion or inclusion when they are adding or removing file keys from the Forest respectively.

Features:

- New `ProofDeltaApplier` interface.
- `ForestVerifier` implements `ProofDeltaApplier`.
- File system extrinsics `bsp_confirm_storing` & `bsp_stop_storing` updates the forest root by applying the delta to the proof submitted.
- Proof Dealer pallet now applies any mutations required from the checkpoint challenges (priority challenges) which now include an optional mutation for each challenged key.


primitives/file-key-verifier/src/lib.rs
```diff
@@ -71,7 +71,7 @@ where
     /// and checks if the resulting leaf is in the proof.
     fn verify_proof(
         expected_file_key: &Self::Commitment,
-        challenges: &[Self::Challenge],
+        challenges: &[(Self::Challenge, Option<ChallengeKeyInclusion>)],
```
snowmead, line 74: We probably don't need to add this here since there will probably be no case for checking a chunk being included or unincluded in a proof. We simply check if the challenge has been responded to appropriately with a _valid_ (based on our proof verifying algorithm) proof.

pallets/file-system/src/utils.rs
```diff
@@ -551,6 +559,7 @@ where
             // Issue a challenge to force the BSPs to update their storage root.
             <T::ProofDealer as shp_traits::ProofsDealerInterface>::challenge_with_priority(
                 &file_key,
+                ChallengeKeyInclusion::NotIncluded,
```
snowmead, line 562: Since we are removing a file key (deleting a storage request), we could force providers to generate a proof of non-inclusion for the given file key. This would force all providers to delete the file key from their forest to successfully respond to the challenge. 

This would not give the runtime the possibility of applying the delta to the Forest itself and updating the root. What we can do potentially if we want to be consistent (having the runtime apply the delta) is to specify add an extra flag to the challenge that states whether or not the file should be deleted or added (basically adding a mutation as well to the challenge). This would mean we specify `None` instead of `ChallengeKeyInclusion::NotIncluded` and whenever a provider submits a proof that includes the file key, the runtime will apply the delta and update the root.

The challenges (including priority ones) would have a storage type that resembles something like this:

```rust
type Challenges = (Key, Option<ChallengeKeyInclusion>, Delta)
```

primitives/file-key-verifier/src/lib.rs
links234: I would try to not change this interface if possible. If this can be handled by the upper layer using it, we should keep this simpler.

snowmead: The only thing I changed was the return type from Vec to BTreeSet

links234 (changes requested):
Great work! Just some nitpick for duplicated code.

As discussed on slack, I think for now we should add a shared (between BSP and client) `get_chunks_to_challenge(file_key_challenge, chunks: i32)` with a hardcoded (for now) number of chunks just so that we can get the ball rolling until we decide on the exact algorithm.

ffarall (changes requested):

pallets/proofs-dealer/src/utils.rs
```diff
@@ -182,22 +181,60 @@ where
...
+        // Apply the delta to the Forest root for all mutations that are in checkpoint challenges.
+        if let Some(challenges) = checkpoint_challenges {
+            // Aggregate all mutations to apply to the Forest root.
+            let mutations: Vec<_> = challenges
+                .iter()
+                .filter_map(|(key, mutation)| match mutation {
+                    Some(Mutation::Remove) if forest_keys_proven.contains(key) => {
+                        Some((*key, Mutation::Remove))
+                    }
+                    Some(Mutation::Remove) | Some(Mutation::Add) | None => None,
```
ffarall, line 213: Technically, it shouldn't be possible to have a key with `Remove` mutation, that IS NOT in `forest_keys_proven`. Because that would mean that a checkpoint challenge has not been verified, therefore the verification should have failed. Basically this is an impossible state, and we should leave that clear in the code. I suggest something similar to what we do in `expect_or_err`, but instead of erroring out, we just return `None` (but in a test build, we panic).

pallets/proofs-dealer/src/utils.rs
```diff
@@ -182,22 +181,60 @@ where
...
+                        forest_proof,
+                    )
+                    .map(|(_, new_root)| new_root)
+                    .map_err(|_| Error::<T>::FailedToApplyDelta)
+                })?;
+
+                // Update root of BSP after all mutations have been applied to the Forest.
+                <T::ProvidersPallet as shp_traits::MutateProvidersInterface>::change_root_bsp(
+                    *submitter, new_root,
+                )?;
```
ffarall, line 234: This pallet is abstracted over the concept of a BSP.

pallets/proofs-dealer/src/utils.rs
```diff
@@ -237,14 +274,17 @@ where
 
     /// Add challenge to ChallengesQueue.
     ///
+    /// Keys that are challenged manually are not random and therefore the key inclusion should be specifed.
```
ffarall, line 277:
```suggestion
    /// Keys that are challenged manually are not random and therefore the key inclusion should be specified.
```

pallets/proofs-dealer/src/utils.rs
```diff
@@ -237,14 +274,17 @@ where
...
+    /// The challenges generated randomly from a seed do not have a specified inclusion type.
```
ffarall, line 278: Is this comment accurate? We didn't end up adding inclusion/uninclusion fields

pallets/proofs-dealer/src/utils.rs
```diff
@@ -305,6 +348,7 @@ where
                 .concat(),
             );
 
+            // The challenge inclusion type is None since we are generating random challenges and don't expect proofs of inclusion or non-inclusion.
```
ffarall, line 351: Also here, I think this comment is unnecessary

pallets/proofs-dealer/src/tests.rs
```diff
@@ -430,13 +430,15 @@ fn proofs_dealer_trait_challenge_with_priority_succeed() {
         let file_key = BlakeTwo256::hash(b"file_key");
 
         // Challenge using trait.
-        <ProofsDealer as shp_traits::ProofsDealerInterface>::challenge_with_priority(&file_key)
-            .unwrap();
+        <ProofsDealer as shp_traits::ProofsDealerInterface>::challenge_with_priority(
```
ffarall, line 433: Shouldn't we add tests for the `apply_delta` functionalities added in this PR?

pallets/file-system/src/utils.rs
```diff
@@ -751,6 +810,27 @@ where
         T::ThresholdType::decode(&mut &xor_result[..])
             .map_err(|_| Error::<T>::FailedToDecodeThreshold)
     }
+
+    /// Compute file key from metadata.
+    pub(crate) fn compute_file_key(
```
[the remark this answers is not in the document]
ffarall: Agreed

pallets/file-system/src/utils.rs
```diff
@@ -445,16 +442,28 @@ where
...
+            )?;
+
+        // Ensure that the file key IS NOT part of the BSP's forest.
+        // Note: The runtime is responsible for adding and removing keys, computing the new root and updating the BSP's root.
+        ensure!(
+            !proven_keys.contains(&file_key),
+            Error::<T>::ExpectedNonInclusionProof
+        );
 
         // TODO: Generate challenges for the key proof properly.
```
ffarall, line 468: My original idea was that this could be as simple as using the `file_key` as a "seed" for generating the key challenges. After all, it has to be something that the provider knows beforehand to be able to respond to it. Although not perfect, I think it is good enough, specially for a first version at least.

pallets/file-system/src/utils.rs
```diff
@@ -679,9 +707,40 @@ where
...
-        // TODO: compute new root from proof and update the storage root of bsp.
+        // Verify the proof of inclusion.
+        let proven_keys =
+            <T::ProofDealer as shp_traits::ProofsDealerInterface>::verify_forest_proof(
+                &bsp_id,
+                &[file_key],
+                &inclusion_forest_proof,
+            )?;
+
+        // TODO: Use BTreeSet instead for faster lookups.
```
ffarall, line 718: Aren't you already using it?

pallets/file-system/src/tests.rs
```diff
@@ -1041,10 +1041,10 @@ fn bsp_confirm_storing_success() {
             location.clone(),
             H256::zero(), // TODO construct a real proof
             ForestProof {
-                encoded_nodes: vec![vec![0]],
+                encoded_nodes: vec![H256::default().as_ref().to_vec()],
```
ffarall, line 1044: Shouldn't we have some tests for the added functionality as well?

ffarall (changes requested):

pallets/file-system/src/utils.rs
snowmead: Will be adding this in another pr for simplicity.

pallets/proofs-dealer/src/utils.rs
snowmead: Modified the types and storage for the proofs dealer pallet to only accept RemoveMutation for checkpoint challenges.

pallets/proofs-dealer/src/lib.rs
```diff
@@ -44,17 +44,25 @@ pub mod pallet {
...
         /// To check if whoever submits a proof is a registered Provider.
         type ProvidersPallet: ProvidersInterface<
-            AccountId = Self::AccountId,
-            MerkleHash = Self::MerkleTrieHash,
-            Balance = Self::NativeBalance,
-        >;
+                AccountId = Self::AccountId,
+                MerkleHash = Self::MerkleTrieHash,
+                Balance = Self::NativeBalance,
+            > + shp_traits::MutateProvidersInterface<
```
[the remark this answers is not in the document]
snowmead: Not really sure how you want to split this then. In the end you need a function to change the root of bucket or change the root of a bsp. The Mutate trait actually has no mention of a bsp, except for the function `change_bsp_root`, which I can rename to `change_provider_global_root`, to infer that that the provider should have some global root associated to them. And keep the `change_bucket_root` separate.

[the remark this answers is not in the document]
snowmead: If you attempt to change a root of an MSP by calling `change_provider_global_root` it will fail with `NoGlobalRootForProvider`. Meaning it is not a BSP in our case.

primitives/forest-verifier/src/tests.rs
```diff
@@ -223,1153 +224,1433 @@ pub fn merklise_file<T: TrieLayout>(file_path: &str) -> (MemoryDB<T::Hash>, Hash
...
-        .expect("Failed to verify proof");
+        // This recorder is used to record accessed keys in the trie and later generate a proof for them.
+        let recorder: Recorder<BlakeTwo256> = Recorder::default();
 
-    assert_eq!(proof_keys, vec![leaf_keys[0]]);
-}
+        let largest_key = leaf_keys
+            .iter()
+            .max()
+            .map(|key| (*key, None::<Mutation>))
```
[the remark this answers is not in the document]
snowmead: I don't want the test to be static and break if we change the order of keys when we build the forest.

primitives/forest-verifier/src/tests.rs
```diff
@@ -223,1153 +224,1433 @@ pub fn merklise_file<T: TrieLayout>(file_path: &str) -> (MemoryDB<T::Hash>, Hash
...
-        let mut trie_recorder = recorder.as_trie_recorder(root);
-        let trie = TrieDBBuilder::<LayoutV1<BlakeTwo256>>::new(&memdb, &root)
-            .with_recorder(&mut trie_recorder)
-            .build();
+        let challenge_key = generate_unique_key(&leaf_keys);
+        let mutations: Vec<(H256, Mutation)> = vec![(challenge_key, Mutation::Add)];
 
-        // Create an iterator over the leaf nodes.
-        let mut iter = trie.into_double_ended_iter().unwrap();
+        let expected_root = apply_delta(memdb.clone(), root.clone(), mutations.clone());
```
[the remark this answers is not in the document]
snowmead: I removed it, it is useless to test for the expected root. I verify anyways that the keys are either removed or added from the trie.

links234 (approved):
LGTM!

primitives/file-key-verifier/src/consts.rs
```diff
@@ -0,0 +1,13 @@
+use sp_core::Hasher;
```
ffarall, line 1: Not sure about putting these constants here. This is the `file-key-verifier` crate, something abstracted over these constants. This seem more suited to be in the runtime, that makes the `FileKeyVerifier` struct concrete by setting these constants, to what it makes sense in the context of the StorageHub runtime.

primitives/forest-verifier/src/tests.rs
```diff
@@ -223,1153 +226,1524 @@ pub fn merklise_file<T: TrieLayout>(file_path: &str) -> (MemoryDB<T::Hash>, Hash
...
+                Err(err) => err,
+                Ok(_) => panic!("Expected error."),
+            };
+
+        assert_eq!(
+            err,
+            DispatchError::Other(
+                "Failed to convert proof to memory DB, root doesn't match with expected."
+            )
+        );
```
ffarall, line 1747: I think you could have used `assert_err!` macro for this.

primitives/forest-verifier/src/lib.rs
```diff
@@ -204,6 +204,64 @@ where
             }
         }
 
-        return Ok(Vec::from_iter(proven_keys));
+        return Ok(proven_keys);
+    }
+}
+
+// TODO: rename to TrieDeltaApplier
```
ffarall, line 211: Was this left dangling? Maybe replace for a doc comment con this implementation?

ffarall (approved):

primitives/forest-verifier/src/tests.rs
snowmead: Ah thank you!

primitives/file-key-verifier/src/consts.rs
snowmead: This was a discussion I had with @links234 to determine how we can unify our common types which the client and runtime depend on to not differ in implementations.

PR #91, opened 2024-06-06
refactor: Unify shared runtime and client types part 2

links234:
In this PR:
- Move `FileKey`, `Leaf`, `ChunkId`, and `Chunk` to runtime.
- `ChunkId` implements `as_trie_key` and `from_trie_key`.
- `FileStorage` interface now generates proofs for `FileKeyProof` (from runtime) instead of `FileProof`. Note: `FileProof` is kept for `FileDataTrie` still needed for the first steps of uploading a file.
- `FileProof` can be built into a `FileKeyProof` by adding the associated `FileMetadata`.
- `FileStorage` now uses `ChunkId` to/from trie keys, compatible with the runtime (`AsCompact` instead of `be_bytes`).
- Ditch `bincode` for serialize/deserialize of protobuf fields in favor of SCALE.
- Add `verify()` and `proven()` to `FileKeyProof`.
- Received file chunks proofs are now being verified.

### DIFF
--- a/primitives/file-key-verifier/src/lib.rs
+++ b/primitives/file-key-verifier/src/lib.rs
@@ -1,6 +1,7 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
-use codec::{Decode, Encode};
+use codec::{Compact, Decode, Encode};
+use core::fmt::Debug;
 use frame_support::sp_runtime::DispatchError;
 use num_bigint::BigUint;
 use scale_info::TypeInfo;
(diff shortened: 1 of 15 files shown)


primitives/file-key-verifier/src/lib.rs
```diff
@@ -49,6 +69,62 @@ impl<const H_LENGTH: usize, const CHUNK_SIZE: u64, const SIZE_TO_CHALLENGES: u64
...
+                .map_err(|e| ProvenFileKeyError::ChunkIdFromKeyError(e))?;
+            let chunk = trie
+                .get(&key)
+                .map_err(|_| ProvenFileKeyError::FailedToGetTrieValue)?
+                .ok_or_else(|| ProvenFileKeyError::KeyNotFoundInTrie)?;
+            proven.push(Leaf::new(chunk_id, chunk));
+        }
+
+        Ok(proven)
+    }
```
ffarall, line 127: This feels out of place here.
1. Why is there a `verify` function for `FileKeyProof` when there's also the `verify_proof` function for `FileKeyVerifier`?
2. What is the purpose of `proven`? Seems to be almost like an "extension" of `verify` that just returns the keys that are part of the proof.
3. Generally speaking, this crate was scoped as an implementation of `CommitmentVerifier` for the key verifier needed for the runtime. Bringing in a bunch of type definitions is one thing, but it seems that you also braught in some extra functionality that looks repetitive and it's purpose is not clear.

It feels like this file used to have a clear, concise role to fulfill, and that is blurred out completely now.

ffarall (changes requested):
Overall great job. Just those `verify` and `proven` functions in the `file-key-verifier` that feel out of place.

ffarall (approved):

PR #94, opened 2024-06-17
fix: 🐛 Debug file upload flow

ffarall:


node/src/tasks/bsp_upload_file.rs
```diff
@@ -328,12 +328,12 @@ where
     async fn on_file_complete(&self, file_key: &HasherOutT<T>) {
         info!(target: LOG_TARGET, "File upload complete ({:?})", file_key);
 
-        // Unregister the file from the file transfer service.
-        self.storage_hub_handler
-            .file_transfer
-            .unregister_file(file_key.as_ref().into())
-            .await
-            .expect("File is not registered. This should not happen!");
```
snowmead, line 306: Why did you comment this?

client/file-manager/src/in_memory.rs
```diff
@@ -154,7 +163,8 @@ where
             .as_str(),
         );
 
-        if metadata.chunks_count() != file_data.stored_chunks_count() {
+        let stored_chunks = file_data.stored_chunks_count()?;
+        if metadata.chunks_count() != stored_chunks {
```
snowmead, line 167: Why do we check this? Wouldn't we check initially after adding a file to storage initially when received from a user so we don't have to check everytime aftewards?

snowmead (approved):

node/src/tasks/bsp_upload_file.rs
ffarall: It was throwing an error and I wanted it to just get to the end. We can fix it in another PR, but I would want to merge this for Tim.

client/file-manager/src/in_memory.rs
ffarall: I guess given the asynchronous nature of this processes, the extra check doesn't hurt.

snowmead: It does because if you have large files, if you have to iterate through the whole thing every time you need to generate a proof for it.

ffarall: Fair point. We can take it into consideration as an optimisation, at the risk of inconsistencies.

PR #99, opened 2024-06-26
feat: ✨ Per-block mocked randomness in dev mode

ffarall:


node/src/service.rs
```diff
@@ -608,7 +609,13 @@ where
 							slot_duration,
 						);
 
-                        let additional_keys = vec![(para_head_key, para_head_data), (relay_slot_key, Slot::from(u64::from(*relay_slot)).encode())];
+                        let current_block_randomness = BlakeTwo256::hash(timestamp.encode().as_slice());
+
+                        let additional_keys = vec![
+                            (para_head_key, para_head_data),
+                            (relay_slot_key, Slot::from(u64::from(*relay_slot)).encode()),
```
snowmead, line 616: Is there a reason why we need to include these keys? I thought we would only need to add  `current_block_randomness_key`.

snowmead (approved):

timbrinded (approved):
Short but sweet 👍

node/src/service.rs
ffarall: Those were there before

snowmead: oh yes I remember now

PR #100, opened 2024-06-26
Feat: BSP wait for threshold

links234:
This PR adds:
 1. Add `wait_for_block` to BlockchainService which will await for a given block number
 2. Add `get_node_public_key` to BlockchainService which returns the node's public key, needed for threshold computation
 3. Add runtime-api for pallet-file-system to expose shared (BSP & node) protocol computations
 4. Add `query_file_earliest_volunteer_block` runtime-api for file-system pallet and BlockchainService counterpart
 5. BSPs now wait for the earliest block which they can volunteer before sending the transaction


pallets/file-system/src/utils.rs
```diff
@@ -57,7 +61,84 @@ macro_rules! expect_or_err {
...
+                    return Err(Error::<T>::ThresholdArithmeticError.into());
+                }
+            };
+
+        // Compute the block number at which the BSP should send the volunteer request.
+        let blocks_to_wait: U256 = blocks_to_wait.into_u256();
+        let volunteer_block_number = storage_request_block.saturating_add(
+            BlockNumberFor::<T>::try_from(blocks_to_wait)
+                .map_err(|_| Error::<T>::FailedToConvertBlockNumber)?,
+        );
```
ffarall, line 137: This way of converting is fine and will work. If you're interested though, there is another way to do it without intermediately casting to a larget number type. You can find it [here](https://github.com/Moonsong-Labs/storage-hub/blob/main/pallets/proofs-dealer/src/utils.rs#L532). Basically it uses the inner type `StakeToBlockNumber` which has to implement the trait `Convert<BalanceFor<Self>, BlockNumberFor<Self>>`

pallets/file-system/src/utils.rs
```diff
@@ -57,7 +61,84 @@ macro_rules! expect_or_err {
...
+                .try_into()
+                .map_err(|_| QueryFileEarliestVolunteerBlockError::FailedToEncodeBsp)?,
+        )
+        .map_err(|_| QueryFileEarliestVolunteerBlockError::ThresholdArithmeticError)?;
+
+        // Compute the block number at which the BSP should send the volunteer request.
+        let volunteer_block_number =
+            match Self::compute_volunteer_block_number(bsp_threshold, storage_request_block) {
+                Ok(Some(volunteer_block_number)) => volunteer_block_number,
+                _ => Zero::zero(),
```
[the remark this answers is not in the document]
ffarall: From what I see in `compute_volunteer_block_number`, `Ok(None)` would mean "volunteer in the current block". `Err(_)` would mean "I don't know when you should volunteer", so probably best to return an error.

pallets/file-system/src/utils.rs
```diff
@@ -792,7 +872,7 @@ impl<T: pallet::Config> Pallet<T> {
     }
 
     /// Calculate the XOR of the fingerprint and the BSP.
-    fn compute_bsp_xor(
+    pub fn compute_bsp_xor(
```
ffarall, line 875: This doesn't seem to be used outside of this crate, is this needed?

pallets/file-system/src/lib.rs
```diff
@@ -715,4 +718,26 @@ pub mod pallet {
             total_used_weight
         }
     }
+
+    pub trait IntoU256 {
```
ffarall, line 722: You would avoid this with an inner type that implements the `Convert` trait.

pallets/file-system/runtime-api/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "pallet-file-system-runtime-api"
+description = "Pallet exposing storage hub protocol related helper runtime API calls."
```
ffarall, line 3:
```suggestion
description = "Crate exposing StorageHub protocol related helper runtime API calls."
```

node/src/tasks/bsp_upload_file.rs
```diff
@@ -252,6 +252,33 @@ where
...
+            .storage_hub_handler
+            .blockchain
+            .query_file_earliest_volunteer_block(
+                node_public_key,
+                H256::from_slice(file_key.as_ref()),
+            )
+            .await
+            .map_err(|e| anyhow!("Failed to query file earliest volunteer block: {:?}", e))?;
+
+        // If earliest_volunteer_block is 0 it means that we can volunteer right away.
```
ffarall, line 273: I would change this behaviour so that if the node can volunteer right away, it returns the current block instead of 0. Otherwise, you could make it so that it returns how many blocks you'd have to wait, then 0 would mean now, but that would be a bigger change.

node/src/tasks/bsp_upload_file.rs
```diff
@@ -252,6 +252,33 @@ where
...
+        if earliest_volunteer_block > U256::zero() {
+            // TODO: if the earliest block is too far away, we should drop the task.
```
ffarall, line 275: How about adding a timout here? Cause it could be either that the block is too far away, or the chain gets stalled and too much time passes by, even if it could have volunteered in a short number of blocks.

ffarall (changes requested):
Excellent code, specially in the blockchain service modifications. I only suggested a few improvements.

Aside from that, I don't see any integrations tests to test this. Given that you need this for the next task that is the confirmation, I'm ok with merging this without further testing, but it should then be tested alongside the confirmation step.

ffarall (approved):

PR #101, opened 2024-06-27
docs: 📝 Proofs Dealer pallet docs

ffarall:


pallets/proofs-dealer/README.md
```diff
@@ -0,0 +1,120 @@
+# Proofs Dealer Pallet
+
+## Overview
+
+The Proofs Dealer pallet is designed for as a very agnostic, unopinionated, highly configurable pallet for sending out challenges to Providers, verifying proof submissions and enforcing punishment for Providers who fail to respond to challenges.
```
TDemeco, line 5:
```suggestion
The Proofs Dealer pallet is designed as a very agnostic, unopinionated, highly configurable pallet for sending out challenges to Providers, verifying proof submissions and enforcing punishment for Providers who fail to respond to challenges.
```

pallets/proofs-dealer/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+This pallet introduces the concept of "Merkle Patricia Forests", which are none other than Merkle Patricia tries, whose leaves are themselves commitments. These commitment leaves, can also take the form of a Merkle Patricia trie roots for example, hence the name Merkle Patricia "Forest". Each leaf in the Merkle Patricia Forest is considered a commitment to providing service for something. That can be, for example, a file that a Provider committed to storing (in this scenario, Providers would be _Storage Providers_). But this pallet is abstracted over that, and the service that Providers are offering can be anything, as long as they can generate a cryptographic commitment to it. A "Provider" can be any entity that is identified by a Merkle Patricia Forest root. Such root is not stored by this pallet, but rather by another hypothetical "Providers" pallet, which must implement the [`ProvidersInterface`](../../primitives/traits/src/lib.rs) trait.
```
TDemeco, line 7:
```suggestion
This pallet introduces the concept of "Merkle Patricia Forests", which are none other than Merkle Patricia tries whose leaves are themselves commitments. These commitment leaves can also take the form of, for example, a Merkle Patricia trie root, hence the name Merkle Patricia "Forest". Each leaf in the Merkle Patricia Forest is considered a commitment to providing service for something. That can be, for example, a file that a Provider committed to storing (in this scenario, Providers would be _Storage Providers_). But this pallet is abstracted over that, and the service that Providers are offering can be anything, as long as they can generate a cryptographic commitment to it. A "Provider" can be any entity that is identified by a Merkle Patricia Forest root. Such root is not stored by this pallet, but rather by another hypothetical "Providers" pallet, which must implement the [`ProvidersInterface`](../../primitives/traits/src/lib.rs) trait.
```

pallets/proofs-dealer/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Ticks vs Blocks
+
+One important distinction that needs to be made is that the challenges are distributed in "ticks". Ticks are like blocks, except that the tick counter is skipped during Multi Block Migrations. A `ChallengesTicker` counter is incremented on every block, when the `on_poll` hook is called. `on_poll` is always called at the beginning of every block, except during Multi Block Migrations, and the challenges ticker is always incremented by 1. Therefore if there are no Multi Block Migrations, the challenges ticker will always be equal to the block number.
+
+The reason for this distinction can be found in the [PR](https://github.com/paritytech/polkadot-sdk/pull/1781) that introduced the poll hook. During Multi Block Migrations, `on_poll` is not called, therefore the challenges ticker wouldn't be incremented. That is a desirable behaviour, since during Multi Block Migrations, Providers wouldn't have the opportunity to submit proofs. By not incrementing the challenges ticker, we ensure that no Provider is slashed during Multi Block Migrations.
+
+![Substrate block lifecycle](./resources/blockLifecycle.png)
```
TDemeco, line 25: A bit repetitive on Multi Block Migrations, maybe it makes sense to abreviate it after the first use? Something like MBM

pallets/proofs-dealer/README.md
```diff
@@ -0,0 +1,120 @@
...
+            seed.encode().as_ref(),
+            provider_id.encode().as_ref(),
+            i.encode().as_ref(),
+        ]
+        .concat(),
+    );
+}
+```
+
+At the end of the tolerance period, if the Provider has not submitted a proof for the challenges it is supposed to respond to, it is marked as "slashable". How that slashing is implemented is out of scope for this pallet.
```
TDemeco, line 48: Comment unrelated to docs but regarding this logic: have we thought about how we could handle when a provider should be slashed multiple times? Since we only mark it as slashable, if between two non-delivered proof the slashing system doesn't get to that provided, it could avoid being double-slashed as it should. Maybe it makes sense to keep a counter of not provided proofs instead?

TDemeco (approved):
Really great docs! 👏🏻 I particularly liked the use of the "ticks" term, very creative.

pallets/proofs-dealer/README.md
ffarall: Yes, I've given it a thought and it can be achieved by something like you suggested, of keeping a counter. However I don't know if it is worth it to continue to complicate an already very complex pallet, given that this could also be solved by adding an offchain worker that just sends the slash transaction whenever there are Providers marked as slashable. After all, StorageHub collators would be incentivised to enforce slashing if that slashing goes to a treasury, which in turn goes to pay the collators.

PR #103, opened 2024-06-27
feat: ✨ Add BSP helper fns and tests

timbrinded:
This PR adds a new AddBSP function to our tests, so we can optionally create scenarios where multiple BSPs are all providing their services to the network.


test/util/bspNet/docker.ts
```diff
@@ -33,3 +37,98 @@ export const showContainers = () => {
     console.log("Error displaying docker containers");
   }
 };
+
+export const addBspContainer = async (options?: {
```
ffarall, line 41: Would be extremely useful to be able to pass additional command line arguments to the BSP to add. For example, we're gonna have to do integration tests for the RocksDB implementation of our storage layer. That requires having a BSP that was spawned with `--storage-layer rocks-db`

ffarall (approved):
Great job Tim! Just one comment:

Running `pnpm typegen` generates some changes and most importantly, it ends with error:
```
  ✖ Found an unknown key `trailingCommas`.
  
    19 │   "javascript": {
    20 │     "formatter": {
  > 21 │       "trailingCommas":"none",
       │       ^^^^^^^^^^^^^^^^
    22 │       "semicolons": "always",
    23 │       "indentStyle": "space",
  
  ℹ Known keys:
  
  - jsxQuoteStyle
  - quoteProperties
  - trailingComma
  - semicolons
  - arrowParentheses
  - bracketSpacing
  - bracketSameLine
  - enabled
  - indentStyle
  - indentWidth
  - lineEnding
  - lineWidth
  - quoteStyle
  - attributePosition
  

configuration ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Biome exited because the configuration resulted in errors. Please fix them.
```

test/util/bspNet/docker.ts
timbrinded: Trailing commas error is because you are running old biome - the new key was introduced in biome 1.8.2 (to replace the old key trailingComma).

Updating your `pnpm` pkgs should fix this for you.

timbrinded: Added additional Args parameter to helper as an option

PR #105, opened 2024-07-01
feat: :sparkles: make bsps sign up with `None` root

TDemeco:
This PR:
- Makes it so, instead of signing up with a 0 root, BSPs sign up with a `None` root, which makes it easier to detect which BSPs have not stored any files yet and prevents unwanted logic to run for them.


TDemeco:
Closed since we decided to take another approach to detect SPs that haven't yet stored a file.

PR #109, opened 2024-07-01
feat: ✨ Initialise Provider's challenge cycle in first file added to Forest

ffarall:
In this PR:
1. Add `initialise_challenge_cycle` to `ProoofsDealerInterface`. This exposes de functionality to initialise the cycle of challenges for a Provider, and enforcing challenges via slashing.
2. Implement it in Proofs Dealer pallet.
3. Use it in `bsp_confirm_storing` when a BSP is confirming storing its very first file.


pallets/proofs-dealer/src/utils.rs
```diff
@@ -689,10 +689,63 @@ impl<T: pallet::Config> ProofsDealerInterface for Pallet<T> {
         )
     }
 
-    // TODO: Add `initialise_provider` method to be called by the FileSystem pallet.
-    // TODO: when a file is first uploaded to a BSP or bucket.
-    // TODO: It would set `LastTickProviderSubmittedProofFor` to the current tick and
-    // TODO: the deadline for submitting a proof in `ChallengeTickToChallengedProviders`.
+    fn initialise_provider(who: &Self::ProviderId) -> DispatchResult {
```
snowmead, line 692: Maybe we should rename this to `start_providing`. I wouldn't have guessed that the ticks for challenges was affected here in any way.

pallets/file-system/src/utils.rs
```diff
@@ -548,8 +548,32 @@ where
...
 
-        // TODO: Check if this is the first file added to the BSP's Forest. If so, initialise
-        // TODO: last block proven by this BSP accordingly.
+        // Check if this is the first file added to the BSP's Forest. If so, initialise last block proven by this BSP.
+        let old_root = expect_or_err!(
+            <T::Providers as shp_traits::ProvidersInterface>::get_root(bsp_id),
+            "Failed to get root for BSP, when it was already checked to be a BSP",
+            Error::<T>::NotABsp
+        );
+        if old_root == Default::default() {
```
snowmead, line 557: Ok so `old_root` can be any type implementing Default. And the file system pallet assumes that if the value is default, then it is the first file. 

I was thinking of `Some(root)` instead. So if it is `None`, it means it is the first file.

snowmead (approved):

pallets/proofs-dealer/src/utils.rs
ffarall: How about `initialise_provider_challenge_cycle` or `initialise_challenge_cycle`?

PR #110, opened 2024-07-01
chore: Improve RocksDB File Manager implementation

a-moreira:
This is part 2 of #96, continuing the implementation of RocksDB with some improvements:

- Removes `Backend<T>` trait, which was created so we could use a `MockStorageDb` in unit tests for File and Forest managers. Now we use `kvdb_memorydb::InMemory` instead, which makes the code simpler.
- Removes unnecessary `RwLock<_>` from the File Storage.
- Removes the now obsolete `nextest.toml` config file, along with the `serial_test` crate, both used to run the `FileStorage` tests in sequence instead of in parallel.
- Adds new method `new_empty_file_data_trie` to `FileStorage` trait so we can create a new empty file trie using the same underlying storage as the one file storage is using.
- Removes `Default` implementations that are now obsolete (due to new method described above) and could cause confusion.
- Adds a new suite of integration tests for RocksDB, separate from the InMemory suite.
- Some other general improvements in the RocksDB implementation of File Storage, such as adding `error!` logs


client/file-manager/src/rocksdb.rs
```diff
@@ -42,35 +42,14 @@ fn open_or_creating_rocksdb(db_path: String) -> io::Result<Database> {
 }
 
 /// Storage backend for RocksDB.
-pub struct StorageDb<Hasher> {
+pub struct StorageDb<T> {
     pub db: Arc<dyn KeyValueDB>,
```
links234, line 46: Can this `dyn` be removed somehow or is it overcomplicating too much?

pallets/proofs-dealer/src/utils.rs
```diff
@@ -134,8 +134,7 @@ where
         ensure!(root == Self::default_forest_root(), Error::<T>::ZeroRoot);
 
         // Get last tick for which the submitter submitted a proof.
-        let last_tick_proven = match LastTickProviderSubmittedProofFor::<T>::get(submitter.clone())
-        {
+        let last_tick_proven = match LastTickProviderSubmittedProofFor::<T>::get(*submitter) {
```
links234, line 137: Not complaining, but why these changes in this PR?

links234 (approved):
Nice!

pallets/proofs-dealer/src/utils.rs
a-moreira: Just to make clippy happy.
When I run `cargo clippy` locally it showed this. Maybe I can remove these and put it in another PR just for small fixes?

links234: Personally I'm fine squeezing it in this PR!

client/file-manager/src/rocksdb.rs
a-moreira: AFAIU we still need it to use the `In Memory implementation` of the `KeyValueDB` trait in the tests. Although it's known at compile time, I wasn't able to make it work with static dispatch instead of dynamic dispatch 🤔

a-moreira: Actually, an update, I was able to remove the dynamic dispatch and use static dispatch instead. Will leave this improvement to another PR, though.

links234 (approved):
LGTM!

Only thing I'm not sure about is the integration tests, seems a bit overwhelming to have duplicate tests for each implementation. I think we should discuss more about this, I'm also concerned about the current state of integration tests which will end up running O(N^2) to test O(N) sequential steps.

a-moreira:
@links234 I understand your concern, let's discuss this with the rest of the team tomorrow after the daily.

ffarall (approved):

PR #112, opened 2024-07-02
feat: Delete file

snowmead:


primitives/traits/src/lib.rs
```diff
@@ -38,9 +38,16 @@ pub trait ProvidersInterface {
...
+        + SimpleBitOps
         + Ord
+        + Default
+        + Copy
+        + CheckEqual
+        + AsRef<[u8]>
+        + AsMut<[u8]>
         + MaxEncodedLen
-        + Copy;
+        + FullCodec;
```
ffarall, line 50: What is all these needed for?

snowmead: This is because I now ensure that `BucketId` and `ProviderId` are the same type (hashes). This requirement was added because now the file system pallet verifies the forest proof. And the forest verifier only accepts `ProviderId`s as an identifier. Working with bucket's and providers should all share the same type under these conventions.

snowmead: To answer your question specifically, the trait bounds were not enough for rust to figure out that a bucket id (being a provider id) has the `as_ref()` method. So I copied exactly the same traits from `BucketId` to `ProviderId`.

pallets/file-system/src/utils.rs
```diff
@@ -837,6 +814,135 @@ where
...
+            None => {
+                let pending_file_deletion_requests = <PendingFileDeletionRequests<T>>::get(&sender);
+
+                // Check if the file key is already in the pending deletion requests.
+                ensure!(
+                    !pending_file_deletion_requests
+                        .iter()
+                        .any(|(key, b_id)| key == &file_key && *b_id == bucket_id),
+                    Error::<T>::FileKeyAlreadyPendingDeletion
+                );
```
ffarall, line 857: File keys are unique for every single file, they include the bucket ID in their construction. So why not using `pending_file_deletion_requests.contains(file_key`?

pallets/file-system/src/utils.rs
```diff
@@ -837,6 +814,135 @@ where
...
+            fingerprint,
+        );
+
+        // Check that the metadata corresponds to the expected file key.
+        ensure!(
+            file_key == computed_file_key,
+            Error::<T>::InvalidFileKeyMetadata
+        );
+
+        let msp_id = <T::Providers as ReadProvidersInterface>::get_msp_of_bucket(&bucket_id)
```
ffarall, line 841: I'd add a check for whether this bucket belongs to that MSP, and if the bucket owner is this user.

ffarall (changes requested):
Great and tidy code, easy to read and follow. Much appreciated!

Just some small pointers.

pallets/file-system/src/utils.rs
snowmead: In this function, we are fetching the bucket data from storage directly. We are only getting the MSP from that data to report it in the event and record it in `PendingFileDeletionRequests` storage for MSPs to track and respond to.

snowmead: I added the check for the latter.

pallets/file-system/src/utils.rs
snowmead: I changed it to `.contains(&(file_key.clone(), bucket_id.clone())),`. 

I don't want to remove the bucket id from storage because then users and providers would have to do perform extra queries to figure out which bucket the file key belongs to since you can't know the pre image without fetching it from some indexer.

pallets/file-system/src/tests.rs
```diff
@@ -2279,6 +2286,359 @@ fn threshold_does_not_exceed_asymptote_success() {
...
+            assert_eq!(
+                FileSystem::pending_file_deletion_requests(owner_account_id),
+                BoundedVec::<_, <Test as crate::Config>::MaxUserPendingDeletionRequests>::try_from(
+                    vec![(file_key, bucket_id)]
+                )
+                .unwrap()
+            );
+        });
+    }
+}
```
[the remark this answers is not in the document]
snowmead: All good, but for 2. the user does not need to specify the MSP since this information is retrieved from bucket storage.

ffarall (approved):

pallets/file-system/src/lib.rs
```diff
@@ -108,7 +109,8 @@ pub mod pallet {
             + MaybeSerializeDeserialize
             + CheckedAdd
             + One
-            + Zero;
+            + Zero
+            + Into<u32>;
```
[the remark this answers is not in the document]
ffarall: Using `AtLeast32Bit` AND `Into<u32>` meant that we were forcing this type to be `u32`, when in fact we just need it to be something unsigned, relatively small. So I solved by just requiring `Into<u32>`, meaning unsigned with 32 bits or smaller.

PR #114, opened 2024-07-02
Chore: add tests for In Memory implementation of File Storage

a-moreira:


links234 (approved):
Looks good! I only left 1 comment which turns out to be more general one.

It's up for debate but from the point of view of maintainability I think it's better to have a bigger test which covers more smaller use cases, especially if there is a lot of shared prerequisites to do it. This way you have a minimal number of changes required later.

ffarall (approved):

PR #116, opened 2024-07-04
feat: ✨ Monitor challenge events in Blockchain service

ffarall:


ffarall:
Test comment

TDemeco (approved):
The runtime side looks good to me 👍🏻

links234 (approved):
Besides the MSP discussion which we can defer till we get there, everything is good!

pallets/proofs-dealer/src/lib.rs
```diff
@@ -306,6 +305,13 @@ pub mod pallet {
 
         /// A slashable provider was found.
         SlashableProvider { provider: ProviderIdFor<T> },
+
+        /// A Provider's challenge cycle was initialised.
+        NewChallengeCycleInitialised {
+            current_tick: BlockNumberFor<T>,
+            provider: ProviderIdFor<T>,
+            maybe_provider_account: Option<T::AccountId>,
```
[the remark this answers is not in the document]
ffarall: It's because the proofs dealer pallet doesn't know that it is actually impossible for a provider to exist without an account. As far as it is concerned, it queries the account for a provider, and that CAN return None.

client/blockchain-service/src/handler.rs
```diff
@@ -741,4 +796,61 @@ impl BlockchainService {
             return Err(anyhow::anyhow!("Failed to get Events storage element"));
         }
     }
+
+    /// Check if the challenges tick is one that this provider has to submit a proof for,
+    /// and if so, emit a `NewChallengeSeed` event.
+    fn should_provider_submit_proof(
```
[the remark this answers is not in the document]
ffarall: Agreed, we had already separated it a bit, but we should revisit

a-moreira (approved):
Great job! 🎉

client/blockchain-service/src/handler.rs
```diff
@@ -435,6 +445,51 @@ impl BlockchainService {
...
+                            // For each Provider ID this node monitors...
+                            for provider_id in &self.provider_ids {
+                                // ...check if the challenges tick is one that this provider has to submit a proof for.
+                                if self.should_provider_submit_proof(
+                                    &block_hash,
+                                    provider_id,
+                                    &challenges_ticker,
+                                ) {
+                                    self.emit(NewChallengeSeed {
+                                        provider_id: provider_id.clone(),
```
a-moreira, line 484: We can avoid a clone here

client/blockchain-service/src/handler.rs
```diff
@@ -741,4 +796,61 @@ impl BlockchainService {
...
+    fn should_provider_submit_proof(
+        &self,
+        block_hash: &H256,
+        provider_id: &ProviderId,
+        current_tick: &BlockNumber,
+    ) -> bool {
+        let last_tick_provided = match self
+            .client
+            .runtime_api()
+            .get_last_tick_provider_submitted_proof(block_hash.clone(), provider_id)
```
a-moreira, line 811: apparently this clone is not necessary either

client/blockchain-service/src/handler.rs
```diff
@@ -741,4 +796,61 @@ impl BlockchainService {
...
+            },
+            Err(e) => {
+                debug!(target: LOG_TARGET, "Runtime API error while getting challenge period for Provider [{:?}]: {:?}", provider_id, e);
+                return false;
+            }
+        };
+
+        trace!(target: LOG_TARGET, "CURRENT TICK: {:?}", current_tick);
+        trace!(target: LOG_TARGET, "LAST TICK PROVED: {:?}", last_tick_provided);
+        trace!(target: LOG_TARGET, "PROVIDER CHALLENGE PERIOD: {:?}", provider_challenge_period);
```
a-moreira, line 853: Maybe it would be better to print these in the same `trace!` call to avoid other logs to show up in the middle in some cases?

PR #119, opened 2024-07-08
Feat: BSP confirm storage

links234:
Complete the BSP confirm file flow after complete upload and fix bugs along the way:
 - Send `bsp_confirm_storing` transaction after a complete file upload
 - Add `query_bsp_confirm_chunks_to_prove_for_file` runtime API call which shares protocol logic with the BSP node. Also add associated commands/helper functions on the BSP's BlockchainService side.
 - Add the blockchain service event `BspConfirmedStoring` (emitted only if it's referring to running BSP)
 - Add `BspConfirmedStoringHandler` to update the BSP Forest storage
 - Multiple fixes for empty tries proofs (client & runtime)


Cargo.toml
```diff
@@ -107,6 +107,7 @@ frame-executive = { git = "https://github.com/paritytech/polkadot-sdk.git", tag
 frame-system-benchmarking = { git = "https://github.com/paritytech/polkadot-sdk.git", tag = "polkadot-v1.9.0", default-features = false }
 frame-system-rpc-runtime-api = { git = "https://github.com/paritytech/polkadot-sdk.git", tag = "polkadot-v1.9.0", default-features = false }
 frame-try-runtime = { git = "https://github.com/paritytech/polkadot-sdk.git", tag = "polkadot-v1.9.0", default-features = false }
+frame-metadata = { git = "https://github.com/paritytech/frame-metadata.git", default-features = false }
```
links234, line 110: *(question from author)*

Should I keep this in case we want to give another shot at proper client-side errors?

primitives/forest-verifier/src/lib.rs
```diff
@@ -34,6 +34,11 @@ where
         challenges: &[Self::Challenge],
         proof: &Self::Proof,
     ) -> Result<BTreeSet<Self::Challenge>, DispatchError> {
+        // Check that `challenges` is not empty.
+        if challenges.is_empty() {
+            return Err("No challenges provided.".into());
+        }
+
```
ffarall, line 41: I see why you made this. The reason for not checking, it, IIRC, was that technically, if there are no challenges, any proof is valid.

primitives/forest-verifier/src/lib.rs
```diff
@@ -49,20 +54,7 @@ where
...
-            #[allow(unreachable_code)]
-            {
-                return Err("No leaves provided in proof.".into());
-            }
-        }
-
-        // Check that `challenges` is not empty.
-        if challenges.is_empty() {
-            return Err("No challenges provided.".into());
+            return Ok(BTreeSet::new());
```
ffarall, line 57: Why are you returning this? It doesn't make sense that this would be a valid way to pass a verification

primitives/forest-verifier/src/lib.rs
```diff
@@ -178,17 +170,9 @@ where
...
-                    #[cfg(test)]
-                    unreachable!(
-                        "This should not happen. We check if the iterator has at least one leaf."
-                    );
-
-                    #[allow(unreachable_code)]
-                    {
-                        return Err("No leaves provided in proof.".into());
-                    }
+                    continue;
```
ffarall, line 175: Again, why would you consider this valid?

pallets/file-system/Cargo.toml
```diff
@@ -31,6 +32,7 @@ sp-trie = { workspace = true }
 
 # Local
 pallet-file-system-runtime-api = { workspace = true }
+pallet-proofs-dealer = { workspace = true }
```
ffarall, line 35: This shouldn't be here. This tightly couples the File System pallet to the Proofs Dealer pallet, instead of being loosely coupled (by an interface trait).

pallets/file-system/src/types.rs
```diff
@@ -61,6 +62,24 @@ pub struct StorageRequestMetadata<T: Config> {
     pub bsps_volunteered: T::StorageRequestBspsRequiredType,
 }
 
+impl<T: Config> StorageRequestMetadata<T> {
+    pub fn to_file_metadata(
+        self,
+    ) -> FileMetadata<
+        { shp_file_key_verifier::consts::H_LENGTH },
+        { shp_file_key_verifier::consts::FILE_CHUNK_SIZE },
+        { shp_file_key_verifier::consts::FILE_SIZE_TO_CHALLENGES },
```
ffarall, line 71: FYI, I'm in the process of removing this type from here, but for this PR is perfectly fine.

pallets/bucket-nfts/Cargo.toml
```diff
@@ -16,6 +16,7 @@ targets = ["x86_64-unknown-linux-gnu"]
 
 [dependencies]
 codec = { workspace = true }
+num-bigint = { workspace = true }
```
ffarall, line 19: If you only need this for the `mock.rs` file, please add them as `dev-dependencies`.

pallets/bucket-nfts/Cargo.toml
```diff
@@ -30,6 +31,7 @@ sp-std = { workspace = true }
 
 # Local
 shp-traits = { workspace = true }
+shp-file-key-verifier = { workspace = true }
```
ffarall, line 34: If you only need this for the `mock.rs` file, please add them as `dev-dependencies`.

node/src/tasks/bsp_confirmed_storing.rs
```diff
@@ -0,0 +1,94 @@
...
+        // Release the file storage lock.
+        drop(read_file_storage);
+
+        // Save [`FileMetadata`] of the newly confirmed stored file in the forest storage.
+        let mut write_forest_storage = self.storage_hub_handler.forest_storage.write().await;
+        write_forest_storage
+            .insert_metadata(&file_metadata)
+            .expect("Failed to insert metadata.");
+        // Release the forest storage lock.
+        drop(write_forest_storage);
```
ffarall, line 90: Why are you inserting the metadata here? I thought it would be already inserted after sending the volunteer transaction.

node/src/tasks/bsp_confirmed_storing.rs
```diff
@@ -0,0 +1,94 @@
+use sc_tracing::tracing::*;
```
ffarall, line 1: I'm gonna need some convincing for having this handler as a separate task. The way I see it, a `task` file would include the whole flow of what it's name describes. The `bsp_upload_file` should include the whole process of a BSP receiving a file from a user. If that includes more than one event handler, so be it, but the flow of logic is easier to follow than having to jump around files. So in short, I'd put this handler in `bsp_upload_file.rs`, but I'm open to discuss it.

docker/local-dev-bsp-compose.yml
```diff
@@ -23,6 +23,7 @@ services:
         "--keystore-path=/keystore",
         "--sealing=manual",
         "--base-path=/data",
+        "--log=runtime=debug",
```
ffarall, line 26: Maybe we shouldn't keep this in main. I'd guess you used it for debugging, and was left here.

docker/local-dev-bsp-compose.yml
```diff
@@ -50,4 +51,5 @@ services:
...
```
ffarall, line 54: Same here

client/forest-manager/src/prove.rs
```diff
@@ -76,7 +75,7 @@ where
             Ok(Proven::new_neighbour_keys(Some(leaf), None)
                 .map_err(|_| ForestStorageError::FailedToConstructProvenLeaves)?)
         }
-        _ => Err(ForestStorageError::InvalidProvingScenario.into()),
+        _ => Ok(Proven::Empty),
```
ffarall, line 78: Same here, why would this be a valid proof?

client/common/src/types.rs
```diff
@@ -55,6 +55,7 @@ pub type ParachainNetworkService = NetworkService<Block, <Block as BlockT>::Hash
 
 /// Proving either the exact key or the neighbour keys of the challenged key.
 pub enum Proven<K, D: Debug> {
+    Empty,
```
ffarall, line 58: Same here, why would this be a valid proven state?

ffarall (changes requested):

node/src/tasks/bsp_upload_file.rs
```diff
@@ -228,6 +234,46 @@ where
...
+        // Release the file storage lock.
+        drop(read_file_storage);
+
+        // Save [`FileMetadata`] of the newly confirmed stored file in the forest storage.
+        let mut write_forest_storage = self.storage_hub_handler.forest_storage.write().await;
+        write_forest_storage
+            .insert_metadata(&file_metadata)
+            .expect("Failed to insert metadata.");
+        // Release the forest storage lock.
+        drop(write_forest_storage);
```
ffarall, line 271: Why are you inserting the metadata here? I thought it would be already inserted after sending the volunteer transaction.

ffarall (changes requested):

node/src/tasks/bsp_upload_file.rs
ffarall: This is writing the metadata in forest storage. The one before, is writing it in file storage.

ffarall (approved):

PR #120, opened 2024-07-09
testing: integration bsp rpc helpers

links234:
In this PR:
1. Refactor `FileStorageRpc` to `StorageHubClientRpc` which now also has access to the forest storage.
2. Add `get_forest_root` RPC to be used in integration tests.
3. Fix minor bugs in InMemory implementation of ForestStorage.
4. Update the integration test to check the forest root (hardcoded value). To be refined in future PRs.


node/src/services/builder.rs
```diff
@@ -131,8 +132,15 @@ where
     }
 
     /// Get the [`FileStorage`] from the builder.
```
a-moreira, line 134: Doc comment is now outdated.

node/src/services/builder.rs
```diff
@@ -131,8 +132,15 @@ where
...
-        &self.file_storage
+    pub fn rpc_config(&self) -> StorageHubClientRpcConfig<T, FL, FS> {
+        StorageHubClientRpcConfig::new(
+            self.file_storage
+                .clone()
+                .expect("File Storage not initialized"),
+            self.forest_storage
+                .clone()
+                .expect("Forest Storage not initialized"),
+        )
```
a-moreira, line 143: Thinking about this, maybe it'd be better to keep separate getter methods for `file_storage` and `forest_storage` (they are different components of StorageHub and its Builder. And then we construct the `StorageHubClientRpcConfig` with them.

test/util/bspNet/api.ts
```diff
@@ -3,7 +3,7 @@ import type { BspNetApi } from "./types";
 import type { SubmittableExtrinsic } from "@polkadot/api/types";
 import type { ISubmittableResult } from "@polkadot/types/types";
 import type { KeyringPair } from "@polkadot/keyring/types";
-import { createBucket, sealBlock, sendLoadFileRpc } from "./helpers";
+import { createBucket, getForestRoot, sealBlock, sendLoadFileRpc } from "./helpers";
 import { assertEventPresent } from "../asserts";
 import type { EventRecord, H256 } from "@polkadot/types/interfaces";
```
timbrinded, line 9: Given we are adding more RPC methods and helpers, might be worth starting to add these directly to the API object. 

https://polkadot.js.org/docs/api/start/rpc.custom

We can add the custom RPC interfaces here and we can get the proper types throughout the code. @TDemeco has experience here as this is what they did extensively at a previous firm

timbrinded:
Given we are adding more RPC methods and helpers, might be worth starting to add these directly to the typegen custom object.

https://polkadot.js.org/docs/api/start/rpc.custom

We can add the custom RPC interfaces here in the typegen folder similar to how moonbeam does it:
https://github.com/moonbeam-foundation/moonbeam/blob/master/typescript-api/src/moonbase/interfaces/moon/definitions.ts

We essentially add this manually specified blob to our own bits here:
https://github.com/Moonsong-Labs/storage-hub/blob/main/api-augment/src/interfaces/custom/definitions.ts

And then we should be able to get the proper rust/ts bindings in our code

node/src/services/builder.rs
ffarall: I tend to agree with you. But this way is fine by me either way.

ffarall (approved):

client/rpc/src/lib.rs
```diff
@@ -27,14 +28,45 @@ use std::path::PathBuf;
...
+            _marker: Default::default(),
+        }
+    }
+}
 
 /// Provides an interface with the desired RPC method.
 /// Used by the `rpc` macro from `jsonrpsee`
 /// to generate the trait that is actually going to be implemented.
-#[rpc(server, namespace = "filestorage")]
+#[rpc(server, namespace = "storagehubclient")]
```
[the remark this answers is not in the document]
links234: I tended to agree with you, but after talking with Facu, we will have to split later the RPCs into storagehubclient and storagehubbsp anyway.

PR #121, opened 2024-07-09
Feat: add save file to disk RPC for BSPs

links234:


client/rpc/src/lib.rs
```diff
@@ -134,7 +162,9 @@ where
         // https://doc.rust-lang.org/std/io/trait.Read.html#errors-1
         loop {
             let mut chunk = Vec::with_capacity(FILE_CHUNK_SIZE as usize);
-            let read_result = file.by_ref().take(FILE_CHUNK_SIZE).read_to_end(&mut chunk);
```
a-moreira, line 137: Why do we need this change? Just trying to understand

client/rpc/src/lib.rs
```diff
@@ -183,6 +213,60 @@ where
...
+        // Acquire FileStorage read lock.
+        let read_file_storage = self.file_storage.read().await;
+
+        let file_key_hash: HasherOutT<T> = TryFrom::<[u8; 32]>::try_from(file_key.to_fixed_bytes())
+            .map_err(|_| into_rpc_error("Invalid file key hash"))?;
+
+        // Retrieve file metadata from File Storage.
+        let file_metadata = match read_file_storage.get_metadata(&file_key_hash) {
+            Ok(metadata) => metadata,
+            Err(FileStorageError::FileDoesNotExist) => return Ok(SaveFileToDisk::FileNotFound),
```
a-moreira, line 230: Shouldn't we return `Err` here as well?

links234: Why? The result is an enum which defines 3 states: we don't have the file, we have it partially (and we return a count of total and current chunk count) and we have it all (in which case we save it). This case is when we do not have the file.

client/rpc/src/lib.rs
links234: `by_ref` in this case is a method part of the `std::io::Read` trait.
Prior to this we only had in scope this trait, but now we also have `std::io::Write` which also has a method `by_ref`. This rightfully confuses the compiler.

client/rpc/src/lib.rs
links234: It is not really an error to not have a requested file and some consumer of this RPC might want for example to monitor the progress of the file being download.

client/rpc/src/lib.rs
a-moreira: Makes sense!

client/rpc/src/lib.rs
a-moreira: > some consumer of this RPC might want for example to monitor the progress of the file being download.

makes sense!

test/suites/integration/bsp-volunteer.test.ts
```diff
@@ -211,7 +211,11 @@ for (const bspNetConfig of bspNetConfigCases) {
       // TODO: check the file key. We need an RPC endpoint to compute the file key.
 
       await it("downloaded file passed integrity checks", async () => {
-        await checkBspForFile("test/whatsup.jpg");
```
ffarall, line 214: Why not keeping this check as well?

ffarall (approved):

test/suites/integration/bsp-volunteer.test.ts
links234: That check is just to wait for the file to be written and nothing more. After we get the response from the `saveFileToDisk` rpc, we know for sure the file is there (if success).

PR #124, opened 2024-07-11
feat: unify integration tests

a-moreira:
feat: unify integration tests under one file with different test cases for NoisyNet and RocksDB storage layer.


ffarall (approved):
Awesome work and elegant solution!

PR #130, opened 2024-07-17
feat: ✨ Implement BSP submit proof task to respond to challenges

ffarall:
In this PR:
1. Add `BspSubmitProofTask` which
    1. Is triggered by a challenge seed relevant to a `ProviderId` the Blockchain Service is paying attention to.
    2. Derives the challenges from the seed.
2. Adds the following runtime APIs to Proofs Dealer pallet:
    1. `get_challenges_from_seed`
    2. `get_forest_challenges_from_seed`
    3. `get_current_tick`
    4. `get_next_deadline_tick`
5. Adds commands to the Blockchain Service to access newly added runtime APIs from BSP tasks.
6. Creates type alias `StorageProofsMerkleTrieLayout` for the `TrieLayout` and `Hasher` used in StorageHub.
7. Adds `ChunkId::from_challenge` helper function to unify how a chunk ID is derived from a challenge hash.
8. Renames mock tasks to be prefixed with `mock_`.
9. Removes the use of `.expect` from `bsp_upload_file` task, to return with errors instead.


test/suites/integration/bsp/challenge-cycle.test.ts
```diff
@@ -1,78 +1,172 @@
...
+      const challengePeriodResult =
+        await bspApi.call.proofsDealerApi.getChallengePeriod(DUMMY_BSP_ID);
+      assert(challengePeriodResult.isOk);
+      const challengePeriod = challengePeriodResult.asOk.toNumber();
+      // Then we calculate the next challenge tick.
+      const nextChallengeTick = lastTickBspSubmittedProof + challengePeriod;
+
+      // Calculate how many blocks to advance until next challenge tick.
+      const currentBlock = await userApi.rpc.chain.getBlock();
+      const currentBlockNumber = currentBlock.block.header.number.toNumber();
```
snowmead, line 65: Here's a one liner to get the current block number:
```typescript
const blockNumber = await api.query.system.number();
```

test/util/bspNet/docker.ts
```diff
@@ -134,3 +135,13 @@ export const addBspContainer = async (options?: {
...
 };
+
+export const pauseBspContainer = async (containerName: string) => {
+  const docker = new Docker();
+  const container = docker.getContainer(containerName);
+  await container.pause();
+};
+
+export const resumeBspContainer = async (_containerName: string) => {
+  throw new Error("This function is not implemented yet.");
```
snowmead, line 146: I believe I am missing something here, isn't this just a call to `unpause`?

snowmead (approved):
Beast!

test/suites/integration/bsp/challenge-cycle.test.ts
```diff
@@ -1,78 +1,172 @@
...
+        // Increment challenge periods until we get a number that is greater than the current tick.
+        const currentTick = (await bspApi.call.proofsDealerApi.getCurrentTick()).toNumber();
+        while (currentTick >= nextChallengeTick) {
+          // Go one challenge period forward.
+          nextChallengeTick += challengePeriod;
+        }
+        // Finally, advance to the next challenge tick.
+        const currentBlock = await userApi.rpc.chain.getBlock();
+        const currentBlockNumber = currentBlock.block.header.number.toNumber();
+        const blocksToAdvance = nextChallengeTick - currentBlockNumber;
```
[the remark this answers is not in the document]
ffarall: Your assumption is correct, but I don't think we should add such parameter because `get_next_deadline_tick` should return the **deadline**, which includes the tolerance. We could have another runtime API for getting the next challenge block, but we already have runtime APIs for the last tick a provider submitted a proof for, and the challenge period of a provider. You can get the next challenge block by adding those two. And as you can see from this PR, adding a runtime API takes A LOT of boilerplate code.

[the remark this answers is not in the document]
snowmead: Ya I assumed that was the reason, but I like the way we offload StorageHub protocols to runtime APIs, since it simplifies client side development. If you don’t think it’s that much of a gain, it’s fine by me, but I would have done it.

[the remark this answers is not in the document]
ffarall: Yeah, feel free to add it whenever if you want.

test/util/bspNet/docker.ts
timbrinded: my first thought would be that the node doesn't know it needs to catch up to head and so gets sync issues. 

My first attempt to impl this will be to mount the kvp db in a temp volume, and restart the container and purge unfinalized blocks. This is similar what devops would do (turn it off and on again).

timbrinded (commented):
no issues with this, looks good. Bar has been set quite high for getting the most out of typescript

api-augment/dist/interfaces/storagehubclient/definitions.js
```diff
@@ -1,3 +1,4 @@
+import { runtime } from "./runtime";
```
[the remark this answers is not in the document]
timbrinded: as in this whole feature, not just this one line

timbrinded (approved):

client/blockchain-service/src/handler.rs
```diff
@@ -72,7 +74,7 @@ lazy_static! {
     // Would be cool to be able to do this...
     // let events_storage_key = frame_system::Events::<storage_hub_runtime::Runtime>::hashed_key();
 
-    // Static and lazily initialized `events_storage_key`
+    // Static and lazily initialised `events_storage_key`
```
TDemeco, line 77: lol 🤣

api-augment/src/interfaces/storagehubclient/runtime.ts
TDemeco, line 1: I actually did this when testing adding runtime APIs for type completion and was going to suggest we map a task for it, it's great that you did it!
I also explored a way of getting the metadata V15 from the chain and using that (which has the runtime APIs embedded in it, while the V14 doesn't) to augment the API and not have to manually add these, but after an afternoon of trying to accomplish that it proved a little more complex than expected. Maybe it's worth to explore it in the future if we keep adding or changing runtime APIs

docker/dev-keystores/bsp-down/6173676ee0e6295fe37348de5608dd982fb3f9d769c465f5a403fcceef0557d9b53c2351
TDemeco, line 1: What are all these dev-keystores files that were added? Are they to have in a file the derivation path of the BSP accounts to simulate a sort of pem file that a BSP would actually have?

runtime/src/configs/mod.rs
```diff
@@ -81,6 +81,8 @@ use super::{
 };
 use xcm_config::{RelayLocation, XcmOriginToTransactDispatchOrigin};
 
+pub type StorageProofsMerkleTrieLayout = LayoutV1<BlakeTwo256>;
```
TDemeco, line 84: Nice abstraction

TDemeco (approved):
LGTM 💯

api-augment/src/interfaces/storagehubclient/runtime.ts
ffarall: I agree with this, let's keep it in the back of our heads.

PR #131, opened 2024-07-18
feat: ✨ hooks to keep track of proof submitters and the system's price index

TDemeco:
This PR modifies the `proofs-dealer` pallet and the `payment-streams` pallet to add the missing functionality of keeping track of a price index to allow BSPs to charge for their storage correctly.

Modifications to the `proofs-dealer` pallet:
- Add a new storage map to the `proofs-dealer` pallet called `ValidProofSubmittersLastTicks`, that holds for each tick the Provider IDs of the Providers that submitted a valid proof in that tick.
- Add to the `submit_proof` function of the `proofs-dealer` pallet the logic to add the submitter (if the proof is valid) to the aforementioned storage.
- Add two constants to the configuration of the pallet, `TargetTicksStorageOfSubmitters` which determines the target amount of ticks to proof submitters lists to keep in storage (to avoid the storage bloat that we would have if we would just let that grow indefinitely) and `MaxSubmittersPerTick` which explicits the already existing implicit limitation of the maximum amount of proof submissions that we can have in a block (to allow us to bound the storage without arbitrarily limiting the capabilities of the system).
- Add the `on_idle` hook to the pallet which is in charge of, whenever possible because there's leftover weight in a block, deleting as much older valid proof submitters lists as possible, always leaving at least `TargetTicksStorageOfSubmitters` left.
- Add a new trait called `ReadProofSubmittersInterface` and implement it for the `proofs-dealer` pallet. This trait allows other pallets to read from the storage that holds the valid proof submitters of each tick, to execute any logic that requires that information.

Modifications to the `payment-streams` pallet:
- Include PR #117 , which adds a new storage structure to keep track of the block and price index up to which Providers can its users, and removes the now redundant storage of those variables from each payment stream.
- Add a new trait called `SystemMetricsInterface` and implement it for the `providers` pallet (adding a missing `UsedBspsCapacity` storage value to the pallet in the process). This trait allows any pallet to get information about the system. For now, it only includes system utilization (total capacity of the system and total capacity used of the system), but it could be extended to include other metrics.
- Adds two new storage elements: `CurrentPricePerUnitPerBlock` which holds the current price of a provided unit per block (in the case of StorageHub, this would be the price to pay BSPs per storage unit they store, per block) and `AccumulatedPriceIndex`, which is the accumulation of the `CurrentPricePerUnitPerBlock` since genesis (basically, the amount a BSP would have been paid if it has been storing one unit of data since genesis).
- Adds the `on_poll` hook which:
   - Reads the valid proof submitters from last block and updates their last chargeable info (block and price index)
   - Updates the `CurrentPricePerUnitPerBlock` according to a yet-to-be-designed equation
   - Updates the `AccumulatedPriceIndex` by adding the just-updated `CurrentPricePerUnitPerBlock`
- Adds a struct that mocks the `ReadProofSubmittersInterface` in a way that allows for testing the `payment-streams` pallet without dealing with actual proofs.
- Refactors all tests to adapt them to all these changes.


pallets/proofs-dealer/src/lib.rs
```diff
@@ -271,6 +283,19 @@ pub mod pallet {
     #[pallet::getter(fn slashable_providers)]
     pub type SlashableProviders<T: Config> = StorageMap<_, Blake2_128Concat, ProviderIdFor<T>, ()>;
 
+    /// A mapping from block number (tick) to Providers, which is set if the Provider submitted a valid proof in that tick.
+    ///
+    /// This is used to keep track of the Providers that have submitted proofs in the last few
+    /// ticks, where availability only up to the last `TargetTicksStorageOfSubmitters` ticks is guaranteed.
+    /// This storage is then made available for other pallets to use through the `ReadProofSubmittersInterface`.
+    #[pallet::storage]
+    pub type ValidProofSubmittersLastTicks<T: Config> = CountedStorageMap<
```
ffarall, line 292: Is it really worth it using a `CountedStorageMap`? It adds complexity to benchmarking reads and writes.

pallets/proofs-dealer/src/utils.rs
```diff
@@ -620,6 +651,59 @@ where
...
+    /// Trim the storage that holds the Providers that submitted valid proofs in the last blocks until there's
+    /// `TargetBlocksOfProofsStorage` blocks left (or until the remaining weight allows it).
+    ///
+    /// This function is called in the `on_idle` hook, which means it's only called when the block has
+    /// unused weight.
+    ///
+    /// It removes the oldest block from the storage that holds the providers that submitted valid proofs
+    /// in the last blocks as many times as the remaining weight allows it, but at most until the storage
+    /// has `TargetBlocksOfProofsStorage` blocks left.
+    pub fn do_trim_valid_proof_submitters_last_blocks(
```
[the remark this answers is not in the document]
ffarall: Generally speaking, just change block for ticks nomenclature everywhere in this function.

pallets/proofs-dealer/src/utils.rs
```diff
@@ -620,6 +651,59 @@ where
...
+        usable_weight: Weight,
+    ) -> Weight {
+        // Initialize the weight used by this function.
+        let mut used_weight = Weight::zero();
+
+        // Since we are using a double map and we don't know how many Providers are going to exist in each block, we delete
+        // not blocks but chunks of Providers. This means blocks older than "TargetBlocksOfProofsStorage" ago are in an undefined
+        // state and should not be used. This is a trade-off to avoid iterating over all Providers in each block, which could be
+        // computationally expensive.
+        let weight_to_remove_block = T::DbWeight::get().reads_writes(1, 1); // TODO: Benchmark this logic to get the actual weight.
```
ffarall, line 675: We can calculate this now, no need to leave it as a TODO.

pallets/proofs-dealer/src/lib.rs
```diff
@@ -527,5 +555,15 @@ pub mod pallet {
         fn integrity_test() {
             // TODO: Check that the `CheckpointChallengePeriod` is greater or equal to the largest period of a Provider.
         }
+
+        /// This hook is used to trim down the `ValidProofSubmittersLastBlocks` StorageMap up to the `TargetBlocksOfProofsStorage`.
+        ///
+        /// It runs when the block is being finalized (but before the `on_finalize` hook) and can consume all remaining weight.
+        /// It returns the used weight, so it can be used to calculate the remaining weight for the block for any other
+        /// pallets that have `on_idle` hooks.
+        fn on_idle(n: BlockNumberFor<T>, weight: Weight) -> Weight {
```
ffarall, line 564: Add tests for this hook

pallets/proofs-dealer/src/utils.rs
```diff
@@ -620,6 +651,59 @@ where
...
+                .unwrap_or(u32::MAX)
+                .min(blocks_in_storage.saturating_sub(target_blocks));
+
+            // Remove all the blocks that we can.
+            for _ in 0..removable_blocks {
+                // Get the oldest block.
+                let mut available_blocks = ValidProofSubmittersLastTicks::<T>::iter_keys()
+                    .collect::<Vec<BlockNumberFor<T>>>();
+                available_blocks.sort();
+                let oldest_block = available_blocks.first().expect("We checked before how many blocks are available and at least one should be there.");
```
ffarall, line 694: First: don't use expect here. Second, this can actually be optimised by having a variable keeping track of the last deleted tick. It's a small price to pay for avoiding collecting and ordering a vec.

pallets/providers/src/types.rs
```diff
@@ -75,6 +75,7 @@ pub type BalanceOf<T> =
     <<T as Config>::NativeBalance as Inspect<<T as frame_system::Config>::AccountId>>::Balance;
 
 /// BackupStorageProviderId is the type that represents an ID of a Backup Storage Provider, uniquely linked with an AccountId
+/// TODO: We should make these IDs configurable from the runtime...
```
ffarall, line 78: Could be... to what benefit?

pallets/payment-streams/src/mock.rs
```diff
@@ -195,6 +193,23 @@ impl Convert<BlockNumberFor<Test>, Balance> for BlockNumberToBalance {
...
+    type ProviderId = <Test as frame_system::Config>::Hash;
+    type BlockNumber = BlockNumberFor<Test>;
+    type MaxProofSubmitters = ConstU32<1000>;
+    fn get_proof_submitters_for_block(
+        block_number: &Self::BlockNumber,
+    ) -> Option<BoundedBTreeSet<Self::ProviderId, Self::MaxProofSubmitters>> {
+        let mut set = BoundedBTreeSet::<Self::ProviderId, Self::MaxProofSubmitters>::new();
+        // We convert the block number + 1 to the corresponding Provider ID, to simulate that the Provider submitted a proof
+        <StorageProviders as ProvidersInterface>::get_provider_id(*block_number + 1)
+            .map(|id| set.try_insert(id));
```
ffarall, line 208: How is this supposed to simulate many providers submitting proofs?

pallets/payment-streams/src/utils.rs
```diff
@@ -547,177 +550,253 @@ where
...
-                .checked_mul(&dynamic_rate_payment_stream.amount_provided.into())
-                .ok_or(Error::<T>::ChargeOverflow)?;
-
-            // Check the free balance of the user
-            let user_balance = T::NativeBalance::reducible_balance(
-                &user_account,
-                Preservation::Preserve,
-                Fortitude::Polite,
-            );
+            // Note: If the last chargeable price index is less than the last charged price index, we charge 0 to the user
```
ffarall, line 641:
```suggestion
            // Note: If the last chargeable price index is less than the last charged price index, we charge 0 to the user, because that would be an impossible state.
```

pallets/payment-streams/src/utils.rs
```diff
@@ -547,177 +550,253 @@ where
...
+        // If there are any proof submitters in the last tick
+        if let Some(proof_submitters) = proof_submitters {
+            // See if we have enough weight limit to process all Providers
+            let weight_per_provider = T::DbWeight::get().reads_writes(1, 1);
+            let weight_limit = weight.remaining();
+            let max_providers = weight_limit
+                .checked_div_per_component(&weight_per_provider)
+                .unwrap_or(0);
+            // If we do have enough weight, update all Providers
+            if proof_submitters.len().try_into().unwrap_or(u64::MAX) <= max_providers {
```
ffarall, line 749: This check is not really needed here, as this executes in `on_poll`. Surely there would be enough weight. But more importantly, this HAS to execute here, otherwise providers would have missed their chance to charge for a submitted proof.

pallets/payment-streams/src/utils.rs
```diff
@@ -547,177 +550,253 @@ where
...
+        let _total_used_capacity =
+            <T::ProvidersPallet as SystemMetricsInterface>::get_total_used_capacity();
+        weight.consume(T::DbWeight::get().reads(1));
+
+        // Get the total capacity of the network
+        let _total_capacity = <T::ProvidersPallet as SystemMetricsInterface>::get_total_capacity();
+        weight.consume(T::DbWeight::get().reads(1));
+
+        // Calculate the current price per unit per block
+        // TODO: Once the curve of price per unit per block is defined, implement it here
```
ffarall, line 779: Create a task in notion for this, and talk to Santi to set priority for it.

ffarall (changes requested):

pallets/payment-streams/src/mock.rs
TDemeco: It does not, but I couldn't find an elegant way to do that, mostly because this registers Provider IDs which have randomness in them, which even if it's mocked it changes block by block so it's difficult to hardcode values here. For example if in a test I register Alice first and Bob second they would have different Provider IDs that if I register Bob first and Alice second

pallets/proofs-dealer/src/lib.rs
TDemeco: If we add the variable that keeps track of the last deleted tick we don't need to know how many ticks are stored, we can just make sure we only delete up to the `ChallengesTicker - TargetTicksStorageOfSubmitters` tick

pallets/providers/src/types.rs
TDemeco: I thought about this mostly because this pallet is not the only pallet using IDs, although we could also argue that since the `storage-providers` pallet is the handler of IDs, other pallets should have generics and interact with this one through the existing interfaces... both seem valid to me

pallets/proofs-dealer/src/lib.rs
TDemeco: I added a few, I could come up with a few more but they would be mostly redundant. Let me know if there's an specific scenario that you want tested

ffarall (approved):

PR #135, opened 2024-07-23
feat: ✨ Hold funds for bucket creation

snowmead:
This PR:

- Holds funds from the user when they create a bucket.
- Releases the held funds when the bucket is deleted.
- Tests


a-moreira (approved):
Nice!

TDemeco (approved):
I like what you did about setting up the account and balances in types instead of hardcoding them 👍🏻 just not sure I love them being a tuple named after the account (just because it might not be entirely clear what the tuple is) but that's just a nitpick

ffarall (approved):

PR #136, opened 2024-07-24
docs: Slashing Protocol

snowmead:


pallets/providers/README.md
```diff
@@ -378,3 +377,45 @@ Error thrown when trying to get a root from a Main Storage Provider without pass
 ### `SpRegisteredButDataNotFound`
 
 Error thrown when a user has a Storage Provider ID assigned to it but its metadata data does not exist in storage (storage inconsistency error, should never happen).
+
+## Slashing Protocol
```
TDemeco, line 381: In general, I would write it not in the future ("will be slashed", "will accrue") but in the present tense ("are slashed", "accrues") considering as I understood this should be the docs of the actual slashing functionality.

pallets/providers/README.md
```diff
@@ -378,3 +377,45 @@ Error thrown when trying to get a root from a Main Storage Provider without pass
...
+
+Storage Providers who fail to submit a proof by the last challenge tick will be slashed, predetermined by the challenge period defined in the proofs-dealer pallet.
```
TDemeco, line 383: "last challenge tick" might get confusing, I would change it to "Storage Providers who fail to submit a proof by their deadline" or something a bit more generic, considering this pallet should be agnostic on how Providers are marked as slashable.

pallets/providers/README.md
```diff
@@ -378,3 +377,45 @@ Error thrown when trying to get a root from a Main Storage Provider without pass
...
+
+Slashing is an asynchronous process, therefore it is possible for a Storage Provider to have failed more than one challenge before being slashed. To avoid all possibility for a Storage Provider to not be slashed for the number of failed proof submissions, the runtime will accrue the number of failed challenges for each Storage Provider. Slashing a Storage Provider will take into account the total number of failed challenges and multiply it by a configurable slash factor.
+
+### Manual and Automatic Slashing
+
+The `slash` extrinsic can be called by any account to manually slash a Storage Provider and only requires the Storage Provider ID of a Storage Provider to be slashed, be it either an MSP or a BSP.
```
TDemeco, line 389:
```suggestion
The `slash` extrinsic can be called by any account to manually slash any Storage Provider that has been marked as slashable, and only requires the Storage Provider ID of a Storage Provider to be slashed, be it either an MSP or a BSP.
```

pallets/providers/README.md
```diff
@@ -378,3 +377,45 @@ Error thrown when trying to get a root from a Main Storage Provider without pass
...
+
+An automated slashing mechanism is implemented in an off-chain worker process to be executed by collators which efficiently slashes many Storage Providers.
+
+### Grace Period and Insolvency
+
+Since the Storage Provider's stake determines their total storage capacity, it is entirely possible that the amount of data currently stored would be above their total storage capacity. StorageHub grants a predetermined configurable grace period for Storage Providers to top-up their stake to have their total capacity equal to or greater than the amount of data they currently store. If the grace period has been reached, they are considered to be insolvent and cease to be a Storage Provider.
```
TDemeco, line 395:
```suggestion
Since the Storage Provider's stake determines their total storage capacity, it is entirely possible that the amount of data currently stored would be above their total storage capacity after slashing. StorageHub grants a predetermined configurable grace period for Storage Providers to top-up their stake to have their total capacity equal to or greater than the amount of data they currently store. If the grace period has been reached, they are considered to be insolvent and cease to be a Storage Provider.
```

pallets/providers/README.md
```diff
@@ -378,3 +377,45 @@ Error thrown when trying to get a root from a Main Storage Provider without pass
...
+
+The grace period is based on the total stake/capacity of the Storage Provider. In essence, the more stake a Storage Provider has, the longer the grace period.
+This is to avoid a high stake Storage Provider from being removed from the network prematurely.
+
+The runtime will automatically process any expired grace periods within the `on_poll` hook to ensure that the redundancy process is initiated as soon as possible. For every insolvent Storage Provider, an event will be emitted to notify the network and also mark the Storage Provider as insolvent, rendering them unable to operate as a Storage Provider. Finally all the of the Storage Provider's stake will be slashed and transfered to the treasury.
+
+### Ensuring Data Redundancy
+
+> [!IMPORTANT]
+> The runtime cannot ensure that all the data stored from an insolvent storage provider would be recovered. It is up to users and storage providers to ensure data redundancy since the runtime has no knowledge of file keys stored by whom.
```
TDemeco, line 405:
```suggestion
> The runtime cannot ensure that all the data stored from an insolvent Storage Provider would be recovered. It is up to users and other Storage Providers to ensure data redundancy since the runtime has no knowledge of file keys stored by whom.
```
Just to keep it consistent with previous capitalization

pallets/providers/README.md
```diff
@@ -378,3 +377,45 @@ Error thrown when trying to get a root from a Main Storage Provider without pass
...
+
+In the event when a BSP would become insolvent, the entire network of BSPs are responsible to regain data redundancy for the data they lost.
+
+To accomplish this, an off-chain indexer is required to discover the file keys which were stored by that insolvent Storage Provider. This is necessary since the runtime does not hold any file key in data in storage.
+
+Any account can call the `add_redundancy` extrinsic which requires a proof of inclusion of a given file key and the number of required BSPs needed to fulfill this request. The root is checked to be a current Bucket or BSP’s forest root to ensure that the file key does indeed exist as part of a Storage Provider's forest.
+
+This creates a traditional storage request with the specified amount of BSPs required. The caller of the extrinsic can optionally pass a list of data servers for the file key, which then will be marked in the storage request for the volunteers to request the data from. The caller is be able to obtain this information from the off-chain indexer.
```
TDemeco, line 413: This isn't entirely clear for me, but I might be missing something. Shouldn't we for BSPs just try to find a new BSP to store the files that the now insolvent BSP stored? In my head the process would be:
1. BSP becomes insolvent
2. Off-chain indexer sees this, and for each file it should issue a new storage request, with n-1/n BSPs already existing (so only one can correctly accept it)
3. After this, every file should have gotten the redundancy it previously already had

pallets/providers/README.md
```diff
@@ -378,3 +377,45 @@ Error thrown when trying to get a root from a Main Storage Provider without pass
...
+
+If the file was originally stored by an MSP, it is up to the user of the lost file or files within a bucket to execute the `transfer_file` or `transfer_bucket` extrinsics exposed by the file system pallet to move the data to a new MSP.
+
+#### Incentives and Storage Cleanup
+
+For every file key submitted for redundancy which was stored by an insolvent Storage Provider, the caller will be rewarded with a configurable amount of tokens which must be less than the slash factor to prevent abuse. The runtime will accrue the file size of each file key submitted for redundancy for the given insolvent Storage Provider. Once the total accrued file size reaches the total data size stored by the insolvent Storage Provider, the Storage Provider is deleted from the runtime.
+
+This process ensures that total redundancy is regained before the insolvent Storage Provider is removed from the network.
```
TDemeco, line 421: Oh ok I see now, so the idea would be to give incentives to BSPs to provide the now lost redundancy for the files of the insolvent BSP so the process of regaining that redundancy is quicker, and we also now when that process is done so we can safely delete the BSP. Is that correct?

pallets/providers/README.md
snowmead: This is how it works yes, but it is optional to precise more than 1 BSPs required to fulfill the storage request. This extrinsic is more generic over adding redundancy to existing file keys which do not need to be file keys lost from insolvent storage providers. Users can add more redundancy if they want to their existing files. But you are incentivized to do so for files that were lost.

pallets/providers/README.md
snowmead: Yes exactly, but anyone can call the `add_redundancy` extrinsic for lost files and get rewarded for it. For files that already exist, it must ofcourse the owner of the file to add more redundancy.

ffarall (approved):

PR #137, opened 2024-07-24
feat: create internal `RequestId` for FileTransferService

a-moreira:
This PR adds a new mapping to the FileTransferService so upper layer tasks can respond back to download requests.


links234 (approved):
Nice!

ffarall (approved):

PR #138, opened 2024-07-25
feat: implement `DownloadResponse` command and Download Task

a-moreira:
Fully implement `RemoteDownloadDataRequest` event processes


client/file-transfer-service/src/commands.rs
```diff
@@ -61,6 +67,10 @@ pub enum RequestError {
     /// File not found in the registry.
     #[error("File not found in registry")]
     FileNotRegistered,
+    #[error("DownloadRequestId not found in internal mapping")]
+    DownloadRequestIdNotFound,
+    #[error("Failed to return response")]
+    ResponseFailure(OutgoingResponse),
```
links234, line 73: This only happens when sending the download response, right? I would rename this to `DownloadResponseFailure`, but we will probably make this specific to the command when we either have another command with this case or when refactoring errors.

client/file-transfer-service/src/handler.rs
```diff
@@ -151,6 +152,73 @@ impl Actor for FileTransferService {
...
+                                            target: LOG_TARGET,
+                                            "Failed to send the response back. Looks like the requester task is gone."
+                                        ),
+                                    };
+                                }
+                            }
+                        }
+                        None => {
+                            error!(target: LOG_TARGET, "No pending response channel found for request id {:?}", request_id);
+                            match callback.send(Err(RequestError::DownloadRequestIdNotFound)) {
```
links234, line 211: This `match callback.send` is replicated 3 times. I think it would be more readable if you do it only once at the end.

links234 (approved):
Besides a small comment everything looks good to merge.

client/file-transfer-service/src/handler.rs
a-moreira: Thanks, I wasn't careful enough here. I improved it!

client/file-transfer-service/src/commands.rs
a-moreira: I added the `Download` prefix to the error and also made it return the `OutgoingResponse` struct in the error message. Will add a `TODO` so we can eventually refactor this and split into different enums

links234 (approved):

ffarall (approved):

PR #145, opened 2024-07-31
chore: remove unused dependencies from the project

a-moreira:
found using https://github.com/est31/cargo-udeps by running
```
cargo +nightly udeps --all-targets --release
```

Looks like there is a false positive coming from the `cumulus-pallet-session-benchmarking` crate (it's actually used), so I left it untouched.


pallets/randomness/Cargo.toml
```diff
@@ -19,34 +19,26 @@ sp-core = { workspace = true }
 sp-io = { workspace = true }
 sp-runtime = { workspace = true }
 sp-std = { workspace = true }
-sp-consensus-babe = { workspace = true }
 codec = { workspace = true }
 
-# Benchmarks
```
links234, line 25: Isn't this a false positive too? Shouldn't this be enabled when building with `runtime-benchmarks`?

links234 (changes requested):
Nice finding, I was wondering if there was something tool to do this.

Please double check that benchmark is indeed not needed there.

pallets/randomness/Cargo.toml
a-moreira: The problem with `runtime-benchmarks` feature is that is wasn't working even before this, so I don't know. I tried to fix it yesterday but honestly couldn't.

links234 (approved):
Approving but we need to add `runtime-bechmarks` and `try-runtime` to CI builds too later so we don't second guess if `udeps` did the right thing.

a-moreira:
I agree, eventually when the need for `try-runtime` and `runtime-benchmarks` come, we will make the features work.

pallets/randomness/Cargo.toml
```diff
@@ -19,34 +19,26 @@ sp-core = { workspace = true }
...
-frame-benchmarking = { workspace = true, optional = true }
```
TDemeco, line 26: I agree with Alexandru, even though we have not benchmarked anything yet I'm not sure if it makes sense to remove a dependency that we know we'll need

TDemeco (approved):
Tested all pallets and the runtime and everything seems to be working still 😬 so LGTM 👍🏻

snowmead (approved):

ffarall (approved):
Thank you for taking the time to look into this, much appreciated!

PR #149, opened 2024-08-06
Improve BCSV keys management

a-moreira:
Add methods to remove and insert BCSV keys instead of just rotating them for more fine grained control. Also adds tests that isolate the containers so changing the keystore doesn't mess with other suites.

Also modifies the workflow file to allow for a image rebuild in case of `client/` changes.


ffarall (approved):
Good job!

PR #151, opened 2024-08-06
fix: 🩹 change XOR threshold to hash threshold

TDemeco:
This PR addresses [issue #2](https://github.com/Moonsong-Labs/internal-storage-hub-design-audit/issues/2) of the audit, changing the way we calculate the threshold for BSP volunteering to use a hash of the concatenated BSP ID and file fingerprint instead of an XOR.


runtime/src/configs/mod.rs
```diff
@@ -653,6 +655,36 @@ impl Convert<BlockNumberFor<Runtime>, ThresholdType> for BlockNumberToThresholdT
...
+
+        // Get the equivalent u128 of the hash after mapping it to the unsigned 128-bit space
+        let hash_in_fixed_u128_space = hash_biguint * max_fixed_point_u128_number / max_u256_number;
+
+        // Convert it to the actual primitive type `u128`
+        let hash_as_u128 = <BigUint as num_traits::ToPrimitive>::to_u128(&hash_in_fixed_u128_space)
+            .expect("The number was mapped to a u128 number, it must fit.");
+
+        // Return it as a FixedU128
+        FixedU128::from_inner(hash_as_u128)
```
ffarall, line 684:
```suggestion
        // Get the hash as a BigUint value
        let hash_bytes = hash.as_ref();
        
        // Get the least significant 16 bytes of the hash and interpret them as a u128
        let truncated_hash_bytes: [u8; 16] =
            hash_bytes[16..].try_into().expect("Hash is 32 bytes; qed");
        let hash_as_u128 = u128::from_be_bytes(truncated_hash_bytes);

        // Return it as a FixedU128
        FixedU128::from_inner(hash_as_u128)
```

pallets/file-system/src/utils.rs
```diff
@@ -436,18 +434,17 @@ where
...
-                .map_err(|_| Error::<T>::FailedToEncodeBsp)?,
-        )?;
+        // Concatenate BSP ID with the file's fingerprint and hash the concatenation to get the volunteering hash.
+        let concatenated = sp_std::vec![
+            bsp_id.encode(),
+            storage_request_metadata.fingerprint.encode()
+        ]
+        .concat();
+        let volunteering_hash =
+            <<T as frame_system::Config>::Hashing as Hash>::hash(concatenated.as_ref());
```
ffarall, line 444: Let's abstract this in a function cause we will probably do a runtime API for it at some point, to have a unique place where this calculation is coded and so that it can be called from the BSP Client.

ffarall (approved):

snowmead (approved):
beast

PR #152, opened 2024-08-07
Feat: BSP forest root blockchain write lock

links234:
In this PR I add locking mechanism for all transactions that write the forest root. Currently this means 1 transaction, confirming a file storage. Changes:
1. The `on_file_complete` triggered by the last chunk will `queue_confirm_bsp_request` instead of directly sending the transaction.
2. The `BlockchainService` keeps tracks of all pending requests that require forest root writing.
3. At every enqueue or block import, the blockchain service checks for pending requests and emits either an `ProcessSubmitProofRequest` or `ProcessConfirmStoringRequest` (prioritizing the submit proof).
4. These events contain a oneshot channel used as a signal from the task processing the events to inform the blockchain service that it can spin further forest root altering tasks.
5. Remove no longer used `BspConfirmedStoring` event.
6. TS testing `api.sealBlock()` now accepts multiple transactions.


client/blockchain-service/src/handler.rs
```diff
@@ -104,6 +126,12 @@ pub struct BlockchainService {
     /// A list of Provider IDs that this node has to pay attention to submit proofs for.
     /// This could be a BSP or a list of buckets that an MSP has.
     provider_ids: Vec<ProviderId>,
+    /// A lock to prevent multiple tasks from writing to the runtime forest root (send transactions) at the same time.
+    forest_root_write_lock: Option<tokio::sync::oneshot::Receiver<()>>,
```
links234, line 130: *(note from author)*

This is a oneshot channel instead of a regular mutex because we want to "lock" in 1 thread (blockchain service) and unlock it at the end of the spawned task. The alternative would be to send a `MutexGuard` which in my opinion is even more confusing.

client/blockchain-service/src/handler.rs
```diff
@@ -902,4 +934,45 @@ impl BlockchainService {
...
+
+        // If we have a task writing to the runtime, we don't want to start another one.
+        if self.forest_root_write_lock.is_some() {
+            return;
+        }
+
+        let (tx, rx) = tokio::sync::oneshot::channel();
+        self.forest_root_write_lock = Some(rx);
+
+        let forest_root_write_tx = Arc::new(Mutex::new(Some(tx)));
```
links234, line 956: *(note from author)*

This is an `Arc<Mutex<Option<T>>>` (in this case `T=oneshot::Sender<()>`) instead of just `T` so that we can keep using the current actors event bus (emit) which requires `Clone` on the event. Clone is required because there is no constraint on the number of listeners that can subscribe to the event (and each is guaranteed to receive all emitted events).

client/blockchain-service/src/handler.rs
```diff
@@ -1046,4 +1092,49 @@ impl BlockchainService {
         };
         current_tick == &last_tick_provided.saturating_add(provider_challenge_period)
     }
+
+    fn check_pending_forest_root_writes(&mut self) {
+        if let Some(mut rx) = self.forest_root_write_lock.take() {
+            // TODO: consider adding a timeout here to prevent a task from blocking the runtime.
```
ffarall, line 1098: What do you mean by "runtime" here? The node running? I'm asking because "runtime" in a substrate node is a loaded word.

client/blockchain-service/src/handler.rs
```diff
@@ -1046,4 +1092,49 @@ impl BlockchainService {
         };
         current_tick == &last_tick_provided.saturating_add(provider_challenge_period)
     }
+
+    fn check_pending_forest_root_writes(&mut self) {
```
ffarall, line 1096: Maybe some docs on this function?

client/blockchain-service/src/handler.rs
```diff
@@ -106,6 +142,12 @@ pub struct BlockchainService {
     /// A list of Provider IDs that this node has to pay attention to submit proofs for.
     /// This could be a BSP or a list of buckets that an MSP has.
     provider_ids: Vec<ProviderId>,
+    /// A lock to prevent multiple tasks from writing to the runtime forest root (send transactions) at the same time.
+    forest_root_write_lock: Option<tokio::sync::oneshot::Receiver<()>>,
+    /// A list of [SubmitProofRequest] that are waiting to be processed (sent to the runtime).
+    pending_submit_proof: Vec<SubmitProofRequest>,
+    /// A list of [ConfirmStoringRequest] that are waiting to be processed (sent to the runtime).
+    pending_confirm_storing: Vec<ConfirmStoringRequest>,
 }
```
ffarall, line 151: How about using `VecDeque` for these ones?

client/blockchain-service/src/handler.rs
ffarall: That's probably worth commenting in the code.

client/blockchain-service/src/handler.rs
ffarall: Probably worth commenting in the code.

client/blockchain-service/src/handler.rs
```diff
@@ -1046,4 +1092,49 @@ impl BlockchainService {
...
+        }
+
+        // If we have a task writing to the runtime, we don't want to start another one.
+        if self.forest_root_write_lock.is_some() {
+            return;
+        }
+
+        let (tx, rx) = tokio::sync::oneshot::channel();
+        self.forest_root_write_lock = Some(rx);
+
```
ffarall, line 1113: This whole flow is a bit confusing. Probably worth adding some comments that explain:
Case 1: there is a message through the one-shot channel => task that had the lock has finished.
Case 2: there are no messages in the receiver => task that has the lock is still using it.
Case 3: the lock is None from the beginning, meaning no task has it.

Or restructure the code in a way that this is more evident.

client/blockchain-service/src/handler.rs
```diff
@@ -1046,4 +1092,49 @@ impl BlockchainService {
...
+            }
+        }
+
+        // If we have a task writing to the runtime, we don't want to start another one.
+        if self.forest_root_write_lock.is_some() {
+            return;
+        }
+
+        let (tx, rx) = tokio::sync::oneshot::channel();
+        self.forest_root_write_lock = Some(rx);
```
ffarall, line 1112: If there are no requests in the queue, what happens to this newly created lock? Does it get dropped at the end of of this function?

ffarall (changes requested):

ffarall:
Great job with a very difficult task. I left several comments, most are minor details, but there are some doubts that I'd like to clear.

test/suites/integration/bsp/forest-write-lock.test.ts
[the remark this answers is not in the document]
links234: If I add this test in the `volunteer` suite for example, by the time this test runs I will also get a submit proof transaction. While this is good to test too, it's not deterministic yet (i.e. if we add a couple more tests before this one, the proof request might not be issued), hence I decided to add a new suite so that I always know what I am testing.

node/src/tasks/bsp_upload_file.rs
```diff
@@ -223,65 +224,105 @@ where
...
 
+        // Send the confirmation transaction and wait for it to be included in the block and
+        // continue only if it is successfull.
+        self.storage_hub_handler
+            .blockchain
+            .send_extrinsic(call)
+            .await?
+            .with_timeout(Duration::from_secs(60))
+            .watch_for_success(&self.storage_hub_handler.blockchain)
+            .await?;
```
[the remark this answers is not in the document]
links234: That's correct.

node/src/tasks/bsp_upload_file.rs
```diff
@@ -223,65 +224,105 @@ where
...
+            event.file_key,
         );
-        let mut successful_metadatas = Vec::new();
 
-        // Get the metadata of the stored file.
+        let forest_root_write_tx = match event.forest_root_write_tx.lock().await.take() {
+            Some(tx) => tx,
+            None => {
+                error!(target: LOG_TARGET, "This is a bug! Forest root write tx already taken.");
+                return Ok(());
```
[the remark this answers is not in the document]
links234: It would make it more readable but wouldn't really change anything right now. I'm changeing it!

ffarall (approved):

PR #156, opened 2024-08-14
Refactor TrieLayout trait bound in StorageHubHandler and tasks

links234:
This PR removes generic `T` and `T: TrieLayout` and `HasherOutT<T>` trait bounds from `StorageHubHandler` and tasks by making it concrete. This makes working with storage and the client side easier by removing the need for awkward conversion between `HasherOutT<T>` and `H256`.

### DIFF
--- a/node/src/tasks/bsp_submit_proof.rs
+++ b/node/src/tasks/bsp_submit_proof.rs
@@ -4,13 +4,12 @@ use anyhow::anyhow;
 use sc_tracing::tracing::*;
 use shp_file_metadata::ChunkId;
 use sp_core::H256;
-use sp_trie::TrieLayout;
 
 use shc_actors_framework::event_bus::EventHandler;
 use shc_blockchain_service::{commands::BlockchainServiceInterface, events::NewChallengeSeed};
 use shc_common::types::{
-    HasherOutT, KeyProof, KeyProofs, Proven, ProviderId, RandomnessOutput, StorageProof,
-    TrieRemoveMutation,
+    KeyProof, KeyProofs, Proven, ProviderId, RandomnessOutput, StorageProof,
+    StorageProofsMerkleTrieLayout, TrieRemoveMutation,
 };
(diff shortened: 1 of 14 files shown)


primitives/file-metadata/src/lib.rs
```diff
@@ -80,6 +81,20 @@ impl<const H_LENGTH: usize> Into<Hash<H_LENGTH>> for FileKey<H_LENGTH> {
...
+        file_key.copy_from_slice(hash.as_bytes());
+        Self(file_key)
+    }
+}
+
+impl Into<H256> for FileKey<32> {
+    fn into(self) -> H256 {
+        H256::from_slice(&self.0)
+    }
+}
```
ffarall, line 96: Not a huge fan of this, as it feels like we're braking the abstraction of this crate, but I agree on it's convenience, so 👍🏼

node/src/tasks/bsp_upload_file.rs
```diff
@@ -404,7 +385,7 @@ where
         Ok(())
     }
 
-    async fn unvolunteer_file(&self, file_key: HasherOutT<T>) -> anyhow::Result<()> {
+    async fn unvolunteer_file(&self, file_key: H256) -> anyhow::Result<()> {
```
ffarall, line 388: How about using `HasherOutT<StorageProofsMerkleTrieLayout>` in this document, instead of `H256`?

node/src/tasks/bsp_submit_proof.rs
```diff
@@ -329,7 +284,7 @@ where
 
     async fn generate_key_proof(
         &self,
-        file_key: HasherOutT<T>,
+        file_key: H256,
```
ffarall, line 287: How about using `HasherOutT<StorageProofsMerkleTrieLayout>` in this document, instead of `H256`?

ffarall (approved):
Awesome work and very necessary. Appreciate it!

PR #158, opened 2024-08-14
feat: Revised threshold formulas

snowmead:
Revised threshold formulas to resolve [audit issue](https://github.com/Moonsong-Labs/internal-storage-hub-design-audit/issues/6).

Formalized documentation of the formulas - [README](https://github.com/Moonsong-Labs/storage-hub/blob/feat/threshold-refactor/pallets/file-system/README.md#volunteering-succeeding-threshold-checks)

- Adds new storage values in file system pallet to configure threshold formulas.
- Adds BSP basic reputation weight and track global weight of all BSPs.
- Adds `GetProviderId` blockchain service command and `get_bsp_provider_id` runtime api to obtain the BSPs provider id (needed to call compute threshold runtime).
- Modified `HashToThresholdTypeConverter` to take the hash's 4 least significant bytes to fit into `u32` threshold type (was a `u128`).
- Removes `SubscribeProvidersInterface`. No longer used due to new threshold formulas being computed on the fly.


pallets/file-system/src/lib.rs
```diff
@@ -131,13 +132,10 @@ pub mod pallet {
...
             + PartialOrd
-            + FixedPointNumber;
+            + Zero;
 
         /// The type to convert a threshold to a block number.
-        type ThresholdTypeToBlockNumber: Convert<Self::ThresholdType, BlockNumberFor<Self>>;
-
-        /// The type to convert a block number to a threshold.
-        type BlockNumberToThresholdType: Convert<BlockNumberFor<Self>, Self::ThresholdType>;
+        type ThresholdTypeToBlockNumber: ConvertBack<Self::ThresholdType, BlockNumberFor<Self>>;
```
TDemeco, line 138: Why do we implement the Convert trait for this one in the mock when here we only need ConvertBack? Maybe a better solution would be to have a Convert from BlockNumber to ThresholdType instead

pallets/file-system/src/lib.rs
```diff
@@ -299,6 +278,7 @@ pub mod pallet {
         ValueQuery,
     >;
 
+    /// Number of BSPs required to fulfill a storage request
```
TDemeco, line 281: This comment shouldn't be here right?

pallets/file-system/src/lib.rs
```diff
@@ -316,42 +296,52 @@ pub mod pallet {
         (BlockNumberFor<T>, StorageData<T>),
     >;
 
-    /// Minimum BSP assignment threshold.
+    // Number of BSPs required to fulfill a storage request
```
TDemeco, line 299:
```suggestion
    /// Number of BSPs required to fulfill a storage request
```

pallets/file-system/src/mock.rs
```diff
@@ -12,11 +12,12 @@ use shp_file_metadata::ChunkId;
 use shp_traits::{
     CommitmentVerifier, MaybeDebug, ProofSubmittersInterface, TrieMutation, TrieProofDeltaApplier,
 };
-use sp_core::{hashing::blake2_256, ConstU128, ConstU32, ConstU64, Get, Hasher, H256};
+use sp_core::{hashing::blake2_256, ConstU128, ConstU32, ConstU64, Get, Hasher, H256, U256};
 use sp_keyring::sr25519::Keyring;
+use sp_runtime::traits::ConvertBack;
```
TDemeco, line 17: Same here^

pallets/file-system/src/mock.rs
```diff
@@ -406,44 +397,35 @@ impl Convert<Balance, BlockNumberFor<Test>> for SaturatingBalanceToBlockNumber {
     }
 }
 
-// Converter from the ThresholdType type (FixedU128) to the BlockNumber type (u64).
+// Converter from the ThresholdType type (FixedU128) to the BlockNumber type (u64) and vice versa.
```
TDemeco, line 400: Same here^

runtime/src/configs/mod.rs
```diff
@@ -641,31 +635,23 @@ impl Convert<Balance, BlockNumberFor<Runtime>> for SaturatingBalanceToBlockNumbe
     }
 }
 
-// Converter from the ThresholdType type (FixedU128) to the BlockNumber type (u64).
+// Converter from the ThresholdType type (FixedU128) to the BlockNumber type (u64) and vice versa.
```
TDemeco, line 638: Same here^

xcm-simulator/src/storagehub/configs/mod.rs
```diff
@@ -674,31 +668,23 @@ impl Convert<Balance, BlockNumberFor<Runtime>> for SaturatingBalanceToBlockNumbe
     }
 }
 
-// Converter from the ThresholdType type (FixedU128) to the BlockNumber type (u64).
+// Converter from the ThresholdType type (FixedU128) to the BlockNumber type (u64) and vice versa.
```
TDemeco, line 671: Same here^

TDemeco (approved):
Nice work! Left some minor comments and a commit to update XCM tests for the new configs

pallets/file-system/src/lib.rs
snowmead: We use both `convert` and `convert_back` in file system pallet.

TDemeco: Oh but we only require the ConvertBack trait in the config of the pallet, how does that work?

snowmead: Oh it's because the `ConvertBack` inherits `Convert`. So you don't have to specify both.

pallets/providers/src/lib.rs
```diff
@@ -139,6 +136,19 @@ pub mod pallet {
...
+            + Member
+            + MaybeSerializeDeserialize
+            + Default
+            + MaybeDisplay
+            + AtLeast32BitUnsigned
+            + Saturating
+            + Copy
+            + MaxEncodedLen
+            + HasCompact
+            + Into<u32>;
```
ffarall, line 150: By restricting this to `AtLeast32BitUnsigned` and `Into<u32>` you're essentially forcing this to be a `u32`. Can we avoid that?

pallets/file-system/src/mock.rs
```diff
@@ -335,29 +333,25 @@ impl crate::Config for Test {
...
     type Nfts = Nfts;
     type CollectionInspector = BucketNfts;
-    type AssignmentThresholdDecayFactor = ThresholdAsymptoticDecayFactor;
-    type AssignmentThresholdAsymptote = ThresholdAsymptote;
-    type AssignmentThresholdMultiplier = ThresholdMultiplier;
-    type TargetBspsRequired = ConstU32<3>;
-    type MaxBspsPerStorageRequest = ConstU32<5>;
+    type MaxBspsPerStorageRequest = ConstU32<10>;
     type MaxBatchConfirmStorageRequests = ConstU32<10>;
+    // TODO: this should probably be a multiplier of the number of maximum multiaddresses per storage provider
```
ffarall, line 347: Why?

pallets/file-system/src/mock.rs
```diff
@@ -406,44 +396,34 @@ impl Convert<Balance, BlockNumberFor<Test>> for SaturatingBalanceToBlockNumber {
...
         // Get the hash as bytes
         let hash_bytes = hash.as_ref();
 
-        // Get the 16 least significant bytes of the hash and interpret them as a u128
-        let truncated_hash_bytes: [u8; 16] =
-            hash_bytes[16..].try_into().expect("Hash is 32 bytes; qed");
-        let hash_as_u128 = u128::from_be_bytes(truncated_hash_bytes);
+        // Get the 4 least significant bytes of the hash and interpret them as an u32
+        let truncated_hash_bytes: [u8; 4] =
+            hash_bytes[28..].try_into().expect("Hash is 32 bytes; qed");
```
ffarall, line 424: Why 28 here as well?

pallets/file-system/src/mock.rs
ffarall: If you make any changes here as a consequence of this comment, remember to make them in all the other `mock.rs` files.

pallets/file-system/src/mock.rs
ffarall: If you make any changes here as a consequence of this comment, remember to make them in all the other `mock.rs` files.

primitives/traits/src/lib.rs
```diff
@@ -159,6 +159,24 @@ pub trait ReadProvidersInterface: ProvidersConfig + ProvidersInterface {
...
     /// Maximum number of multiaddresses a provider can have.
     type MaxNumberOfMultiAddresses: Get<u32>;
+    /// Type that represents the reputation weight of a Storage Provider.
+    type ReputationWeight: Parameter
+        + Member
+        + MaybeSerializeDeserialize
+        + Default
+        + Ord
+        + AtLeast32BitUnsigned
+        + Into<u32>
```
ffarall, line 169: These two restrictions make it essentially have to be `u32`.

runtime/src/configs/mod.rs
```diff
@@ -674,13 +659,11 @@ impl Convert<<Runtime as frame_system::Config>::Hash, ThresholdType>
         // Get the hash as bytes
         let hash_bytes = hash.as_ref();
 
-        // Get the 16 least significant bytes of the hash and interpret them as a u128
-        let truncated_hash_bytes: [u8; 16] =
-            hash_bytes[16..].try_into().expect("Hash is 32 bytes; qed");
-        let hash_as_u128 = u128::from_be_bytes(truncated_hash_bytes);
+        // Get the 4 least significant bytes of the hash and interpret them as an u32
+        let truncated_hash_bytes: [u8; 4] =
+            hash_bytes[28..].try_into().expect("Hash is 32 bytes; qed");
```
ffarall, line 664: Why `28` here?

pallets/file-system/src/utils.rs
```diff
@@ -1310,10 +1301,10 @@ mod hooks {
         fn process_expired_storage_request(file_key: MerkleHash<T>, remaining_weight: &mut Weight) {
             let db_weight = T::DbWeight::get();
 
-            // As of right now, the upper bound limit to the number of BSPs required to fulfill a storage request is set by `TargetBspsRequired`.
+            // As of right now, the upper bound limit to the number of BSPs required to fulfill a storage request is set by `ReplicationTarget`.
```
ffarall, line 1304:
```suggestion
            // As of right now, the upper bound limit to the number of BSPs required to fulfil a storage request is set by `ReplicationTarget`.
```

pallets/file-system/src/utils.rs
```diff
@@ -1196,49 +1174,62 @@ where
         // Return the threshold needed for the BSP to be able to volunteer for the storage request.
         T::HashToThresholdType::convert(volunteering_hash)
     }
-}
 
-impl<T: crate::Config> shp_traits::SubscribeProvidersInterface for Pallet<T> {
-    type ProviderId = ProviderIdFor<T>;
+    pub fn compute_threshold_to_succeed(
```
ffarall, line 1178: Probably worth documenting this function. I found myself having to read the code to understand what exactly it is supposed to do.

pallets/file-system/src/utils.rs
```diff
@@ -1196,49 +1174,62 @@ where
...
+        bsp_id: &ProviderIdFor<T>,
+        requested_at: BlockNumberFor<T>,
+    ) -> Result<(T::ThresholdType, T::ThresholdType), DispatchError> {
+        let maximum_threshold: u32 = MaximumThreshold::<T>::get().into();
```
ffarall, line 1182: This essentially makes the `MaximumThreshold` type useless 😂

pallets/file-system/src/utils.rs
```diff
@@ -1196,49 +1174,62 @@ where
...
+        }
 
-        bsp_assignment_threshold = base_threshold
-            .checked_div(&T::AssignmentThresholdDecayFactor::get())
-            .ok_or(Error::<T>::ThresholdArithmeticError)?
-            .saturating_add(T::AssignmentThresholdAsymptote::get());
+        // Global threshold starting point from which all BSPs begin their threshold slope. All BSPs start at this point
+        // with the starting reputation weight.
+        let threshold_global_starting_point = maximum_threshold
+            .checked_mul(ReplicationTarget::<T>::get().into() / global_weight.into() / 2)
```
ffarall, line 1194: Maybe using some `saturated_div`s here? Also, this would saturate to zero in the divisions. `global_weight` is potentially a very large number, and `ReplicationTarget::<T>::get()` is something around 10. My suggestion would be to do:
```
maximum_threshold.saturated_div(global_weight).saturated_mul(ReplicationTarget::<T>::get()).saturated_div(2)
```
In that order. What you achieve is that:
1. In a regular scenario, `maximum_threshold` would be very large, and you'd start bringing it down with `global_weight`, also a large number. That way, when you multiply it by `ReplicationTarget::<T>::get()` you should still be within the numerical domain.
2. If `global_weight` is low still (the beginning of the network), when multiplying with `ReplicationTarget::<T>::get()` you'll get at most `u32::MAX` and then the threshold would be `u32::MAX / 2` (still pretty high).
3. If `maximum_threshold` is very low (like some times you set it in tests), the division would saturate to 1, and then the threshold would be `ReplicationTarget::<T>::get() / 2` (still very low).

pallets/file-system/src/utils.rs
```diff
@@ -1196,49 +1174,62 @@ where
...
 
-    fn subscribe_bsp_sign_up(_who: &Self::ProviderId) -> DispatchResult {
-        // Adjust bsp assignment threshold by applying the decay function after removing the asymptote
-        let mut bsp_assignment_threshold = BspsAssignmentThreshold::<T>::get();
-        let base_threshold =
-            bsp_assignment_threshold.saturating_sub(T::AssignmentThresholdAsymptote::get());
+        // Rate of increase from the weighted threshold starting point up to the maximum threshold within a block range.
+        let threshold_slope = (<u32 as Into<T::ThresholdType>>::into(maximum_threshold)
+            .saturating_sub(threshold_weighted_starting_point))
+            / T::ThresholdTypeToBlockNumber::convert_back(BlockRangeToMaximumThreshold::<T>::get());
```
ffarall, line 1211: Use `saturated_div` here.

pallets/file-system/src/utils.rs
ffarall: Given how much thought has to be put into this calculation and it's edge cases, probably worth adding some tests for it with different parameters.

pallets/file-system/Cargo.toml
```diff
@@ -16,6 +16,7 @@ targets = ["x86_64-unknown-linux-gnu"]
 
 [dependencies]
 codec = { workspace = true }
+log = { workspace = true }
```
ffarall, line 19: Is this needed?

pallets/file-system/README.md
```diff
@@ -13,3 +13,59 @@ If the proof is provided, the runtime will verify the proof and queue a priority
 If the proof is not provided, a pending file deletion request is created with a defined constant expiration time. The MSP who is supposed to be storing the file, will have until the expiration time to provide a proof of (non-)inclusion of the file in their _forest_. If it is a proof of non-inclusion, this means the file does not exist, and the pending file deletion request will be removed. This prevents all the providers from responding to a priority challenge for a file that does not exist. If it is a proof of inclusion, a priority challenge will be queued in the proofs-dealer pallet, following the same process described above.
 
 Once a pending file deletion request reaches its expiration time and it has not been responded to, a priority challenge will be queued, following the same process described above.
+
+### Volunteering: Succeeding Threshold Checks
```
ffarall, line 17: Much appreciated docs.

ffarall (changes requested):
Great work! This was a challenging change and I saw a lot of good stuff. Just some things to take into account and fix.

pallets/file-system/src/lib.rs
```diff
@@ -316,42 +294,52 @@ pub mod pallet {
         (BlockNumberFor<T>, StorageData<T>),
     >;
 
-    /// Minimum BSP assignment threshold.
+    /// Number of BSPs required to fulfill a storage request
```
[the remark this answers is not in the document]
snowmead: I should be unbiased according to Grammarly, but I have always used “fulfill”. 😄  

image

pallets/file-system/Cargo.toml
snowmead: I throw a warning in the file system pallet in the compute threshold function.

xcm-simulator/src/storagehub/configs/mod.rs
```diff
@@ -707,13 +692,11 @@ impl Convert<<Runtime as frame_system::Config>::Hash, ThresholdType>
         // Get the hash as bytes
         let hash_bytes = hash.as_ref();
 
-        // Get the 16 least significant bytes of the hash and interpret them as a u128
-        let truncated_hash_bytes: [u8; 16] =
-            hash_bytes[16..].try_into().expect("Hash is 32 bytes; qed");
-        let hash_as_u128 = u128::from_be_bytes(truncated_hash_bytes);
+        // Get the 4 least significant bytes of the hash and interpret them as an u32
+        let truncated_hash_bytes: [u8; 4] =
+            hash_bytes[28..].try_into().expect("Hash is 32 bytes; qed");
```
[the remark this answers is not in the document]
snowmead: For all runtimes including mocks, I set the threshold type to `u32`. Therefore I take now 4 bytes instead of 16 (threshold type was `u128`) from the hash.

pallets/file-system/src/utils.rs
snowmead: Thanks for point this out along with the other related comments regarding the use of `u32`.

pallets/file-system/src/lib.rs
```diff
@@ -122,6 +120,8 @@ pub mod pallet {
             + Debug
             + Default
             + MaybeDisplay
+            + Into<u32>
+            + From<u32>
```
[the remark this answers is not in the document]
snowmead: I revised the associated types bounds for all mentioned ones in your review. Only restricting them to the necessary traits needed.

pallets/providers/src/lib.rs
snowmead: Removed reliance of `Into<u32>` trait bounds for all associated types in file system pallet. Only requiring necessary `From<...>` trait bounds.

pallets/file-system/src/utils.rs
snowmead: I have amended the formulas based on your comments. I had to use `checked_div` instead of `saturated_div` since it does not exist.

xcm-simulator/src/storagehub/configs/mod.rs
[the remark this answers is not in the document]
ffarall: Right, so shouldn't it be `4`? I believe `hash_bytes` is a stream of bytes, not bits. So you don't need the first 28 bytes, just the first 28 bits (4 bytes)

pallets/file-system/Cargo.toml
ffarall: But wouldn't that not be included when compiled to WASM? The runtime API is executed by instantiating a WASM blob.

pallets/file-system/src/lib.rs
```diff
@@ -316,42 +294,52 @@ pub mod pallet {
...
+
+    /// Maximum threshold a BSP can attain.
     #[pallet::storage]
-    #[pallet::getter(fn bsps_assignment_threshold)]
-    pub type BspsAssignmentThreshold<T: Config> = StorageValue<_, T::ThresholdType, ValueQuery>;
+    pub type MaximumThreshold<T: Config> = StorageValue<_, T::ThresholdType, ValueQuery>;
+
+    /// Number of blocks until all BSPs would reach the [`Config::MaximumThreshold`] to ensure that all BSPs are able to volunteer.
+    #[pallet::storage]
+    pub type BlockRangeToMaximumThreshold<T: Config> =
```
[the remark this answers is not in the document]
ffarall: Fair enough about the getters. Although it probably doesn't hurt to have that one liner.

All good about making them storage elements.

pallets/file-system/src/lib.rs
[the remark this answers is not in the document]
ffarall: Oh, my spellchecker was marking it as wrong, but that's because "fulfill" is American English apparently.

pallets/providers/src/lib.rs
ffarall: I still see both `AtLeast32BitUnsigned` and `Into<u32>` as trait restrictions here. Am I seeing something wrong?

pallets/file-system/src/utils.rs
```diff
@@ -1208,49 +1177,80 @@ where
...
-        Ok(())
-    }
+        // Global threshold starting point from which all BSPs begin their threshold slope. All BSPs start at this point
+        // with the starting reputation weight.
+        let threshold_global_starting_point = maximum_threshold
+            .checked_div(&global_weight)
+            .unwrap_or(T::ThresholdType::one())
+            .checked_mul(&ReplicationTarget::<T>::get().into()).unwrap_or({
+            log::warn!("Global starting point is beyond MaximumThreshold. Setting it to half of the MaximumThreshold.");
+            maximum_threshold / 2.into()
```
ffarall, line 1208: I'd still use `saturated_div(2)` for the last operation. And probably document why we do operations in this order.

pallets/bucket-nfts/src/mock.rs
```diff
@@ -367,44 +348,34 @@ pub fn new_test_ext() -> sp_io::TestExternalities {
...
         // Get the hash as bytes
         let hash_bytes = hash.as_ref();
 
-        // Get the 16 least significant bytes of the hash and interpret them as a u128
-        let truncated_hash_bytes: [u8; 16] =
-            hash_bytes[16..].try_into().expect("Hash is 32 bytes; qed");
-        let hash_as_u128 = u128::from_be_bytes(truncated_hash_bytes);
+        // Get the 4 least significant bytes of the hash and interpret them as an u32
+        let truncated_hash_bytes: [u8; 4] =
+            hash_bytes[28..].try_into().expect("Hash is 32 bytes; qed");
```
ffarall, line 376: Check the 28 here too.

client/blockchain-service/src/handler.rs
```diff
@@ -348,6 +348,44 @@ impl Actor for BlockchainService {
                         }
                     }
                 }
+                BlockchainServiceCommand::GetProviderId {
```
ffarall, line 351: Are you using this for something?

pallets/file-system/src/lib.rs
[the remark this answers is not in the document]
snowmead: I will add it, but it would be unused.

xcm-simulator/src/storagehub/configs/mod.rs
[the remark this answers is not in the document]
snowmead: Yes, it is an array of bytes. We are grabbing the 4 least significant bytes (equaling 32 bits), fitting into the 32 bit unsigned integer threshold type.

pallets/providers/src/lib.rs
snowmead: Thanks for spotting this. Removed.

pallets/file-system/Cargo.toml
snowmead: It was for when we execute tests.

pallets/file-system/src/utils.rs
snowmead: Adding something like `checked_div` conveys that this can have a result of `0`, which is the only condition for which it returns `None`. Which is impossible in our case since we would never set `maximum_threshold` to be 2 or less.

I will add docs.

client/blockchain-service/src/handler.rs
snowmead: Yes

https://github.com/Moonsong-Labs/storage-hub/blob/bc1ad6c6e229a71df12ea6e48c9c396a5d4e2eee/node/src/tasks/bsp_upload_file.rs#L363-L369

pallets/bucket-nfts/src/mock.rs
snowmead: Resolving discussion in duplicate comment.

pallets/file-system/src/mock.rs
snowmead: Resolving discussion in duplicate comment.

xcm-simulator/src/storagehub/configs/mod.rs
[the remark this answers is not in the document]
ffarall: Right, so shouldn't this be 4 instead of 28 then?

pallets/file-system/src/utils.rs
ffarall: I understand and agree. The insistence on still using `checked_div` is to avoid at all costs panics in the runtime code. Even if it is impossible.

pallets/file-system/Cargo.toml
snowmead: I can understand that, was not a big deal to me since I looked around and found log prints in substrate pallets as well. I can remove it if you insist.

pallets/file-system/src/utils.rs
snowmead: Ok will do.

pallets/file-system/Cargo.toml
snowmead: I really think this is nice to have though.

https://github.com/Moonsong-Labs/storage-hub/blob/f80e30150bd845de84a95e4aced971a9e363f7b0/pallets/file-system/src/utils.rs#L1206-L1212

ffarall: It's fine, leave it then if you find it useful

ffarall (approved):

PR #161, opened 2024-08-16
feat: ✨ Monitor events in finality

ffarall:
This PR:
1. Adds a new notification stream for the Blockchain Service: finalised blocks. Now the blockchain service can also emit events of finalised transactions.
2. Adds `MutationsApplied` to Proofs Dealer pallet, to advertise when mutations are applied as a consequence of the `submit_proof` extrinsic.
3. Adds `get_provider_id` runtime API to Storage Providers pallet, to fetch a BSP or MSP ID linked with an `AccountId`.
4. Updates the `provider_ids` that the Blockchain Service monitors on every block, based on the keys in the node's keystore.
5. In the Blockchain Service, filter `BspConfirmedStoring` events using provider IDs, instead of the keys in the keystore.


client/blockchain-service/src/handler.rs
```diff
@@ -745,305 +670,55 @@ impl BlockchainService {
...
+                        // New storage request event coming from pallet-file-system.
+                        RuntimeEvent::ProofsDealer(
+                            pallet_proofs_dealer::Event::MutationsApplied {
+                                provider,
+                                mutations,
+                                new_root,
+                            },
+                        ) => {
+                            // Check if the provider ID is one of the provider IDs this node is tracking.
+                            for provider_id in self.provider_ids.iter() {
```
links234, line 700: Maybe we can change this to a `HashSet`?

client/blockchain-service/src/handler.rs
```diff
@@ -544,38 +497,14 @@ impl BlockchainService {
...
-                        error!(target: LOG_TARGET, "Failed to notify task about block number.");
-                    }
-                }
-            }
-        }
-
-        for key in keys_to_remove {
-            self.wait_for_block_request_by_number.remove(&key);
-        }
+        self.notify_block_number(block_number);
```
links234, line 500: I would rename this to `notify_import_block_number` to remove any doubt.

links234 (approved):
Looks good!

PR #162, opened 2024-08-19
test: 🧪 Add BSP testshelper

timbrinded:
## Added

- New helper: `assertExtrinsicPresent()`
```ts
async function assertExtrinsicPresent(
  api: BspNetApi,
  options: {
    blockHeight?: string;
    blockHash?: string;
    skipSuccessCheck?: boolean;
    checkTxPool?: boolean;
    module: string;
    method: string;
    ignoreParamCheck?: boolean;
  }
): Promise<
  {
    module: string;
    method: string;
    extIndex: number;
  }[]
> 
```
   - Will check the last block for extrinsic of `module.method`, throws or returns filtered list
   - Option to check against pending txPool instead of last block
   - Option to check of a particular block instead
   - Will verify whether the input module.method exists in the api blob (e.g. check for typos). Has option to disable
   - Will verify that it has associated extrinsicSuccess Event, with option to disable.

### DIFF
--- a/test/util/asserts.ts
+++ b/test/util/asserts.ts
@@ -5,28 +5,6 @@ import type { AugmentedEvent } from "@polkadot/api/types";
 import type { BspNetApi } from "./bspNet";
 
 //TODO: add ability to search nested extrinsics e.g. sudo.sudo(balance.forceTransfer(...))
-/**
- * Asserts that an extrinsic is present in a block or transaction pool.
- * 
- * @param api - The connected API instance for interacting with a running BspNet.
- * @param options - Configuration options for the extrinsic assertion.
- * @param options.blockHeight - The block height to check. If not provided, checks the latest block.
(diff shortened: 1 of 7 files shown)


test/util/asserts.ts
```diff
@@ -2,6 +2,79 @@ import type { EventRecord } from "@polkadot/types/interfaces";
 import { strictEqual } from "node:assert";
 import type { ApiPromise } from "@polkadot/api";
 import type { AugmentedEvent } from "@polkadot/api/types";
+import type { BspNetApi } from "./bspNet";
+
+//TODO: add ability to search nested extrinsics e.g. sudo.sudo(balance.forceTransfer(...))
+export const assertExtrinsicPresent = async (
```
ffarall, line 8: Might be worth adding some doc comments on what the options are and what is their behaviour.

ffarall (approved):
Good work! Approved with just one comment

PR #163, opened 2024-08-20
feat: ✨ refactor runtime interfaces

TDemeco:
This PR:
- Refactors all interfaces used in the runtime to be more clear, better modularized, and also separates the ProvidersInterface into two: one for Challegeable Providers (BSPs) to be used by the `proofs-dealer` pallet, and another one for all Providers that's used in the `file-system` and `payment-streams` pallets.
- Adds a new function to the ProofsDealerInterface called `verify_generic_forest_proof`, used when wanting to verify a proof of inclusion without requiring a Provider (since the function receives directly the root of the forest). This is useful since we now not consider Buckets a challegeable entity but the `file-system` pallet has to challenge a Bucket for inclusion proofs when removing files


client/common/src/types.rs
```diff
@@ -43,21 +43,12 @@ pub type ForestVerifierProof = pallet_proofs_dealer::types::ForestVerifierProofF
...
     sp_io::SubstrateHostFunctions,
     cumulus_client_service::storage_proof_size::HostFunctions,
 );
 
-#[cfg(feature = "runtime-benchmarks")]
-type HostFunctions = (
-    // TODO: change this to `cumulus_client_service::ParachainHostFunctions` once it is part of the next release
-    sp_io::SubstrateHostFunctions,
-    cumulus_client_service::storage_proof_size::HostFunctions,
-    frame_benchmarking::benchmarking::HostFunctions,
```
ffarall, line 58: I think this probabaly has to stay here 🤔

primitives/traits/src/lib.rs
```diff
@@ -233,38 +448,110 @@ pub trait MutateProvidersInterface: ProvidersConfig + ProvidersInterface {
...
+    /// Get the default value for the root of a Merkle Patricia Forest.
+    fn get_default_root() -> Self::MerkleHash;
+
+    /// Get the stake for a registered Provider.
+    fn get_stake(
+        who: Self::ProviderId,
+    ) -> Option<<Self::Balance as fungible::Inspect<Self::AccountId>>::Balance>;
+}
+
+/// A trait to mutate the state of a genneric Provider, such as updating their root.
```
ffarall, line 496:
```suggestion
/// A trait to mutate the state of a generic Provider, such as updating their root.
```

ffarall (approved):

PR #166, opened 2024-08-22
Feat: BSP batch confirm file storing

links234:
This PR builds on top of the forest write lock to batch multiple confirm file requests.


node/src/tasks/bsp_upload_file.rs
```diff
@@ -250,50 +250,79 @@ where
             }
         };
 
-        // Query runtime for the chunks to prove for the file.
-        let chunks_to_prove = self
+        let own_bsp_id = self
```
ffarall, line 253: Shouldn't this be called `own_bsp_account`? Or `own_bsp_account_id`. Because `own_bsp_id` makes it sound as if it were the BSP ID.

node/src/tasks/bsp_upload_file.rs
```diff
@@ -250,50 +250,79 @@ where
...
+                .await
+            {
+                Ok(chunks_to_prove) => {
+                    file_key_with_chunks_to_prove.push((*file_key, chunks_to_prove));
+                }
+                Err(e) => {
+                    error!(target: LOG_TARGET, "Failed to query chunks to prove for file {:?}: {:?}\nEnqueuing file key again!", file_key, e);
+                    self.storage_hub_handler
+                        .blockchain
+                        .queue_confirm_bsp_request(ConfirmStoringRequest::new(*file_key))
```
ffarall, line 275: This is a fair strategy, but could lead to something always being there in the queue infinitely, if this is a consistent error. I think for this particular case, if this fails and we fail to confirm a file, it is not that bad to just give up that file. Let me know your thoughts.

node/src/tasks/bsp_upload_file.rs
```diff
@@ -250,50 +250,79 @@ where
...
+                (Ok(proof), Ok(metadata)) => {
+                    file_keys_and_proofs.push((file_key, proof));
+                    file_metadatas.push(metadata);
+                }
+                _ => {
+                    error!(target: LOG_TARGET, "Failed to generate proof or get metadatas for file {:?}.\nEnqueuing file key again!", file_key);
+                    self.storage_hub_handler
+                        .blockchain
+                        .queue_confirm_bsp_request(ConfirmStoringRequest::new(file_key))
+                        .await?;
```
ffarall, line 299: Same here.

ffarall (commented):
Great job! Just some comments and things to discuss.

ffarall (approved):

client/blockchain-service/src/handler.rs
```diff
@@ -525,6 +535,22 @@ impl Actor for BlockchainService {
                         }
                     }
                 }
+                BlockchainServiceCommand::QueryStorageProviderId { account, callback } => {
```
ffarall, line 538: @snowmead also added a similar functionality to this one. So depending on who merges first, try to make it so that they don't collide.

ffarall (approved):

PR #167, opened 2024-08-23
test: 🧪 Add testing tools to pause and resume BSP container

ffarall:
In this PR:
1. Add functionality to resume a BSP container that was paused.
2. Add a test case that uses such functionality to prevent two BSPs from volunteering to a new storage request.
3. Add helper function to send new storage request.


timbrinded (approved):
LGTM

PR #171, opened 2024-08-27
fix: :bug: Solve next expiration block being pushed every time it was called by splitting queues

ffarall:


pallets/file-system/src/lib.rs
```diff
@@ -261,21 +261,40 @@ pub mod pallet {
...
     >;
 
     /// A pointer to the earliest available block to insert a new storage request expiration.
     ///
     /// This should always be greater or equal than current block + [`Config::StorageRequestTtl`].
     #[pallet::storage]
-    #[pallet::getter(fn next_available_expiration_insertion_block)]
-    pub type NextAvailableExpirationInsertionBlock<T: Config> =
+    #[pallet::getter(fn next_available_storage_proof_expiration_block)]
+    pub type NextAvailableStorageProofExpirationBlock<T: Config> =
```
snowmead, line 289: Why not call this `NextAvailableStorageRequestExpirationBlock`

pallets/file-system/src/utils.rs
```diff
@@ -1117,27 +1117,62 @@ where
...
+    /// Compute the next block number to insert an expiring storage request.
     ///
-    /// This function attempts to insert an expiration item at the next available block after the given TTL starting from
-    /// the current [`NextAvailableExpirationInsertionBlock`].
-    pub(crate) fn queue_expiration_item(
-        expiration_item_ttl: BlockNumberFor<T>,
-        expiration_item: ExpiredItems<T>,
+    /// This function attempts to insert a storage request expiration at the next available block starting from
+    /// the current [`NextAvailableStorageProofExpirationBlock`].
+    pub(crate) fn enqueue_storage_request_expiration(
```
snowmead, line 1124: I would merge `enqueue_storage_request_expiration` and `enqueue_file_deletion_request_expiration` functions since they do exactly the same thing and have the function accept the expiration item, TTL and next available block. Keeping the `ExpiredItems` enum and implementing 2 functions: `get_ttl` and `get_next_available_expiration_block` which matches on each variant and returns the appropriate values.

snowmead (changes requested):

snowmead (approved):

PR #173, opened 2024-08-27
Feat: persistent state store for BlockchainService

links234:


client/blockchain-service/src/handler.rs
```diff
@@ -611,8 +627,42 @@ impl BlockchainService {
...
+                    continue;
+                }
+                Err(e) => {
+                    error!(target: LOG_TARGET, "Error fetching block hash for block #{}: {:?}", block_number, e);
+                    continue;
+                }
+            };
+
+            self.process_block_import(&block_hash, &block_number).await;
         }
```
ffarall, line 667: Shouldn't this go in utils.ts?

ffarall (approved):
Approved with some comments

PR #174, opened 2024-08-27
feat: :sparkles: Add runtime API for the PaymentStreams pallet

TDemeco:
The runtime API added to the PaymentStreams pallet has a single function that, given a Provider ID and a balance threshold, returns all users that currently owe more than the threshold to the Provider. Useful for BSPs to check which users they should charge


a-moreira (approved):
Looks good to me!
Just a nit: for simplicity I'd call the method 
```
fn get_users_with_debt(provider_id: &ProviderIdFor<Runtime>, min_debt: Balance) -> ...
```

ffarall (approved):
Neat job! You might want to add this new runtime API to the types bundle package to be used by TS tests though.

PR #175, opened 2024-08-28
test: 🧪 Improve test writing DX 

timbrinded:
In an attempt to improve the DX of writing BSPNet tests, here is an experimental PR which adds a new `describe()` syntax for writing test cases.

The intention is:
- To reduce the boilerplate in writing a test
- To collect logic , spread around different files, to a single file
- To allow for simplified way to do paramterized tests
- To give top level options for skip, timeout, only


test/util/bspNet/types.ts
```diff
@@ -116,3 +118,62 @@ export interface FileMetadata {
...
   fileSize: number;
 }
+
+export type BspNetConfig = {
+  noisy: boolean;
+  rocksdb: boolean;
+  capacity?: bigint;
+};
+
+// TODO: Add DOcs here
```
ffarall, line 128: I subscribe to this

ffarall (approved):

PR #177, opened 2024-08-28
feat: ✨ logic to handle users without funds

TDemeco:
This PR:
- Modifies the `payment-streams` pallet to handle insolvent users correctly when Providers try to charge them.
  - The idea is users have a grace period in ticks per payment stream equivalent to the deposit they had to pay when the corresponding payment stream was created (set in the `NewStreamDeposit` constant on the pallet's configuration) to fund their account enough to pay for their unpaid payment streams.
  - If, after this grace period, the same Provider tries to charge the payment stream and the user still doesn't have the funds to pay for it, it is automatically flagged as a UserWithoutFunds.
  - The Provider then receives the funds that the user originally deposited, and that payment stream is deleted, as it is considered that the Provider has been repaid by the deposit.
  - ANY Provider that has a payment stream with this user can now try to charge it and, since it has been flagged as without funds, the Provider will get the deposit and the payment stream will be deleted.
  - The user has two choices to be unflagged as a user without funds:
    - Wait until all its Providers have tried charging him, which would mean all its payment streams have been deleted and he lost all deposits.
    - Execute a new extrinsic called `pay_outstanding_debt`, which pays all remaining payment streams using the user's deposits. This choice allows the user to recoupe some of his tokens that were deposited since the Providers are only paid what they are owed instead of the whole deposit. This incentive makes it so the now non-useful storage from the runtime of the payment streams of this user is cleared quickly. The weight cost of the iteration of this deletion in the runtime will probably be lower than what the user would lose if it doesn't execute it.
- Adds a new trait, `ReadUserSolvencyInterface` that's implemented by the `payment-streams` pallet, which allows other pallets to check whether a user is currently flagged as insolvent.
- Implements a new extrinsic in the `file-system` pallet called `stop_storing_for_insolvent_user`, which allows any Storage Provider to stop storing a file from a user that is currently flagged as insolvent without paying any penalty costs. Its logic is similar to `bsp_request_stop_storing` and `bsp_confirm_stop_storing`, with the difference that the SP does not have to wait since its root and capacity used are updated in the same function run.


pallets/file-system/src/lib.rs
```diff
@@ -840,7 +855,51 @@ pub mod pallet {
...
 
+        /// Executed by a SP to stop storing a file from an insolvent user.
+        ///
+        /// This is used when a user has become insolvent and the SP needs to stop storing the files of that user, since
+        /// it won't be getting paid for it anymore.
+        /// The validations are similar to the ones in the `bsp_request_stop_storing` and `bsp_confirm_stop_storing` extrinsics, but the SP doesn't need to
+        /// wait for a minimum amount of blocks to confirm to stop storing the file nor it has to be a BSP.
         #[pallet::call_index(9)]
+        #[pallet::weight(10_000 + T::DbWeight::get().reads_writes(1,1).ref_time())]
+        pub fn stop_storing_for_insolvent_user(
```
ffarall, line 866: There should be some testing for this extrinsic.

pallets/payment-streams/src/lib.rs
```diff
@@ -651,6 +656,41 @@ pub mod pallet {
...
+        /// 6. Unflag the user as without funds.
+        ///
+        /// Emits a 'UserSolvent' event when successful.
+        ///
+        /// Notes: this extrinsic iterates over all payment streams of the user and charges them, so it can be expensive in terms of weight.
+        /// The fee to execute it should be high enough to compensate for the weight of the extrinsic, without being too high that the user
+        /// finds more convenient to wait for Providers to get its deposits one by one instead.
+        #[pallet::call_index(7)]
+        #[pallet::weight(Weight::from_parts(100_000, 0) + T::DbWeight::get().reads_writes(1, 1))]
+        pub fn pay_outstanding_debt(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
```
ffarall, line 681: There should be some testing for this function.

pallets/payment-streams/src/utils.rs
```diff
@@ -550,180 +552,389 @@ where
...
+                            // that the user deposited when it was created), consider it as without funds.
+                            if ticks_since_out_of_funds >= T::NewStreamDeposit::get() {
+                                Self::manage_user_without_funds(
+                                    &provider_id,
+                                    &user_account,
+                                    &PaymentStream::FixedRatePaymentStream(
+                                        fixed_rate_payment_stream,
+                                    ),
+                                )?;
+                            }
```
ffarall, line 626: Why not inserting this user as without funds here?

pallets/payment-streams/src/utils.rs
```diff
@@ -550,180 +552,389 @@ where
...
+                            // that the user deposited when it was created), consider it as without funds.
+                            if ticks_since_out_of_funds >= T::NewStreamDeposit::get() {
+                                Self::manage_user_without_funds(
+                                    &provider_id,
+                                    &user_account,
+                                    &PaymentStream::DynamicRatePaymentStream(
+                                        dynamic_rate_payment_stream,
+                                    ),
+                                )?;
+                            }
```
ffarall, line 764: Same here, why not inserting it as without fund?

pallets/payment-streams/src/utils.rs
```diff
@@ -838,6 +1049,82 @@ where
...
+            "Provider should exist and have a payment account if its ID exists.",
+            Error::<T>::ProviderInconsistencyError
+        );
+        T::NativeBalance::release(
+            &HoldReason::PaymentStreamDeposit.into(),
+            &user_account,
+            deposit,
+            Precision::Exact,
+        )?;
+        T::NativeBalance::transfer(
```
ffarall, line 1083: What if what the user owed to this Provider is less than the deposited amount?

pallets/payment-streams/src/utils.rs
```diff
@@ -550,180 +552,389 @@ where
...
         }
 
-        // If the dynamic-rate payment stream exists:
-        if let Some(dynamic_rate_payment_stream) = dynamic_rate_payment_stream {
-            // Calculate the difference between the last charged price index and the price index at the last chargeable tick
-            // Note: If the last chargeable price index is less than the last charged price index, we charge 0 to the user, because that would be an impossible state.
+        Ok(total_amount_charged)
+    }
+
+    pub fn do_pay_outstanding_debt(user_account: &T::AccountId) -> DispatchResult {
```
ffarall, line 838: This is releasing the held funds from the user, right?

ffarall: What happens if the user's deposit is greater than what is owed to that Provider?

ffarall (changes requested):

a-moreira (approved):
Nice, very well documented!

a-moreira (approved):

pallets/file-system/src/lib.rs
TDemeco: Added!

pallets/payment-streams/src/lib.rs
TDemeco: Added as well!

pallets/payment-streams/src/utils.rs
TDemeco: The user is inserted as without funds in the `manage_user_without_funds` function

pallets/payment-streams/src/utils.rs
TDemeco: Same as above!

pallets/payment-streams/src/utils.rs
TDemeco: Changed so the user is only charged what it owes, the rest goes to that user's free balance

pallets/payment-streams/src/utils.rs
TDemeco: Same as above!

pallets/file-system/src/tests.rs
```diff
@@ -3972,6 +3972,593 @@ mod compute_threshold {
...
+            assert_eq!(
+                pallet_storage_providers::BackupStorageProviders::<Test>::get(bsp_id)
+                    .expect("BSP should exist in storage")
+                    .capacity_used,
+                0
+            );
+        });
+    }
+
+    // TODO: Not sure if needed, since we might not store any MSP info in the future
```
ffarall, line 4163: Yes, we still store the bucket roots, we just don't require proofs from them regularly.

ffarall (approved):

pallets/file-system/src/tests.rs
TDemeco: Great, when the `msp_accept_storage_request` PR is merged I'll add this test using that.

PR #178, opened 2024-08-29
feat: BSP charge fees task

a-moreira:
Adds a new task to charge users with debt with the BSP.


pallets/payment-streams/src/mock.rs
```diff
@@ -270,6 +270,10 @@ impl ExtBuilder {
         .assimilate_storage(&mut t)
         .unwrap();
 
+        crate::GenesisConfig::<Test> { current_price: 1 }
```
links234, line 273: What is this used for?

ffarall (approved):
Well done!

links234 (approved):
Great work!

pallets/payment-streams/src/mock.rs
TDemeco: I added this because the price of the dynamic-rate payment streams were initializing at 0 (since we don't have the equation for their price defined yet) and that was making testing more difficult than what it needed to be

PR #179, opened 2024-08-30
feat: Forest Storage Handler - Supporting MSP bucket storage

snowmead:
Enable support for MSP bucket storage, allowing an MSP to track multiple forests at once unlike the previous implementation which tracked a single forest supporting BSP execution. 

New `ForestStorageHandler` 

Changes:

- `StorageHubHandler` now holds an implementation of `ForestStorageHandler` instead of a simple implementation of `ForestStorage`.
- Refactors the builder pattern to support role based and storage layer combinations; abstracting specific storage layer implementations from the service node calls.
- Add typesafe `ForestStorageHandler` abstraction in tasks.

#### Some things to consider changing later

Creating distinct role based execution parameters (i.e. split `StorageHubHandler`). We could have `StorageHubProviderHandler` and `StorageHubUserHandler` with distinct fields and parameters necessary for their own tasks executions.


node/src/tasks/mock_bsp_volunteer.rs
```diff
@@ -5,49 +5,49 @@ use shc_actors_framework::event_bus::EventHandler;
...
-    storage_hub_handler: StorageHubHandler<FL, FS>,
+    storage_hub_handler: StorageHubHandler<FL, FSH>,
 }
 
-impl<FL, FS> Clone for BspVolunteerMockTask<FL, FS>
+impl<FL, FSH> Clone for BspVolunteerMockTask<FL, FSH>
 where
-    FL: Send + Sync + FileStorage<StorageProofsMerkleTrieLayout>,
-    FS: Send + Sync + ForestStorage<StorageProofsMerkleTrieLayout> + 'static,
+    FL: FileStorage<StorageProofsMerkleTrieLayout> + Send + Sync,
```
links234, line 25: I think it's worth moving the `Send`, `Sync`, `'static' and `Clone` bounds in the `FileStorage` so that you don't have to specify it everywhere. What do you think?

node/src/tasks/bsp_submit_proof.rs
```diff
@@ -261,10 +271,10 @@ where
     }
 }
 
-impl<FL, FS> BspSubmitProofTask<FL, FS>
+impl<FL, FSH> BspSubmitProofTask<FL, FSH>
 where
     FL: FileStorage<StorageProofsMerkleTrieLayout> + Send + Sync,
-    FS: ForestStorage<StorageProofsMerkleTrieLayout> + Send + Sync + 'static,
+    FSH: ForestStorageHandler<Key = NoKey> + Clone + Send + Sync + 'static,
```
links234, line 277: This trait could be "wrapped" as a `trait BspForestStorageHandler: ForestStorageHandler<Key = NoKey> + Clone + Send + Sync + 'static` so it's more clear what this is about.

node/src/services/builder.rs
```diff
@@ -98,105 +163,170 @@ where
...
     }
 
-    /// Add a new [`FileStorage`] to the builder.
-    ///
-    /// This is the set of tools that allows a StorageHub node to store files as Merkle Patricia
-    /// Tries, in the way that the StorageHub protocol specifies.
-    pub fn with_file_storage(&mut self, file_storage: Arc<RwLock<FL>>) -> &mut Self {
-        self.file_storage = Some(file_storage);
-        self
+    fn build_handler(
```
links234, line 166: *(nitpick)*

This could also be typed so that the compiler doesn't let you build something invalid or without missing parts.

links234 (approved):
Great job, I like how it turned out!

I only left a couple of nitpicks/comments.

node/src/cli.rs
```diff
@@ -54,7 +54,6 @@ pub enum ProviderType {
     Msp,
     /// Backup Storage Provider
     Bsp,
-    /// TODO: This is temporary. Remove it once we have implemented all the User related code.
```
ffarall, line 57: Maybe replace the comment?

client/rpc/src/lib.rs
```diff
@@ -92,7 +93,7 @@ pub trait StorageHubClientApi {
     ) -> RpcResult<SaveFileToDisk>;
 
     #[method(name = "getForestRoot")]
-    async fn get_forest_root(&self) -> RpcResult<H256>;
+    async fn get_forest_root(&self, key: String) -> RpcResult<H256>;
```
ffarall, line 96: Can we make this parameter optional?

ffarall (approved):
Remarkable work on the refactor

ffarall (approved):

node/src/services/builder.rs
```diff
@@ -6,52 +6,116 @@ use sp_keystore::KeystorePtr;
 use std::sync::Arc;
 use tokio::sync::RwLock;
 
+use super::{
+    forest_storage::{ForestStorageCaching, ForestStorageSingle},
+    handler::StorageHubHandler,
+};
+use crate::tasks::{BspForestStorageHandlerT, FileStorageT, MspForestStorageHandlerT};
```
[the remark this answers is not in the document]
snowmead: I will put it at the bottom, but I merge imports based on rust analyzers recommendation of optimal import paths. Should we be adding imports manually to all be in `crate` then? Because I can't use `super` to fetch the task traits for obvious reasons.

node/src/services/builder.rs
```diff
@@ -6,52 +6,116 @@ use sp_keystore::KeystorePtr;
...
+impl StorageTypes for (MspProvider, RocksDbStorageLayer) {
+    type FL = RocksDbFileStorage<StorageProofsMerkleTrieLayout, kvdb_rocksdb::Database>;
+    type FSH = ForestStorageCaching<
+        Vec<u8>,
+        RocksDBForestStorage<StorageProofsMerkleTrieLayout, kvdb_rocksdb::Database>,
+    >;
+}
+
+// TODO: Implement default empty implementations for the storage layers.
+/// There is no default empty implementation for [`FileStorage`] and [`ForestStorageHandler`] so
```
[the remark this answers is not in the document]
snowmead: > The user still needs a file storage implementation to work.

Updated the comment.

node/src/services/forest_storage.rs
```diff
@@ -0,0 +1,125 @@
...
+            .write()
+            .await
+            .insert(key.clone(), fs.clone());
+        fs
+    }
+
+    async fn remove_forest_storage(&mut self, key: &Self::Key) {
+        self.fs_instances.write().await.remove(key);
+    }
+}
```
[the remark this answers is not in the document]
snowmead: Will decide on this at another point in time whether there is a better spot to move this to.

PR #181, opened 2024-09-03
fix: :bug: change how file tries are constructed to avoid sp_trie issue

TDemeco:
File tries leaves are now chunk ID + chunk data instead of only chunk data, to avoid sp_trie optimizing the trie into fewer leaves and generating issues when users send proofs to BSPs

### DIFF
--- a/test/util/bspNet/consts.ts
+++ b/test/util/bspNet/consts.ts
@@ -29,7 +29,7 @@ export const TEST_ARTEFACTS = {
   "res/adolphus.jpg": {
     size: 416400n,
     checksum: "739fb97f7c2b8e7f192b608722a60dc67ee0797c85ff1ea849c41333a40194f2",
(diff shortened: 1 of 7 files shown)


test/util/bspNet/consts.ts
```diff
@@ -29,7 +29,7 @@ export const TEST_ARTEFACTS = {
   "res/adolphus.jpg": {
     size: 416400n,
     checksum: "739fb97f7c2b8e7f192b608722a60dc67ee0797c85ff1ea849c41333a40194f2",
-    fingerprint: "0x9e86ecad3f86f52891acf9c0d47c5c5243a3592fc7bcf99811eb7db078997dcb"
+    fingerprint: "0x34eb5f637e05fc18f857ccb013250076534192189894d174ee3aa6d3525f6970"
```
ffarall, line 32: Do you think you can also change the fingerprints of the other files here? They will most likely cause failures down the line.

ffarall (approved):
Great job!

links234 (approved):
Great work!

PR #183, opened 2024-09-04
feat: ✨ Make BSPs catch up to proof submissions after downtime or initialisation

links234:
In this PR:
1. Use `proof_submission_catch_up` in initialisation of the node and in every block a BSP would have to submit a proof.
2. Add `QueryChallengePeriod` and `QueryNextChallengeTickForProvider` commands to the Blockchain Service.
3. Track the latest block processed in the Blockchain Service, to use it as trigger for launching initialisation checks and updates.
4. Avoid submitting proofs when the proof has already expired (i.e. it is for a tick that the runtime is no longer expecting).


client/blockchain-service/src/utils.rs
```diff
@@ -442,15 +456,15 @@ impl BlockchainService {
...
                     state_store_context.commit();
                 }
                 Err(TryRecvError::Closed) => {
                     error!(target: LOG_TARGET, "Forest root write task channel closed unexpectedly. Lock is released anyway!");
                     let state_store_context = self.persistent_state.open_rw_context_with_overlay();
                     state_store_context
-                        .access_value(&OngoingForestWriteLockTaskDataCf)
+                        .access_value(&OngoingProcessConfirmStoringRequestCf)
                         .delete();
                     state_store_context.commit();
```
ffarall, line 469: In here, if we release the lock because a `submit_proof` task just finished, would we be deleting a potential `OngoingProcessConfirmStoringRequest` that is not actually happening?

links234: Not really since we are always writing the ongoing task when we create a new channel. This seems like it since no matter what task we delete the OngoingProcessConfirmStoringRequest, but in reality there can be only 1 ongoing task that writes the forest, so when the lock is for a SubmitProof this is a noop.

ffarall: Might be worth adding it in a comment

ffarall (approved):

PR #186, opened 2024-09-05
feat: Consider remaining capacity - BSP volunteering and confirming

snowmead:
The runtime extrinsics for BSPs volunteering and confirming storage requests now error out if the BSPs remaining available storage capacity is insufficient to hold the the associated file size.

While a BSP can confirm multiple files in a single extrinsic call, it is treated as atomic, and will error out if any one of them fail, essentially rolling back all changes applied from previous succeeding confirmations.

To ensure BSPs avoid sending out volunteering and confirming extrinsics which would surely fail due to insufficient available storage capacity, they will attempt to increase their storage capacity by sending a `change_capacity` extrinsic call to cover the file size. The provider `max_storage_capacity` parameter allows the BSP to set a cap on the amount of storage capacity they wish to increase up to. The increased storage capacity will saturate to the max storage capacity and skip volunteering if the new available capacity is less then the file size.

Some small features added:

New `ProviderConfig` held in `StorageHubHandler` to manage provider specific parameters with newly included `max_storage_capacity` and `jump_capacity` (additional cli options to configure these parameters).

Added various runtime APIS:
- `query_storage_provider_capacity`
- `query_available_storage_capacity`
- `query_earliest_change_capacity_block`

Removed the possibility to configure the maximum threshold value used for threshold calculations to avoid any potential mistakes setting it below the maximum value of the type itself (e.g. Threshold type (u32) would have the maximum threshold value be `u32::MAX`). 

Set all BSP starting reputation weight runtime configurations to `1`.


runtime/src/lib.rs
```diff
@@ -122,6 +125,9 @@ pub type Executive = frame_executive::Executive<
     AllPalletsWithSystem,
 >;
 
+/// Type representing the storage data units in StorageHub.
+pub type StorageDataUnit = u32;
```
ffarall, line 129: This feels like something that should be defined in `configs/mod.rs` and imported here 🤔

node/src/tasks/bsp_upload_file.rs
```diff
@@ -450,6 +441,140 @@ where
...
+            const GIB: u32 = 1 << 30;
+
+            // Calculate GiBs needed, rounding up
+            let gibs_needed = (event.size + GIB - 1) / GIB;
+
+            // Ensure at least 1 GiB is added
+            let gibs_to_add = max(gibs_needed, 1);
+
+            // Calculate bytes to add
+            let bytes_to_add = gibs_to_add * GIB;
```
ffarall, line 500: We might want to make the step configurable somehow.

node/src/tasks/bsp_upload_file.rs
```diff
@@ -450,6 +441,140 @@ where
...
+            .map_err(|e| {
+                error!(
+                    target: LOG_TARGET,
+                    "Failed to query available storage capacity: {:?}", e
+                );
+                anyhow::anyhow!("Failed to query available storage capacity: {:?}", e)
+            })?;
+
+        // Increase storage capacity if the available capacity is less than the file size.
+        if available_capacity < event.size {
```
ffarall, line 458: This whole section of adding capacity is probably worth wrapping in a helper function, for better readability

ffarall (changes requested):
Excellent work! Only some small things to fix

node/src/cli.rs
```diff
@@ -93,7 +112,7 @@ pub struct ProviderConfigurations {
     pub storage_layer: Option<StorageLayer>,
 
     /// Storage location in the file system
-    #[clap(long, required_if_eq("storage-layer", "rocks-db"))]
+    #[clap(long, required_if_eq("storage_layer", "rocks-db"))]
```
ffarall, line 115: Most arguments in a substrate node follow the `this-casing`, why using a different kind?

node/src/cli.rs
```diff
@@ -82,6 +83,24 @@ pub struct ProviderConfigurations {
...
+    pub max_storage_capacity: Option<StorageDataUnit>,
+
+    /// Jump capacity (bytes).
+    ///
+    /// Default: 4294967295 bytes (4GiB)
+    #[clap(long, required_if_eq_any([
+        ("provider_type", "msp"),
+        ("provider_type", "bsp")
+    ]))]
+    pub jump_capacity: Option<StorageDataUnit>,
```
ffarall, line 102: We probably want to make these NOT optional and with default types. We can have our testing setup give them default values so that it is not cumbersome for us, but BSP operators should be aware of these configurations.

ffarall (approved):

node/src/cli.rs
snowmead: We have concluded the opposite from one of our meetings where we want to have it be required if you run the node as a storage provider and force node operators to be aware of these configurations to avoid any misconfigured storage capacities. 

It is only considered optional if the node is run as a user as of right now. It will fail and require you to provide these options otherwise. Also it didn't take me long to add these options to the CI. It is more clear as well IMO that we have this explicitely set in the CI.

snowmead: I'll update the comment to reflect.

node/src/cli.rs
snowmead: This was changed due to this error:
```
Command storage-hub-node: Argument or group 'storage-layer' specified in 'required_if_eq*' for 'storage_path' does not exist
```

node/src/cli.rs
ffarall: Sorry, my comment was not clear. I meant what you said, in case the provider types is "msp" or "bsp", then they should be required. I thought that was not the case here in the code, but it is. So great 👌🏼

node/src/cli.rs
ffarall: My bad, I misinterpreted some stuff. All good.

PR #189, opened 2024-09-10
test 🧪: Added threshold tests

timbrinded:
Scenarios cover when multiple BSPs volunteer and whether the runtime accepts their request or not. We briefly cover reputation here, but requires more coverage when the underlying feature is implemented further. 

> [!NOTE]  
> These scenarios are a bit naive and coarse-grained. But they will at least tell us if we seriously break something which affect BSP volunteer thresholds. 
>
> The rust tests should be relied on for verifying the curves here.


test/suites/integration/bsp/bsp-thresholds.test.ts
```diff
@@ -0,0 +1,175 @@
...
+
+      assert(
+        repAfter > repBefore,
+        "Reputation should increase after successful storage request fufilled"
+      );
+      console.log(`Reputation increased from ${repBefore} to ${repAfter}`);
+    });
+
+    it("zero reputation can still volunteer and be accepted", async () => {
+      // Create a new BSP and onboard with no reputation
```
ffarall, line 90: This actually shows something is wrong. A BSP shouldn't have 0 reputation. They start with 1. Good that you raised this.

test/suites/integration/bsp/bsp-thresholds.test.ts
```diff
@@ -0,0 +1,175 @@
...
+      );
+
+      assert(
+        filtered.length === 1,
+        "Zero reputation BSP should be able to volunteer and be accepted"
+      );
+      await api.docker.stopBspContainer("sh-bsp-down");
+    });
+
+    // Not sure if this is good test? the times are very dependent on threshold created
```
ffarall, line 121: I think it is a good test, but you can make it less "manual" using the runtime API `query_earliest_file_volunteer_block` which would tell you when a BSP can volunteer for a block.

test/suites/integration/bsp/bsp-thresholds.test.ts
```diff
@@ -0,0 +1,175 @@
...
+      const events = await api.query.system.events();
+      const matchedEvents = api.assert.eventMany("fileSystem", "AcceptedBspVolunteer", events); // T1
+
+      const filtered = matchedEvents.filter(
+        ({ event }) =>
+          (api.events.fileSystem.AcceptedBspVolunteer.is(event) && event.data.bspId.toString()) ===
+          ShConsts.BSP_THREE_ID
+      );
+
+      // Verify that the BSP with reputation is prioritised over the zero reputation
```
ffarall, line 170: Here again, the other BSP shouldn't have 0 reputation in a real world scenario.

test/suites/integration/bsp/debt-collection.test.ts
```diff
@@ -117,7 +117,7 @@ describeBspNet(
       }
 
       // Wait for tasks to execute and for the BSPs to submit proofs.
-      await sleep(500);
+      await sleep(2500);
```
ffarall, line 120: I assume this is because it failed sometimes in the CI?

ffarall (approved):
Approved with comments. Neat job!

test/suites/integration/bsp/bsp-thresholds.test.ts
timbrinded: ahh very convenient, thanks!

test/suites/integration/bsp/bsp-thresholds.test.ts
timbrinded: Replaced 0 rep with min (1), and changed default bsp to 5

test/suites/integration/bsp/bsp-thresholds.test.ts
timbrinded: comparing against 0 reputation is useful as it is lower than standard reputation

timbrinded: fixed

test/suites/integration/bsp/debt-collection.test.ts
timbrinded: indeed, these runners are very slow

PR #191, opened 2024-09-12
feat: ✨ add RPC methods for file and forest access and proof generation

TDemeco:
This PR:
- Adds RPC methods to:
  - Check if a particular file is in the Forest storage.
  - Check if a particular file is in the File storage.
  - Get the metadata of a file in Forest storage (returns None if the file doesn't exist)
  - Generate a SCALE-encoded forest proof for a given list of keys.
- Do the ToDo to change the `get_metadata` function from the file manager to return an Option.
- Add a `get_file_metadata` function to the forest manager to get the encoded file metadata from a forest. Note: this is returning an empty value since we are not currently storing the file metadata in the forest. ~This will be updated in another PR which will be linked here when it's open.~ done in #194


client/common/src/types.rs
```diff
@@ -93,8 +94,10 @@ impl<K, D: Debug> Proven<K, D> {
 }
 
 /// Proof of file key(s) in the forest trie.
+#[derive(Clone, Encode, Decode, Debug)]
 pub struct ForestProof<T: TrieLayout> {
     /// The file key that was proven.
```
snowmead, line 99:
```suggestion
    /// The file keys that were proven.
```

client/common/src/types.rs
```diff
@@ -93,8 +94,10 @@ impl<K, D: Debug> Proven<K, D> {
...
+    #[codec(skip)]
```
snowmead, line 100: Curious why this is added?

client/rpc/src/lib.rs
```diff
@@ -272,6 +297,75 @@ where
...
+            .get_metadata(&file_key)
+            .map_err(into_rpc_error)?
+        {
+            None => return Ok(false),
+            Some(_) => Ok(true),
+        }
+    }
+
+    // Note: this method could use either the file storage or the forest storage, but it's using the forest storage.
+    // WARNING: Right now, forests don't have the file metadata saved to them, so don't expect to get the file
```
snowmead, line 329: We could also add a fallback to either get it from forest or file manager.

snowmead (approved):
LGTM, just a few comments.

client/file-manager/src/traits.rs
```diff
@@ -134,7 +134,7 @@ pub trait FileStorage<T: TrieLayout>: 'static {
 
     // TODO: Return Result<Option> instead of Result only
```
ffarall, line 135: Shouldn't this be removed?

client/rpc/src/lib.rs
```diff
@@ -272,6 +297,75 @@ where
         Ok(read_fs.root())
     }
 
+    async fn is_file_in_forest(&self, key: Option<String>, file_key: H256) -> RpcResult<bool> {
```
ffarall, line 300: Should we call the `key` parameter `forest_key`? So that it is clearer. `key` alone feels a bit confusing on what key is it. Given that we have so many keys.

client/common/src/types.rs
ffarall: I believe it is because he encodes/decodes this struct only for the RPC method, in which case it doesn't care about the `proven` field. Is that correct?

types-bundle/src/rpc.ts
```diff
@@ -49,6 +49,59 @@ export const rpcDefinitions: Record<string, Record<string, DefinitionRpc | Defin
...
+      params: [
+        {
+          name: "file_key",
+          type: "H256"
+        }
+      ],
+      type: "bool"
+    },
+    getFileMetadata: {
+      description: "Get the metadata of a file.",
```
ffarall, line 77: Here it is probably worth specifying if it is from Forest or File Storage.

ffarall (approved):
Approved with some small comments

client/common/src/types.rs
TDemeco: Yep! That's exactly right.

client/rpc/src/lib.rs
TDemeco: Do you think we should have two methods? One from the Forest storage and one from the File storage?
If it helps I believe we don't really need access to metadata that's only on the File storage, since for now the idea is to only use it when deleting already-stored files from an insolvent user (which should be in the Forest storage). Although now that I think about it, we also have storage requests that might be pending...

client/file-manager/src/traits.rs
TDemeco: Yep, already implemented the change so that TODO is resolved. I'll delete it

client/rpc/src/lib.rs
TDemeco: I wanted to keep the naming convention of the other RPC methods. I agree with you though, it was a bit confusing for me just calling it `key`. Should I change all of them?

TDemeco: Went ahead and changed them all 😬

PR #192, opened 2024-09-12
fix: 🩹 add extra required leafs in proof generation to be able to use `apply_delta`

TDemeco:
This PR:
- Adds, for each leaf of a proof, the previous and next one to make sure the generated proof is complete enough to be able to use `apply_delta` for file removals.
- Adds the test to check that this logic is working.
- Completes the proof submission integration test except for actually checking if a file is in storage (which could be done after #191 is merged)

### DIFF
--- a/client/forest-manager/src/prove.rs
+++ b/client/forest-manager/src/prove.rs
@@ -44,6 +44,10 @@ where
     let next = iter.next().transpose()?;
     let prev = iter.next_back().transpose()?;
 
+    // This is just done to allow the `apply_delta` function to remove files from the trie without failing because of an incomplete database.
+    iter.next().transpose()?;
+    iter.next_back().transpose()?;
+
     match (prev, next) {
         // Scenario 1: Exact match
         (_, Some((key, _))) if challenged_file_key.as_ref() == key => Ok(Proven::new_exact_key(
--- a/client/forest-manager/src/prove.rs
+++ b/client/forest-manager/src/prove.rs
@@ -44,16 +44,17 @@ where
     let next = iter.next().transpose()?;
     let prev = iter.next_back().transpose()?;
 
-    // This is just done to allow the `apply_delta` function to remove files from the trie without failing because of an incomplete database.
-    iter.next().transpose()?;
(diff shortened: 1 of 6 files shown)


client/forest-manager/src/rocksdb.rs
```diff
@@ -503,6 +503,75 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_generate_proof_includes_neighbor_keys() {
```
snowmead, line 507: Do you think it's worth it to create a test or modify this one to include the apply delta execution?

snowmead (approved):
Beast!

client/forest-manager/src/prove.rs
```diff
@@ -44,6 +44,10 @@ where
     let next = iter.next().transpose()?;
     let prev = iter.next_back().transpose()?;
 
+    // This is just done to allow the `apply_delta` function to remove files from the trie without failing because of an incomplete database.
+    iter.next().transpose()?;
+    iter.next_back().transpose()?;
```
ffarall, line 49: Wouldn't it be more efficient to handle this in the `Scenario 1` case?

ffarall:
Shouldn't you also update the commented test `mutate_root_remove_multiple_keys_success` in `primitives/forest-verifier`?

TDemeco:
> Shouldn't you also update the commented test `mutate_root_remove_multiple_keys_success` in `primitives/forest-verifier`?

Not really, since the forest verifier is not using the forest manager for its tests. I could make it so the proof generation in the tests of the forest verifier primitive crate uses the forest manager method if you think that's the way to go, but personally I don't think mixing those up is correct.

client/forest-manager/src/rocksdb.rs
TDemeco: Just added it!

client/forest-manager/src/prove.rs
TDemeco: Yep, changed.

ffarall (approved):

PR #193, opened 2024-09-16
Feat: automatic tip increase on retry

links234:
In this PR:
- Add `RetryStrategy` which defines the number of retries and how the tip will increase.
- Add `submit_extrinsic_with_retry` helper method for `BlockchainService` actor interface.
- Expose via runtime the worst case scenario for missing a proof.
- Submit proof and confirm storing now use retry. The submit proof maximum tip is the worst case scenario of missing the proof.
- Add CLI argument `--extrinsic-retry-timeout S`
- Update TS helpers to properly inject/pass extrinsic timeout param to bsps


client/blockchain-service/src/commands.rs
```diff
@@ -576,4 +574,75 @@ impl BlockchainServiceInterface for ActorHandle<BlockchainService> {
...
+        retry_strategy: RetryStrategy,
+    ) -> Result<()> {
+        let call = call.into();
+
+        for retry_count in 0..=retry_strategy.max_retries {
+            let tip = retry_strategy.compute_tip(retry_count);
+            let mut transaction = self
+                .send_extrinsic(call.clone(), Tip::from(tip as u128))
+                .await?
+                .with_timeout(retry_strategy.timeout);
```
snowmead, line 631: Should we check if they have enough funds to cover the tip?

snowmead (approved):
Beast!

client/blockchain-service/src/types.rs
```diff
@@ -128,3 +127,93 @@ pub enum ExtrinsicResult {
...
+    fn compute_tip_multiplier(&self) -> f64 {
+        (self.base_multiplier.ln() / self.max_retries as f64).exp()
+    }
+
+    pub fn compute_tip(&self, retry_count: u32) -> f64 {
+        let multiplier = self.compute_tip_multiplier();
+
+        // Calculate the geometric progression factor for each retry
+        let factor = (multiplier.powf(retry_count as f64) - 1.0)
+            / (multiplier.powf(self.max_retries as f64) - 1.0);
```
ffarall, line 202: Optimisation:
```rust
multiplier.powf(self.max_retries as f64)
```

Is just `base_multiplier`

ffarall: Is it ok that this function doesn't guard against `retry_count` being grater than `max_retires`?

client/blockchain-service/src/commands.rs
```diff
@@ -276,11 +300,13 @@ impl BlockchainServiceInterface for ActorHandle<BlockchainService> {
     async fn send_extrinsic(
         &self,
         call: impl Into<storage_hub_runtime::RuntimeCall> + Send,
+        tip: Tip,
```
ffarall, line 303: I think I would have made this optional, but also ok having it like this.

test/util/bspNet/helpers.ts
```diff
@@ -109,16 +111,26 @@ export const runSimpleBspNet = async (bspNetConfig: BspNetConfig) => {
     if (bspNetConfig.rocksdb) {
       file = "local-dev-bsp-rocksdb-compose.yml";
     }
-    if (bspNetConfig.noisy) {
-      file = "noisy-bsp-compose.yml";
-    }
```
ffarall, line 114: Why removing this?

client/blockchain-service/src/commands.rs
```diff
@@ -22,16 +23,21 @@ use shc_common::types::{
 };
 use storage_hub_runtime::{AccountId, Balance, StorageDataUnit};
 
-use crate::{
+use crate::types::{RetryStrategy, Tip};
+
+use super::{
```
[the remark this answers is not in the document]
links234: To be frank I usually just leave it to rust lint to sort it out. I can tidy it up a little bit.

client/blockchain-service/src/commands.rs
```diff
@@ -22,16 +23,21 @@ use shc_common::types::{
...
     handler::BlockchainService,
     transaction::SubmittedTransaction,
     types::{ConfirmStoringRequest, Extrinsic, ExtrinsicResult, SubmitProofRequest},
 };
 
+const LOG_TARGET: &str = "blockchain-service-interface";
```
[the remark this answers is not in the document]
links234: I don't feel strongly about either. My reasoning was that the errors/traces/logs in the helper functions are not technically in the `blockchain-service` and errors with that tag would be more bad.

Let me know what you think.

client/blockchain-service/src/commands.rs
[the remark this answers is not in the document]
ffarall: Yeah me too, I figured that was the case. Whenever I see this I flag it or fix it, to keep things tidy. Sometimes it goes through me too.

client/blockchain-service/src/commands.rs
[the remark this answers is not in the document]
ffarall: Sounds reasonable, leave it like that 👌🏼

test/suites/integration/bsp/automatic-tipping.test.ts
```diff
@@ -0,0 +1,76 @@
...
+      const confirmPending = await userApi.rpc.author.pendingExtrinsics();
+      strictEqual(
+        confirmPending.length,
+        1,
+        "There should be one pending extrinsic from BSP (confirm store)"
+      );
+
+      await sleep(6000); // wait for the bsp to send confirm again
+      const confirmPending2 = await userApi.rpc.author.pendingExtrinsics();
+      const expectedRetryCount = 4;
```
timbrinded, line 62: We should be using assert.extrinsicPresent() so that we are filtering for a particular ext. otherwise this test will break if extra exts are added to BSP tasks

test/suites/integration/bsp/automatic-tipping.test.ts
```diff
@@ -0,0 +1,76 @@
...
+      strictEqual(
+        confirmPending2.length,
+        expectedRetryCount,
+        "There should be 4 pending extrinsic from BSP (confirm store) with increasing tip"
+      );
+      for (let i = 1; i < confirmPending2.length; ++i) {
+        assert(
+          confirmPending2[i].tip > confirmPending2[i - 1].tip,
```
timbrinded, line 70: If this is scale encoded, don't we need to add .toBigInt() on tip ?

test/suites/integration/bsp/automatic-tipping.test.ts
```diff
@@ -0,0 +1,76 @@
...
+    before(async () => {
+      userApi = await createUserApi();
+    });
+
+    it("Confirm storing failure results in increased tip", async () => {
+      const source = "res/whatsup.jpg";
+      const destination = "test/whatsup.jpg";
+      const bucketName = "nothingmuch-2";
+
+      const newBucketEventEvent = await userApi.createBucket(bucketName);
```
timbrinded, line 20: This can all be replaced with .file.newStorageRequest()

timbrinded (changes requested):

test/util/bspNet/helpers.ts
links234: By mistake :). Adding it back.

ffarall (approved):

timbrinded (approved):
LGTM

PR #194, opened 2024-09-16
feat: ✨ add file metadata to leafs in forest's tries

TDemeco:
This PR:
- Makes it so each leaf (file) in the forest of a Provider now has as value that file's metadata instead of being empty. This is useful to keep track of Provider's capacity during file deletions and to allow Providers to know if they are storing files for a specific user without needed an external indexer.
- Modifies the `apply_delta` function of the ForestVerifier to also use the file's metadata as value when getting the new root for a Provider.
- Adds a method to the ForestStorage trait that returns the metadata of a file if it's in the trie, and tests for it.
- Adds tests to the in-memory implementation of the ForestStorage trait.
- Fixes some std issues that were happening when trying to run the tests for the forest manager on their own.


primitives/forest-verifier/src/lib.rs
```diff
@@ -238,15 +246,33 @@ where
...
+                        match node_value {
+                            trie_db::Value::Inline(value) => Some(value.to_vec()),
+                            trie_db::Value::Node(value_hash) => {
+                                let value = trie
+                                    .get(value_hash.as_ref())
+                                    .map_err(|_| "Failed to get value from trie.")?;
+                                value.map(|v| v.to_vec())
+                            }
+                            trie_db::Value::NewNode(_, value) => Some(value.to_vec()),
+                        }
```
snowmead, line 271: I believe you can do this.

```suggestion
                        match node_value {
                            trie_db::Value::Inline(value) | trie_db::Value::NewNode(_, value) => Some(value.to_vec()),
                            trie_db::Value::Node(value_hash) => {
                                let value = trie
                                    .get(value_hash.as_ref())
                                    .map_err(|_| "Failed to get value from trie.")?;
                                value.map(|v| v.to_vec())
                            }
                        }
```

primitives/traits/src/lib.rs
```diff
@@ -762,7 +770,14 @@ pub trait TrieProofDeltaApplier<H: sp_core::Hasher> {
...
         proof: &Self::Proof,
-    ) -> Result<(sp_trie::MemoryDB<H>, Self::Key), DispatchError>;
+    ) -> Result<
+        (
+            sp_trie::MemoryDB<H>,
+            Self::Key,
+            Vec<(Self::Key, Option<Vec<u8>>)>,
+        ),
+        DispatchError,
+    >;
```
snowmead, line 780: I would update the docs to reflect the new return type.

test/package.json
```diff
@@ -38,6 +38,7 @@
     "test:full": "node --no-deprecation --import tsx --test ./suites/zombie/**.spec.ts",
     "test:bspnet": "NODE_OPTIONS='--no-deprecation' pnpm tsx scripts/checkRunning.ts && node --no-deprecation --test-concurrency 1 --import tsx --test ./suites/integration/bsp/**.test.ts",
     "test:bspnet:only": "NODE_OPTIONS='--no-deprecation' pnpm tsx scripts/checkRunning.ts && node --no-deprecation --import tsx --test-concurrency 1 --test --test-only ./suites/integration/bsp/**.test.ts",
+	"test:bspnet:challenge-cycle": "NODE_OPTIONS='--no-deprecation' pnpm tsx scripts/checkRunning.ts && node --no-deprecation --test-concurrency 1 --import tsx --test ./suites/integration/bsp/challenge-cycle.test.ts",
```
snowmead, line 41: Do we need this?

snowmead (approved):
Overall LGTM, just a few changes mentioned. 👍🏼

primitives/forest-verifier/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
 use frame_support::sp_runtime::DispatchError;
+use scale_info::prelude::vec::Vec;
```
ffarall, line 4: Better to import it from `sp_std` in my opinion.

pallets/file-system/src/utils.rs
```diff
@@ -740,6 +765,16 @@ where
...
             )?;
 
+            // Get the file metadata to insert into the Provider's trie under the file key.
+            let file_metadata = storage_request_metadata.clone().to_file_metadata();
+            let encoded_trie_value = file_metadata.encode();
+            expect_or_err!(
+                file_keys_and_metadatas.try_push((file_key.0, encoded_trie_value)),
+                "Failed to push file key and metadata",
+                Error::<T>::FileMetadataProcessingQueueFull,
+                result
```
ffarall, line 775: Why shouldn't this ever fail?

pallets/file-system/src/utils.rs
```diff
@@ -482,11 +497,15 @@ where
             &file_proof,
         )?;
 
+        // Get the file metadata to insert into the bucket under the file key.
+        let file_metadata = storage_request_metadata.clone().to_file_metadata();
+        let encoded_trie_value = file_metadata.encode();
```
ffarall, line 502: We should think how to be careful to always encode consistently.

node/src/tasks/bsp_submit_proof.rs
```diff
@@ -161,6 +161,8 @@ where
                 .await
                 .ok_or_else(|| anyhow!("Failed to get forest storage."))?;
 
+            info!(target: LOG_TARGET, "Forest root of BSP: {:?}", fs.read().await.root());
```
ffarall, line 164: Do you intend these logs to stay here? Wouldn't it be too much maybe

ffarall (approved):

pallets/file-system/src/utils.rs
TDemeco: As long as we use the FileMetadata type, we will always encode and decode correctly since that's the type that derives the Encode and Decode traits. Do you have any other safeguards in mind that we could implement?

pallets/file-system/src/utils.rs
TDemeco: Both Bounded vectors are limited by the same constant (`file_keys_and_proofs` used in the for loop and `file_keys_and_metadatas` where we push the values) so the calls to `try_push` will never actually fill the vector.

node/src/tasks/bsp_submit_proof.rs
TDemeco: Ups, leftovers from debugging. I'll delete them

PR #197, opened 2024-09-18
feat: ✨ update provider's and payment stream's info when submitting proof with mutations

TDemeco:
This PR:
- Makes it so that, when a BSP submits a proof with a checkpoint challenge of a file deletion, the `apply_delta` correctly updates both the used capacity of that BSP in the StorageProviders pallet (decreasing it by the file's size) and the provided amount of the corresponding Payment Stream between the BSP and the owner of the file (also decreasing it by the file's size)
- Updates all StorageDataUnit types on the mocked runtimes to be u64 instead of u32, so it's consistent with the client expected type.
- Adds a PaymentStreamsInterface required type to the StorageProviders pallet, since it now has to modify payment streams when receiving a request from `submit_proof` to update a Provider. Note: this may be temporary since the ideal solution would probably be handling this from the FileSystem pallet that already has access to both the StorageProviders and PaymentStreams pallets (and act as a hub for them).


snowmead:
I would add tests for the new call from proofs dealer pallet to `update_provider_after_key_removal`.

ffarall (approved):

PR #198, opened 2024-09-18
feat: :sparkles: change `isFileInFileStorage` rpc return to be more granular

TDemeco:
This PR makes it so instead of returning a boolean (true if the file is found in the File storage, false if it's not), the `isFileInFileStorage` RPC returns three options:
- FileNotFound if the file wasn't found in the File storage.
- IncompleteFile with file metadata, the amount of chunks stored and the total amount of chunks that the file has, if the file is found in the File storage but is still missing chunks.
- Success with the file metadata if the file was found in the File storage and it's complete (not missing any chunks).


client/rpc/src/lib.rs
```diff
@@ -334,8 +347,24 @@ where
...
         {
-            None => return Ok(false),
-            Some(_) => Ok(true),
+            None => Ok(GetFileFromFileStorageResult::FileNotFound),
+            Some(file_metadata) => {
+                let stored_chunks = read_file_storage
+                    .stored_chunks_count(&file_key)
+                    .map_err(into_rpc_error)?;
+                let total_chunks = file_metadata.chunks_count();
+                if stored_chunks < total_chunks {
```
snowmead, line 356: Should we check for an exact match of total chunks as well and return something like `InconsistentFile` (when the number of chunks returned from storage are greater then expected).

snowmead (approved):
LGTM, just a small concern.

client/rpc/src/lib.rs
links234: Good question given the highlighted code, however I'm not sure about codifying the client interface with possible bug scenarios. (Why should I as a client care about you not having a proper file in storage? - just tell me you don't have it and log it for the server operator to see it).

links234 (approved):
LGTM! Thank you!

client/rpc/src/lib.rs
```diff
@@ -71,6 +71,13 @@ pub enum SaveFileToDisk {
     IncompleteFile(IncompleteFileStatus),
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub enum GetFileFromFileStorageResult {
```
ffarall, line 75: Given that this struct is the exact same as the one above, how about changing the name and reusing it?

ffarall (approved):

client/rpc/src/lib.rs
snowmead: Well as a user, I want to know if my data stored is consistent with what I expect it to be. If it isn't then, I would re-upload the data to ensure it is. We are not just saying if the file exists or not, we are also saying if we are missing data. I think we should also cover the opposite **unlikely** event we have more data then what is expected.

TDemeco: Changed it to account for this!

client/rpc/src/lib.rs
TDemeco: With the latest changes it's more differentiated so I think it makes sense to keep both

PR #199, opened 2024-09-19
feat: ✨ user without funds task

TDemeco:
This PR:
- Adds the `ProcessStopStoringForInsolventUserRequest` event emitted by the BlockchainService, which is then handled by the ChargeFees task to execute the process of deleting a file from an insolvent user.
  - This event is queued up when receiving a `UserWithoutFunds` or `SpStopStoringInsolventUser` event from the runtime, which is the queue that the BSP is able to queue up a file deletion transaction.
  - The queue lives in permanent storage, so in the event of a client reboot the Provider can resume deleting all the files for insolvent users.
  - Since each file deletion modifies the forest root of the Provider, this event is only processed when there's neither a submit proof nor a confirm storing event pending (since those take priority, although that could be easily modified).
- Adds handlers for the `UserWithoutFunds` and `SpStopStoringInsolventUser` events from the runtime, which queue up a `ProcessStopStoringForInsolventUserRequest` event.
- Adds a `get_files_by_user` method to the ForestStorage trait, which should return a vector of tuples that hold file keys and metadatas, for each file stored in the Forest storage that belongs to the specified user.
- Makes it so when a BSP confirms that it's storing a file in the FileSystem pallet, the payment stream between that Provider and the owner of the file is created or updated (if it already existed) with the new provided amount.
- Fixes a bug where the `user_deposit` field of the payment streams in the PaymentStreams pallet was not getting updated correctly.


links234 (approved):
Focused mostly on the client side. I left some comments and a small possible refactor (fine to leave it for now, required if and when we'll have more forest root changing flows). Besides that, great job! :D

client/blockchain-service/src/state.rs
```diff
@@ -30,6 +30,15 @@ impl SingleScaleEncodedValueCf for OngoingProcessConfirmStoringRequestCf {
     const SINGLE_SCALE_ENCODED_VALUE_NAME: &'static str = "ongoing_process_confirm_storing_request";
 }
 
+/// Current ongoing task which requires a forest write lock.
```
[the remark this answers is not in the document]
TDemeco: Hey! Totally agree. Just talked to Santi and we are going to leave a task for this on Notion to tackle in the future instead of doing it in this PR

node/src/tasks/bsp_charge_fees.rs
```diff
@@ -103,3 +197,193 @@ where
...
+        // Get all the files that belong to the insolvent user, delete the first one.
+        let user_files = fs
+            .read()
+            .await
+            .get_files_by_user(&insolvent_user)
+            .map_err(|e| anyhow!("Failed to get metadata from Forest: {:?}", e))?;
+
+        if !user_files.is_empty() {
+            let (file_key, metadata) = user_files.first().expect("User files is not empty");
+            let bucket_id = sp_core::H256::from_slice(metadata.bucket_id.as_ref());
```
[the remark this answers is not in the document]
TDemeco: Updated!

node/src/tasks/bsp_charge_fees.rs
```diff
@@ -50,6 +82,68 @@ where
...
+            .get(&NoKey)
+            .await
+            .ok_or_else(|| anyhow!("Failed to get forest storage."))?;
+
+        let root = { fs.read().await.root() };
+
+        trace!(target: LOG_TARGET, "Provider root according to Forest Storage: {:?}", root);
+
+        if root != onchain_root {
+            error!(target: LOG_TARGET, "CRITICAL❗️❗️ Applying mutations yielded different root than the one on-chain. This means that there is a mismatch between the Forest root on-chain and in this node. \nThis is a critical bug. Please report it to the StorageHub team.");
```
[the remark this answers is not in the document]
TDemeco: We return an error that then gets propagated up through the result from `handle_event`, I believe that should be sufficient, no? @ffarall does the same thing in the `submit_proof` tasks so if we modify this one we might have to modify both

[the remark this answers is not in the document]
links234: The difference is that if you miss a submit proof you can still catch up, if you have a root mismatch where/how do you sort it out?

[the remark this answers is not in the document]
TDemeco: Yeah but it's not for missing to submit a proof, it's actually exactly the same code: after submitting a valid proof, it checks that the forest root matches the one on-chain, and errors out if it doesn't. It's not something you can recover from though, and it should never happen, that's why the error states to contact the StorageHub team since it's an inconsistency error

node/src/tasks/bsp_charge_fees.rs
```diff
@@ -51,6 +83,68 @@ where
             storage_hub_handler,
         }
     }
+
+    async fn remove_file_from_forest(&self, file_key: &H256) -> anyhow::Result<()> {
```
ffarall, line 87: I'd rather we moved these implementations to the bottom of the file. That way at the top you see what events this task reacts to, and how it handles them.

ffarall, line 87: I'd move these implementations to the bottom of the file. That way it is easier to check what events this file reacts to, and how it handles them.

node/src/tasks/bsp_charge_fees.rs
```diff
@@ -104,3 +198,197 @@ where
         Ok(())
     }
 }
+
+impl<FL, FSH> EventHandler<UserWithoutFunds> for BspChargeFeesTask<FL, FSH>
+where
+    FL: FileStorageT,
+    FSH: BspForestStorageHandlerT,
+{
+    async fn handle_event(&mut self, event: UserWithoutFunds) -> anyhow::Result<()> {
```
ffarall, line 207: This task is not filtering for a user that this BSP is serving. That means that this will queue up `StopStoringForInsolventUserRequest`s for every single user without funds, regardless of whether it can charge it or not.

node/src/tasks/bsp_charge_fees.rs
```diff
@@ -104,3 +198,197 @@ where
...
+        Ok(())
+    }
+}
+
+impl<FL, FSH> EventHandler<SpStopStoringInsolventUser> for BspChargeFeesTask<FL, FSH>
+where
+    FL: FileStorageT,
+    FSH: BspForestStorageHandlerT,
+{
+    async fn handle_event(&mut self, event: SpStopStoringInsolventUser) -> anyhow::Result<()> {
```
ffarall, line 234: This task is not filtering for a user that this BSP is serving. That means that this will queue up StopStoringForInsolventUserRequests for every single user without funds, regardless of whether it can charge it or not.

node/src/tasks/bsp_charge_fees.rs
```diff
@@ -104,3 +198,197 @@ where
...
+                        info!(target: LOG_TARGET, "Submitted extrinsic to charge users with debt: {}", submitted_transaction.hash());
+                    }
+                    Err(e) => {
+                        error!(target: LOG_TARGET, "Failed to send extrinsic to charge users with debt: {}", e);
+                    }
+                }
+            }
+        }
+
+        // Release the forest root write "lock".
```
ffarall, line 389: I'd at least log if there is an error here. It would be critical.

node/src/tasks/bsp_charge_fees.rs
```diff
@@ -104,3 +198,197 @@ where
...
+
+        // Get all the files that belong to the insolvent user, delete the first one.
+        let user_files = fs
+            .read()
+            .await
+            .get_files_by_user(&insolvent_user)
+            .map_err(|e| anyhow!("Failed to get metadata from Forest: {:?}", e))?;
+
+        if !user_files.is_empty() {
+            let (file_key, metadata) = user_files.first().expect("User files is not empty");
```
ffarall, line 305: Are you just deleting the first file?

client/forest-manager/src/in_memory.rs
```diff
@@ -124,6 +124,27 @@ where
...
+        let trie = TrieDBBuilder::<T>::new(&self.memdb, &self.root).build();
+        let mut files = Vec::new();
+        let mut trie_iter = trie
+            .iter()
+            .map_err(|_| ForestStorageError::FailedToCreateTrieIterator)?;
+
+        while let Some((_, value)) = trie_iter.next().transpose()? {
+            let metadata = FileMetadata::decode(&mut &value[..])?;
+            let file_key = metadata.file_key::<T::Hash>();
+            if metadata.owner == user.encode() {
```
ffarall, line 141: Why not encoding the user once, before the `while` loop. So that you don't execute it every time.

test/suites/integration/bsp/debt-collection.test.ts
```diff
@@ -181,27 +152,533 @@ describeBspNet(
...
+      ///////** INITIAL SETUP **///////
+
+      // Make sure the payment stream between Alice and the DUMMY_BSP_ID actually exists
+      const paymentStreamExistsResult =
+        await userApi.call.paymentStreamsApi.getUsersOfPaymentStreamsOfProvider(
+          ShConsts.DUMMY_BSP_ID
+        );
+      // Check if the first element of the returned vector is the user
+      assert(paymentStreamExistsResult[0].toString() === userAddress);
+      assert(paymentStreamExistsResult.length === 1);
```
ffarall, line 190: If you have to do this as initial steps of every test, you can use the `beforeEach` hook.

test/suites/integration/bsp/debt-collection.test.ts
```diff
@@ -181,27 +152,533 @@ describeBspNet(
...
+        ShConsts.DUMMY_BSP_ID
+      );
+
+      // Seal one more block with the pending extrinsics.
+      await userApi.sealBlock();
+
+      // Get the current price of storage from the runtime, the new stream deposit and the ED
+      const currentPriceOfStorage = await userApi.query.paymentStreams.currentPricePerUnitPerTick();
+      const newStreamDeposit = await userApi.consts.paymentStreams.newStreamDeposit;
+      const existentialDeposit = await userApi.consts.balances.existentialDeposit;
```
ffarall, line 244: These `await`s are not needed

test/suites/integration/bsp/debt-collection.test.ts
```diff
@@ -181,27 +152,533 @@ describeBspNet(
...
+      const [userAccount, providerId, newAmountProvided] = fetchEventData(
+        userApi.events.paymentStreams.DynamicRatePaymentStreamUpdated,
+        await userApi.query.system.events()
+      );
+      // Assert that the information on-chain is correct
+      strictEqual(userAccount.toString(), userAddress);
+      strictEqual(providerId.toString(), ShConsts.DUMMY_BSP_ID.toString());
+      strictEqual(newAmountProvided.toNumber(), newAmountProvidedForInsolvency.toNumber());
+
+      ///////** INITIAL PROOF SUBMISSION **///////
```
ffarall, line 286: Suggestion: separate these parts into different `it` blocks.

ffarall (changes requested):
Amazing work ecompassing many areas of our codebase! Just a few things to iron out.

node/src/tasks/bsp_charge_fees.rs
TDemeco: Yep, this task deletes one file from the user each block until it has correctly deleted all its files, to avoid having multiple root-changing transactions in one block

node/src/tasks/bsp_charge_fees.rs
TDemeco: Yeah I saw this done in a few other tasks so I did the same but I was sceptical of not managing the result of the actual forest write. I'll update it in all tasks

test/suites/integration/bsp/debt-collection.test.ts
TDemeco: I don't HAVE to do it as initial steps since checking once in the first test is enough, but I did it here just to be safe. I'm reluctant to add it as a beforeEach hook just because a future test might need for the payment stream to NOT exist

test/suites/integration/bsp/debt-collection.test.ts
TDemeco: Lol 😝

node/src/tasks/bsp_charge_fees.rs
TDemeco: Changed!

TDemeco: Changed!

client/forest-manager/src/in_memory.rs
TDemeco: Yep, good catch. Changed

node/src/tasks/bsp_charge_fees.rs
TDemeco: Yes, how it works right now is that the handler of the actual deletion process checks the forest storage to see if there's any file belonging to that user, but I see that it can be more efficient to check before since we don't need to wait to acquire a write lock on the forest root (since we are only reading) so I modified both runtime event handlers to do exactly that

node/src/tasks/bsp_charge_fees.rs
TDemeco: Same as above^

pallets/file-system/Cargo.toml
```diff
@@ -33,6 +33,7 @@ sp-trie = { workspace = true }
 
 # Local
 pallet-file-system-runtime-api = { workspace = true }
+pallet-payment-streams = { workspace = true }
```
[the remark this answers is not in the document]
TDemeco: Nope, not sure why I put it there in the first place

pallets/file-system/src/utils.rs
```diff
@@ -957,6 +957,27 @@ where
                 storage_request_metadata.size,
             )?;
 
+            // Check if a payment stream between the user and provider already exists.
+            // If it does not, create it. If it does, update it.
+            match <T::PaymentStreams as PaymentStreamsInterface>::get_dynamic_rate_payment_stream_amount_provided(&bsp_id, &storage_request_metadata.owner) {
```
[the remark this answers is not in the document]
TDemeco: Just added rust tests for this

pallets/file-system/src/utils.rs
```diff
@@ -1354,9 +1375,13 @@ where
...
 
-        // Check that the owner of the file has been flagged as insolvent
+        // Check that the owner of the file has been flagged as insolvent OR that the Provider does not
+        // have any active payment streams with the user.
         ensure!(
-            <T::UserSolvency as ReadUserSolvencyInterface>::is_user_insolvent(&owner),
+            <T::UserSolvency as ReadUserSolvencyInterface>::is_user_insolvent(&owner)
+                || !<T::PaymentStreams as PaymentStreamsInterface>::has_active_payment_stream(
+                    &sp_id, &owner
+                ),
```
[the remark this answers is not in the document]
TDemeco: Added

pallets/payment-streams/src/utils.rs
```diff
@@ -208,6 +208,7 @@ where
                 Error::<T>::PaymentStreamNotFound
             );
             payment_stream.rate = new_rate;
+            payment_stream.user_deposit = new_deposit;
```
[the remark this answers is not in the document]
TDemeco: Added some checks for this in existing tests!

pallets/payment-streams/src/utils.rs
```diff
@@ -1375,35 +1394,39 @@ where
...
-                    .price_index
-                    .saturating_sub(dynamic_stream.price_index_when_last_charged);
-                let amount_to_charge = price_index_difference
-                    .checked_mul(&dynamic_stream.amount_provided.into())
-                    .ok_or(GetUsersWithDebtOverThresholdError::AmountToChargeOverflow)?;
-                debt = debt
-                    .checked_add(&amount_to_charge)
-                    .ok_or(GetUsersWithDebtOverThresholdError::DebtOverflow)?;
-            }
+            if !UsersWithoutFunds::<T>::contains_key(&user) {
```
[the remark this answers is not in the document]
TDemeco: Added

test/suites/integration/bsp/debt-collection.test.ts
TDemeco: Done

node/src/tasks/bsp_charge_fees.rs
ffarall: Exactly, that was my point, avoiding having to acquire the write lock.

node/src/tasks/bsp_charge_fees.rs
ffarall: I see. So it would be ideal that we make a batch deletion transaction, right?

ffarall (approved):

node/src/tasks/bsp_charge_fees.rs
TDemeco: 100%, the same way we have a batch for BSPs confirming to store a file we should have batching for deletion as well.

node/src/tasks/bsp_charge_fees.rs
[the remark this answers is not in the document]
ffarall: I think the error log is the right way to go. We're signalling strongly that this is a critical bug, but we're not panicking. Even if this is unrecoverable, the worst that can happen is that from here on the forest-related transactions will fail, until someone manually does something (most likely). But, by keeping the node alive, you can maybe query it or investigate further why this happened.

PR #200, opened 2024-09-19
feat: MSP batch respond storage requests

snowmead:
MSP batch respond to storage requests with best-effort strategy. 

MSPs can accept and reject storage requests in batches (multiple buckets - multiple file keys). The corresponding batch event is emitted to signal the network of users and SPs which keys have been accepted, rejected or failed to process.

Best-effort strategy, meaning all file keys that were failed to be processed does not halt the extrinsic's execution and instead records all failed file keys to notify the MSP.

MSP event handler implementations for `MspUploadFileTask`:

- `NewStorageRequest`
- `RemoteUploadRequest`
- `ProcessMspRespondStoringRequest`

User event handler skeleton for `UserSendsFileTask`:

- `NewStorageRequest`: Will be implemented in another PR.

Added various runtime APIs.

Updates to polkadotJs fork will be merged after. https://github.com/Moonsong-Labs/pjs-ui/pull/43


docker/local-dev-full-rocksdb-compose.yml
```diff
@@ -0,0 +1,89 @@
+services:
```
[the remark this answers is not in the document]
timbrinded: e.g.

- bsp: 9666
- msp: 9777
- user: 9888

ffarall (commented):
Partial review. So far amazing job with a complicated functionality.

node/src/tasks/msp_upload_file.rs
```diff
@@ -0,0 +1,854 @@
...
+    async fn handle_event(&mut self, event: NewStorageRequest) -> anyhow::Result<()> {
+        info!(
+            target: LOG_TARGET,
+            "Registering user peer for file_key {:?}, location {:?}, fingerprint {:?}",
+            event.file_key,
+            event.location,
+            event.fingerprint
+        );
+
+        let result = self.handle_new_storage_request_event(event).await;
```
TDemeco, line 99: Why did you externalize this logic into a function if it's only used once?

node/src/tasks/msp_upload_file.rs
```diff
@@ -0,0 +1,854 @@
...
+                .wait_for_block(earliest_change_capacity_block)
+                .await?;
+
+            self.storage_hub_handler
+                .blockchain
+                .send_extrinsic(call, Tip::from(0))
+                .await?
+                .with_timeout(Duration::from_secs(60))
+                .watch_for_success(&self.storage_hub_handler.blockchain)
+                .await?;
```
TDemeco, line 680: After taking a look at the code for `wait_for_block` I'm still not totally sure so might as well ask. This task gets frozen here until the condition is met, right?
Would it be more efficient in `wait_for_block` to (instead of just blocking the task until the required block number arrives) receive a callback and calling that callback when the required block number is received? That way we could free resources quicker

docker/local-dev-bsp-compose.yml
```diff
@@ -31,7 +31,7 @@ services:
     platform: linux/amd64
     container_name: docker-sh-user-1
     ports:
-      - "9977:9944"
+      - "9888:9944"
```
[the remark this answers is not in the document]
snowmead: Yup definitely, thx

node/src/tasks/msp_upload_file.rs
TDemeco: Also, if there're two of this task running, when the first one changes the capacity, the `earliest_change_capacity_block` changes but the second task never realizes that, so it'll try to send the extrinsic and it will fail. That could also be solved using the callback model. Again, this is only valid if `wait_for_block` works as I think it does 😬

TDemeco: IF it does and IF we decide to change this, I don't think I'd do it in this PR, just as a FYI. @ffarall wdyt?

pallets/file-system/src/types.rs
```diff
@@ -98,6 +99,113 @@ impl<T: Config> StorageRequestMetadata<T> {
...
+pub enum RejectedStorageRequestReason {
+    ReachedMaximumCapacity,
+    ReceivedInvalidProof,
+    InternalError,
+}
+
+/// Input for MSPs to respond to storage request(s).
+///
+/// The input must be a list of (file_key, response) grouped by bucket id.
+pub type FileKeyResponsesInput<T> = BoundedVec<
```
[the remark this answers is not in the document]
snowmead: Ya wasn't sure how to name this. I am open to any other ideas.

node/src/tasks/msp_upload_file.rs
```diff
@@ -0,0 +1,854 @@
...
+            .map_err(|e| anyhow!("Failed to insert file in file storage: {:?}", e))?;
+        drop(write_file_storage);
+
+        Ok(())
+    }
+
+    /// Calculate the new capacity after adding the required capacity for the file.
+    ///
+    /// The new storage capacity will be increased by the jump capacity until it reaches the
+    /// `max_storage_capacity` or it
```
TDemeco, line 785: Don't leave me hanging like this! 😆

pallets/file-system/src/utils.rs
```diff
@@ -574,96 +621,381 @@ where
...
+                    .try_push((
+                        file_key.clone(),
+                        Error::<T>::InvalidBucketIdFileKeyPair.into(),
+                    ))
+                    .map_err(|_| Error::<T>::TooManyStorageRequestResponses)?;
+                continue;
+            }
+
+            // Check that the MSP is the one storing the bucket.
+            if !<T::Providers as ReadBucketsInterface>::is_bucket_stored_by_msp(
```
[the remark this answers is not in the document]
snowmead: No, it is ensuring that the file key taken from the batch of accepted file keys is part of the current bucket being processed. If it isn't, we skip it.
We process batches of file keys each belonging to a specific bucket.

node/src/tasks/msp_upload_file.rs
```diff
@@ -0,0 +1,854 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+/// Handles the `ProcessConfirmStoringRequest` event.
+///
+/// This event is triggered by the runtime when it decides it is the right time to submit a confirm
+/// storing extrinsic (and update the local forest root).
```
TDemeco, line 372: These docs should be updated

node/src/tasks/msp_upload_file.rs
```diff
@@ -0,0 +1,854 @@
...
+where
+    FL: FileStorageT,
+    FSH: MspForestStorageHandlerT,
+{
+    async fn handle_event(&mut self, event: ProcessMspRespondStoringRequest) -> anyhow::Result<()> {
+        info!(
+            target: LOG_TARGET,
+            "Processing ConfirmStoringRequest: {:?}",
+            event.data.respond_storing_requests,
+        );
```
TDemeco, line 383: This log too

pallets/file-system/src/utils.rs
```diff
@@ -578,96 +625,381 @@ where
...
+
+            let owner = <T::Providers as ReadBucketsInterface>::get_bucket_owner(&bucket_id)
+                .map_err(|_| Error::<T>::BucketNotFound)?;
+
+            if let Some(accepted_file_keys) = file_key_responses.accept {
+                // Call do_msp_accept_storage_request, which returns the new_bucket_root
+                let (new_bucket_root, accepted_file_keys, failed_file_keys) =
+                    Self::do_msp_accept_storage_request(
+                        msp_id.clone(),
+                        bucket_id.clone(),
```
TDemeco, line 674: Clippy warns about these two since they implement `Copy`, so no need to clone them :)

pallets/file-system/src/utils.rs
```diff
@@ -578,96 +625,381 @@ where
...
-            <T::Providers as ReadStorageProvidersInterface>::available_capacity(&sp_id)
-                >= storage_request_metadata.size,
-            Error::<T>::InsufficientAvailableCapacity
-        );
+                // Create batch responses
+                if !accepted_file_keys.is_empty() {
+                    let accepted_batch = MspAcceptedBatchStorageRequests {
+                        file_keys: accepted_file_keys,
+                        bucket_id: bucket_id.clone(),
+                        new_bucket_root: new_bucket_root.clone(),
```
TDemeco, line 683: Same here^

pallets/file-system/src/utils.rs
```diff
@@ -578,96 +625,381 @@ where
...
+
+                    rejected
+                        .try_push((file_key.0, file_key.1.clone()))
+                        .map_err(|_| Error::<T>::TooManyStorageRequestResponses)?;
+                }
+
+                if !rejected.is_empty() {
+                    let rejected_batch = MspRejectedBatchStorageRequests {
+                        file_keys: rejected,
+                        bucket_id: bucket_id.clone(),
```
TDemeco, line 737: And here^

pallets/file-system/src/utils.rs
```diff
@@ -578,96 +625,381 @@ where
...
+                    batch_responses
+                        .try_push(BatchResponses::Rejected(rejected_batch))
+                        .map_err(|_| Error::<T>::TooManyBatchResponses)?;
+                }
+            }
+
+            if !failed.is_empty() {
+                let failed_batch = MspFailedBatchStorageRequests {
+                    file_keys: failed,
+                    bucket_id: bucket_id.clone(),
```
TDemeco, line 750: Same here^

TDemeco: I think both Provider/Bucket IDs and Hashes (file keys, roots, etc) implement copy, so no need to clone them

TDemeco (commented):
Amazing work!!

pallets/file-system/src/utils.rs
```diff
@@ -574,96 +621,381 @@ where
...
+            }
+
+            // Get the file metadata to insert into the bucket under the file key.
+            let file_metadata = storage_request_metadata.clone().to_file_metadata();
+            let encoded_trie_value = file_metadata.encode();
+
+            // Skip any duplicates.
+            if accepted_file_keys_and_metadata.contains_key(&file_key) {
+                continue;
+            }
```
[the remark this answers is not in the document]
snowmead: good point.

pallets/file-system/src/utils.rs
```diff
@@ -574,96 +621,381 @@ where
...
+                <T::Providers as MutateStorageProvidersInterface>::increase_capacity_used(
+                    &msp_id,
+                    storage_request_metadata.size,
+                )
+            {
+                failed_file_keys
+                    .try_push((file_key.clone(), e))
+                    .map_err(|_| Error::<T>::TooManyStorageRequestResponses)?;
+                continue;
+            }
```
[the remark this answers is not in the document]
snowmead: Ya just not sure if I should expect_or_error? or return an error early. I think this is just safer.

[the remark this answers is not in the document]
snowmead: Ya, added expect_or_err and added a comment.

pallets/file-system/src/utils.rs
```diff
@@ -574,96 +621,381 @@ where
...
+                // Update storage request metadata.
+                <StorageRequests<T>>::set(&file_key, Some(storage_request_metadata.clone()));
+            }
+
+            expect_or_err!(
+                accepted_file_keys_and_metadata
+                    .insert(file_key, encoded_trie_value)
+                    .is_none(),
+                "Failed to insert file key and metadata into accepted_file_keys_and_metadata",
+                Error::<T>::InconsistentState,
```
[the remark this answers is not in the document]
snowmead: Renamed to `InconsistentStateKeyAlreadyExists` and updated doc.

pallets/file-system/src/utils.rs
```diff
@@ -1100,12 +1389,30 @@ where
...
+
+        Ok(())
+    }
+
+    /// When a storage request is revoked and has already been confirmed by some BSPs, a challenge (with priority) is
+    /// issued to force the BSPs to update their storage root to uninclude the file from their storage.
+    ///
+    /// All BSPs that have volunteered to store the file are removed from the storage request and the storage request is deleted.
+    fn cleanup_storage_request(
+        revoker: EitherAccountIdOrProviderId<T>,
```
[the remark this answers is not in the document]
snowmead: That's a good point.

node/src/tasks/msp_upload_file.rs
snowmead: I copied the same pattern used in the BSP task just to keep it consistent.

node/src/tasks/msp_upload_file.rs
snowmead: `wait_for_block` does exactly that. But yes the task will wait until it is reached and it is possible to have mutliple tasks waiting waiting to add capacity. In that case we should probably check after waiting for so long if the capacity has already changed and if it covers what we needed.

node/src/tasks/msp_upload_file.rs
snowmead: Remnant, sorry. xD

pallets/file-system/src/utils.rs
```diff
@@ -186,6 +195,44 @@ where
...
+                        .try_into()
+                        .map_err(|_| QueryMspConfirmChunksToProveForFileError::InternalError)?,
+                );
+
+                Ok(challenged_chunk)
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
+        Ok(chunks_to_prove)
+    }
```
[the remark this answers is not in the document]
snowmead: agreed

docker/local-dev-full-rocksdb-compose.yml
```diff
@@ -0,0 +1,89 @@
...
+        "--unsafe-rpc-external",
+        "--rpc-methods=unsafe",
+        "--port=30444",
+        "--rpc-cors=all",
+        "--node-key=0x13b3b1c917dda506f152816aad4685eefa54fe57792165b31141ac893610b314",
+        "--bootnodes=/ip4/${BSP_IP:-default_bsp_ip}/tcp/30350/p2p/${BSP_PEER_ID:-default_bsp_peer_id}",
+        "--keystore-path=/keystore",
+        "--sealing=manual",
+        "--base-path=/data",
+      ]
```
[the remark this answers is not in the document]
snowmead: good point!

node/src/tasks/msp_upload_file.rs
ffarall: Conclusion: We actually need a unified piece of code in the blockchain service that "accumulates" all the needed storage capacity changes in the various tasks, and when we reach the block where the capacity change is allowed, it makes the capacity change for all of them considered. This would be left for another PR.

pallets/file-system/src/utils.rs
```diff
@@ -574,96 +621,381 @@ where
         Ok(file_key)
     }
 
-    pub(crate) fn do_msp_accept_storage_request(
+    pub(crate) fn do_msp_respond_storage_request(
```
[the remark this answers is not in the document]
snowmead: After some discussion, it was concluded that we would have another extrinsic to accept only a single batch of file keys for a single bucket. This would give MSPs the opportunity to call the single bucket extrinsic if there are many file keys to respond to all belonging to one bucket. This is because benchmarking the multi-bucket extrinsic would require us to reduce the number of file keys you can submit for a single bucket.

ffarall (approved):

PR #202, opened 2024-09-23
Feat: indexer service structure

links234:
This adds the structure for the Storage Hub Indexer, specifically:
1. Add `indexer-db` crate which contains the migrations, model definitions (currently just the service state) and helpers for Postgres database:
 - Uses `diesel` ORM both for easy typed models and `diesel-async` for async usage.
 - A database pool is used not because it is necessarily more performant but because it automatically handles reconnects.
3. Adds a new background service (`indexer-service` crate) which listens to finality events and maintains connection to the DB.
4. Update CLI to start the indexer independently of storage provider roles.
5. Support docker compose file to start a Postgres instance for development.


client/indexer-db/README.md
```diff
@@ -0,0 +1,81 @@
+# Storage Hub Indexer Database
```
ffarall, line 1: These docs are much appreciated.

client/indexer-service/README.md
```diff
@@ -0,0 +1,7 @@
+# Storage Hub Indexer Service
```
ffarall, line 1: As we add indexed fields/events/whatever, it will come very handy to have some docs here on what are the steps required to add "something else" to what is indexed.

client/indexer-service/src/handler.rs
```diff
@@ -0,0 +1,183 @@
...
+    db_pool: DbPool,
+}
+
+// Implement the Actor trait for IndexerService
+impl Actor for IndexerService {
+    type Message = IndexerServiceCommand;
+    type EventLoop = IndexerServiceEventLoop;
+    type EventBusProvider = (); // We're not using an event bus for now
+
+    fn handle_message(
```
ffarall, line 34: Any reason why you didn't follow the `blockchain-service/src/handler.rs` structre here? Meaning, having the `ActorEventLoop` implementation at the top, with the `run()` function, then followed by the implementations of `handle_message` and `handle_finality_notification`.

client/indexer-service/README.md
ffarall: Also, it would be very helpful to have a "Run node as indexer" docs here.

ffarall (approved):
Good job! Approved with some side comments.

client/indexer-service/src/handler.rs
links234: No particular reason, I can rearrange them the be consistent. I didn't gave much thought about it since it's lacking an `EventBus`, there are no commands/messages yet (probably won't be) and hence not identical as the blockchain service in this regard.

client/indexer-service/Cargo.toml
```diff
@@ -0,0 +1,69 @@
...
+repository = { workspace = true }
+edition = { workspace = true }
+publish = false
+
+[lib]
+
+[lints]
+workspace = true
+
+[dependencies]
```
snowmead, line 17: It seems like there are a lot of unused dependencies. If that's the case, can we remove them?

snowmead (approved):
Nice job

PR #203, opened 2024-09-23
test: 🧪 Placeholder for `missing chunks` test scenario

timbrinded:
This adds a test which for now, doesn't do much, but will be useful when we deliver the implementation to handle when a BSP doesn't get all the file chunks for a storage request. This test uses existing toxiproxy impl to register a toxic which will cut off data transfer after 50kb.

This PR also contains new api method:
- api.assert.log() : this will throw if no matches found for an input string, or if successful return the log line


ffarall (approved):
Good job!

PR #204, opened 2024-09-24
bug: Threshold slope increases based on bsp weight

snowmead:
Threshold slope should increase based on the BSP's weight.

### DIFF
--- a/pallets/file-system/src/utils.rs
+++ b/pallets/file-system/src/utils.rs
@@ -1768,13 +1768,20 @@ where
         let threshold_weighted_starting_point =
             bsp_weight.saturating_mul(threshold_global_starting_point);
 
-        // Rate of increase from the weighted threshold starting point up to the maximum threshold within a block range.
-        let threshold_slope = maximum_threshold
-            .saturating_sub(threshold_weighted_starting_point)
+        let base_slope = maximum_threshold
+            .saturating_sub(threshold_global_starting_point)
             .checked_div(&T::ThresholdTypeToBlockNumber::convert_back(
                 BlockRangeToMaximumThreshold::<T>::get(),
             ))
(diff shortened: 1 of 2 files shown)


pallets/file-system/src/utils.rs
```diff
@@ -1768,14 +1768,21 @@ where
...
+            .saturating_sub(threshold_global_starting_point)
             .checked_div(&T::ThresholdTypeToBlockNumber::convert_back(
                 BlockRangeToMaximumThreshold::<T>::get(),
             ))
             .unwrap_or(T::ThresholdType::one());
 
+        let threshold_slope = base_slope.saturating_add(
+            base_slope
+                .checked_mul(&bsp_weight)
+                .unwrap_or(T::ThresholdType::zero())
```
ffarall, line 1781: If the `checked_mul` fails here, wouldn't it be because of an overflow? In that case, it wouldn't make much sense to set it to 0. Wouldn't it be more appropriate to set it to the maximum threshold?

pallets/file-system/src/utils.rs
```diff
@@ -1768,14 +1768,21 @@ where
...
+                .checked_div(&global_weight)
+                .unwrap_or(T::ThresholdType::zero()),
```
ffarall, line 1783: Wouldn't it make more sense to set it to 1 here?

ffarall (changes requested):

ffarall:
Also, consider adding a quick test to catch this case that we missed.

pallets/file-system/src/utils.rs
snowmead: Good point.

ffarall (approved):

PR #206, opened 2024-09-26
refactor: :art: add `FileMetadataInterface` to handle file metadatas in the runtime

TDemeco:
This refactor is done to allow runtime pallets to interact in a generic way with file metadata, instead of being hardcoded to use the client's type.
Currently the only requirements that we ask from the file metadata is that it has the file size and the file owner, but more could be added in the future


ffarall (approved):
Great job!

PR #208, opened 2024-09-27
feat: ✨ Add pallet params and use it for custom test params

ffarall:
This PR:
1. Adds `pallet-parameters` to the runtime configuration (non-mocks).
2. Uses `pallet-parameters` for `SlashAmountPerMaxFileSize` and `StakeToChallengePeriod` constants, making them now updatable parameters through sudo calls.
3. Adds the `get_worst_case_scenario_slashable_amount` runtime API to access the `SlashAmountPerMaxFileSize` value from outside of a wasm context, now that it's not just a constant.
4. Updates `SlashAmountPerMaxFileSize` and `StakeToChallengePeriod` parameters to use more real world cases in runtime config.
5. Updates `runSimpleBspNet` test initialisation to set `SlashAmountPerMaxFileSize` and `StakeToChallengePeriod` to convenient values for testing.
6. Fixes bug where `stake_to_challenge_period` was not saturating properly to `MinChallengePeriodFor`.


snowmead:
I want to raise a couple of important points here taken from the [pallet-parameters docs](https://paritytech.github.io/polkadot-sdk/master/pallet_parameters/index.html)

1. "_**THIS CRATE IS NOT AUDITED AND SHOULD NOT BE USED IN PRODUCTION.**_" which to me maybe isn't such a big deal since we are not necessarily launching StorageHub immediately after implementation and could have some time before the pallet is audited.
2. "_Note that this is incurring one storage read per access. This should not be a problem in most cases but must be considered in weight-restrained code._" important to keep in mind if ever we use any parameters that are part of some hook implementation.

@ffarall I think we should still go ahead with this, but it was just to have this written down somewhere.

ffarall:
@snowmead thanks for bringing these up.

> 1. "_**THIS CRATE IS NOT AUDITED AND SHOULD NOT BE USED IN PRODUCTION.**_" which to me maybe isn't such a big deal since we are not necessarily launching StorageHub immediately after implementation and could have some time before the pallet is audited.

Indeed we were aware that pallet parameters is not supposedly production ready. The assumption, as you said, is that there will be enough time before we have to go to production. Additionally, Moonbeam is already using it in Moonbase, so they'll probably test it before we do as well.

> 2. "_Note that this is incurring one storage read per access. This should not be a problem in most cases but must be considered in weight-restrained code._" important to keep in mind if ever we use any parameters that are part of some hook implementation.

Yes, this is an important aspect to keep in mind.

snowmead (approved):

PR #209, opened 2024-09-27
Feat: index msp, bsp, bucket

links234:


client/indexer-db/src/models/bsp.rs
```diff
@@ -0,0 +1,92 @@
...
+
+use crate::{
+    models::multiaddress::MultiAddress, schema::bsp, schema::bsp_multiaddress, DbConnection,
+};
+
+/// Table that holds the BSPs.
+/// The account is guaranteed to be unique across both MSPs and BSPs.
+#[derive(Debug, Queryable, Insertable, Selectable)]
+#[diesel(table_name = bsp)]
+pub struct Bsp {
```
ffarall, line 14: What do you think about storing the stake of the BSP here? I can't think of an immediate use case, but I think it would also be handy.

ffarall: Also, any reason why you're not storing the forest root of the BSP here?

client/indexer-db/src/models/bucket.rs
```diff
@@ -0,0 +1,81 @@
...
+use diesel::prelude::*;
+use diesel_async::RunQueryDsl;
+
+use crate::{schema::bucket, DbConnection};
+
+/// Table that holds the Buckets.
+#[derive(Debug, Queryable, Insertable, Selectable)]
+#[diesel(table_name = bucket)]
+#[diesel(belongs_to(Msp, foreign_key = msp_id))]
+pub struct Bucket {
```
ffarall, line 11: Here as well, why not store the bucket's forest root?

client/indexer-db/src/models/bucket.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub account: String,
+    pub blockchain_id: String,
+    pub name: Vec<u8>,
+    pub collection_id: Option<String>,
+    pub private: bool,
+    pub created_at: NaiveDateTime,
+    pub updated_at: NaiveDateTime,
+}
+
+impl Bucket {
```
ffarall, line 23: Shouldn't this also have a `delete` function? It can be added later anyway.

client/indexer-db/src/models/bucket.rs
ffarall: I imagine that it is not handled yet, but there should be a way of storing files with their metadata, and link those files to this bucket. Also link the BSPs storing that file to the file.

client/indexer-db/src/models/msp.rs
```diff
@@ -0,0 +1,94 @@
...
+        account: String,
+    ) -> Result<(), diesel::result::Error> {
+        diesel::delete(msp::table)
+            .filter(msp::account.eq(account))
+            .execute(conn)
+            .await?;
+        Ok(())
+    }
+
+    pub async fn get_by_blockchain_id<'a>(
```
ffarall, line 84: What is this for?

client/indexer-db/src/models/bsp.rs
```diff
@@ -0,0 +1,92 @@
...
+    pub id: i32,
+    pub account: String,
+    pub capacity: BigDecimal,
+    pub created_at: NaiveDateTime,
+    pub updated_at: NaiveDateTime,
+    pub blockchain_id: String,
```
ffarall, line 20: What is this field for?

client/indexer-db/src/models/msp.rs
```diff
@@ -0,0 +1,94 @@
...
+#[derive(Debug, Queryable, Insertable, Selectable)]
+#[diesel(table_name = msp)]
+pub struct Msp {
+    pub id: i32,
+    pub account: String,
+    pub capacity: BigDecimal,
+    pub value_prop: String,
+    pub created_at: NaiveDateTime,
+    pub updated_at: NaiveDateTime,
+    pub blockchain_id: String,
```
ffarall, line 22: What is this for?

client/indexer-service/src/handler.rs
```diff
@@ -107,6 +115,269 @@ impl IndexerService {
...
+                self.index_payment_streams_event(conn, event).await?
+            }
+            RuntimeEvent::ProofsDealer(event) => {
+                self.index_proofs_dealer_event(conn, event).await?
+            }
+            RuntimeEvent::Providers(event) => self.index_providers_event(conn, event).await?,
+            RuntimeEvent::Randomness(event) => self.index_randomness_event(conn, event).await?,
+            // Runtime events that we're not interested in.
+            // We add them here instead of directly matching (_ => {})
+            // to ensure the compiler will let us know to treat future events when added.
```
ffarall, line 137: Good clarification

ffarall (approved):
Great job! Just some minor comments

client/indexer-db/src/models/msp.rs
links234: Probably this needs to be clarified in a comment. The `id`s from SQL models (and foreign keys `msp_id`/`bsp_id`) are internal, integers for faster indexing by Postgres. It is not the same as the msp_id/bsp_id from the runtime. This is what `blockchain_id` is (`Bsp.blockchain_id == bsp_id from runtime`).

client/indexer-db/src/models/bsp.rs
links234: I will track once I process the events which emit (or should emit) the forest roots.

client/indexer-db/src/models/bsp.rs
```diff
@@ -0,0 +1,92 @@
...
+use crate::{
+    models::multiaddress::MultiAddress, schema::bsp, schema::bsp_multiaddress, DbConnection,
+};
+
+/// Table that holds the BSPs.
+/// The account is guaranteed to be unique across both MSPs and BSPs.
+#[derive(Debug, Queryable, Insertable, Selectable)]
+#[diesel(table_name = bsp)]
+pub struct Bsp {
+    pub id: i32,
```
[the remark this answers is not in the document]
links234: Updated with a comment.

The `id` is Postgres internal id while `blockchain_id` (now `onchain_bsp_id`) is the hash from the runtime.

PR #211, opened 2024-09-30
test: ✅ Add integrity tests to Proofs Dealer pallet

ffarall:
This PR:
1. Implements `integrity_test` hook in the Proofs Dealer pallet.
    1. Checks that `CheckpointChallengePeriod` is greater or equal to the longest period a Provider can have.
    2. Checks that `BlockFullnessPeriod` is smaller or equal than `ChallengeTicksTolerance`.
2. Makes the corresponding changes in every mocked runtime to abide by this test.
3. Adds `get_min_stake` function to `ReadChallengeableProvidersInterface`.

This ensures that we never build a runtime where those constraints are not met, given the constants that we set. You can test it by setting the `CheckpointChallengePeriod` in `runtime/src/configs/mod.rs` to something sufficiently small, and run `cargo test`.

### DIFF
--- a/pallets/providers/src/mock.rs
+++ b/pallets/providers/src/mock.rs
@@ -245,7 +245,7 @@ impl crate::Config for Test {
     type ProvidersProofSubmitters = MockSubmittingProviders;
     type ReputationWeightType = u32;
     type Treasury = TreasuryAccount;
-    type SpMinDeposit = ConstU128<10>;
+    type SpMinDeposit = ConstU128<{ 10 * UNITS }>;
     type SpMinCapacity = ConstU64<2>;
     type DepositPerData = ConstU128<2>;
     type MaxFileSize = ConstU64<{ u64::MAX }>;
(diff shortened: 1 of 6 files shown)


pallets/providers/src/mock.rs
```diff
@@ -335,13 +335,15 @@ pub fn _new_test_ext() -> sp_io::TestExternalities {
...
-    pub const GEORGE: (u64, u128) = (6, 600_000_000_000);
+    use super::UNITS;
+
+    pub const ALICE: (u64, u128) = (0, 5_000_000 * UNITS);
+    pub const BOB: (u64, u128) = (1, 10_000_000 * UNITS);
+    pub const CHARLIE: (u64, u128) = (2, 20_000_000 * UNITS);
+    pub const DAVID: (u64, u128) = (3, 30_000_000 * UNITS);
+    pub const EVE: (u64, u128) = (4, 400_000_000 * UNITS);
+    pub const FERDIE: (u64, u128) = (5, 5_000_000_000 * UNITS);
+    pub const GEORGE: (u64, u128) = (6, 600_000_000_000 * UNITS);
```
TDemeco, line 346: Just an observation since it doesn't really change anything but I added these already considering the 12 decimals of DOT 😄 it's nicer to make that explicit tho

xcm-simulator/src/storagehub/configs/mod.rs
```diff
@@ -564,7 +564,7 @@ parameter_types! {
     pub const ChallengesQueueLength: u32 = 100;
     pub const CheckpointChallengePeriod: u32 = 10;
     pub const ChallengesFee: Balance = 1 * UNIT;
-    pub const StakeToChallengePeriod: Balance = 1_000_000 * UNIT;
+    pub const StakeToChallengePeriod: Balance = 200 * UNIT; // TODO: Change this value into something much higher like 1_000_000 * UNIT
```
TDemeco, line 567: Is there a reason for this change? Was it generating issues in tests?

TDemeco (approved):
Looks good!

snowmead (approved):
Nice!

xcm-simulator/src/storagehub/configs/mod.rs
ffarall: Yes, the maximum challenge period was too big, compared to the checkpoint challenges. For tests, periods of this magnitude are fine, but for the actual runtime configurations it should be numbers like `1_000_000 * UNIT`. This is handled with the introduction of pallet parameters

pallets/providers/src/mock.rs
ffarall: There was a test that was failing as a consequence of one of these users not having enough free balance, so i incremented it.

PR #212, opened 2024-09-30
feat: ⏫ upgrade to Polkadot SDK v1.11.0

TDemeco:
Update from Polkadot SDK 1.10.0 to 1.11.0

- Changelog for `Node Dev`:
    - [#4027](https://github.com/paritytech/polkadot-sdk/pull/4027) - Add `candidates_pending_availability` runtime API, which returns all the candidates for a parachain at a given relay block number (useful for elastic scaling)
    - [#3979](https://github.com/paritytech/polkadot-sdk/pull/3979) - Deprecate `para_id()` from CoreState in polkadot primitives in favor of direct usage of ClaimQueue (which provides a better view of what’s scheduled on each core). Again, useful to know for elastic scaling
    - [#4169](https://github.com/paritytech/polkadot-sdk/pull/4169) - transactionBroadcast RPC has been stabilized to version 1 and is no longer unstable. Useful RPC to broadcast and stop broadcasting tx to the chain. Not to be confused with tokyo’s broadcast
    - [#4171](https://github.com/paritytech/polkadot-sdk/pull/4171) - transactionWatch RPC has been stabilized to version 1. Useful RPC to submit an extrinsic and watch its status using `TransactionEvent`s
    - [#3512](https://github.com/paritytech/polkadot-sdk/pull/3512)
        - [Issue #3045](https://github.com/paritytech/polkadot-sdk/issues/3045): Runtime Omni Bencher - introduces a new binary crate `frame-omni-bencher` that allows to directly benchmark a WASM blob without needing a node or chain spec, making it easier to generate pallet weights. It works for all FRAME runtimes that don’t have 3rd party host calls (we don’t) or non `BlakeTwo256` block hashing (we do) so it should work for us. It is backwards compatible with the old CLI args too.
        - [Issue #2664](https://github.com/paritytech/polkadot-sdk/issues/2664): Build benchmark genesis state in the runtime - Adds `—runtime` and `—genesis-builder=none|runtime|spec` arguments to the `benchmark pallet` command
        - Simpler args for `PalletCmd::run`: `run()` is deprecated, we either remove all benchmark code from the Node and use the new omni-bencher or change it for `run_with_spec()`.
            - Note: changed runtime to use `run_with_spec` temporarily, although when we start benchmarking we should remove all benchmarking code from the node and use the `omni-bencher` as recommended.
    - [#4168](https://github.com/paritytech/polkadot-sdk/pull/4168) - chainHead RPC has been stabilized to version 1. Useful RPC to track the state of the head of the chain (finallized, non-finalized, best blocks), also to get those blocks’ bodies (txs), headers, read storage entries and call runtime APIs at a specified block’s state.
    - [#2714](https://github.com/paritytech/polkadot-sdk/pull/2714) - Added GenesisConfig presets for runtimes (e.g. `local`, `staging`, etc). To use them, the ChainSpecBuilder is extended with the `with_genesis_config_preset_name` method.
        - Note: the `GenesisBuilder` API changed to make use of these presets, I already updated it on our runtime but we don’t have any presets yet, it’s worth exploring if we could make good use of them. Usage example for Rococo [here](https://github.com/paritytech/polkadot-sdk/pull/2714/files#diff-3dc7410a664a21b2e83ff8481f604fdf803f5dfba36791971ee273543ebf7478)
    - [#2944](https://github.com/paritytech/polkadot-sdk/pull/2944) - `litep2p` integrated into Polkadot SDK improving performance but not fully replacing `rust-libp2p` yet.
        - Note: Generalized everything over the new `NetworkBackend` type, which can be good old libp2p or new flashy litep2p. Took a while 😮‍💨 
        - Note:  maybe it’s worth taking a more in-depth look at this since we rely on p2p comms as well?
    - [#3801](https://github.com/paritytech/polkadot-sdk/pull/3801) - Removed `slot_duration` from `aura::collators::basic::Params`, it now has to be fetched using Aura’s runtime API
        - Note: change done in `node/src/service.rs` + removed unneeded code from `node/src/command.rs` (logging parachain account)
- Changelog for `Runtime Dev`:
    - [#3485](https://github.com/paritytech/polkadot-sdk/pull/3485) - Adds runtime API for `pallet-broker` to get the current price of a core in the ongoing sale
    - [#3708](https://github.com/paritytech/polkadot-sdk/pull/3708) - QoL improvement for the XCM builder pattern: no longer are the .into() needed when constructing messages, it automatically converts the instruction parameters
        - Note: no changes needed since we don’t currently use the builder pattern for XCM, but good to know.
    - [#2119](https://github.com/paritytech/polkadot-sdk/pull/2119) - Adds example pallet for Multi-Block-Migrations under `frame/examples` and `MigrationId` to frame-support 
        - Note: not useful right now but useful if we ever need to do a MBM
    - [#2292](https://github.com/paritytech/polkadot-sdk/pull/2292) - Migrate fee payment from `Currency` to `Fungible`
        - Note: updated runtime with `FungibleAdapter` and `file-system` pallet since it was using the to-be-deprecated Currency trait
    - [#4006](https://github.com/paritytech/polkadot-sdk/pull/4006) - Fix `dynamic_params` name expand on `pallet-parameters`
        - Note: @ffarall tagging you just in case you have to modify anything on the parameters pallet
    - [#4075](https://github.com/paritytech/polkadot-sdk/pull/4075) - Adds ability to trigger tasks via unsigned transactions.
        - `validate_unsigned` hook for `frame-system` allows valid tasks to be submitted as unsigned transactions
        - The `is_valid` call on a task is the one that validates if the task is valid to be submitted as an unsigned transaction. As such, to prevent DoS attacks, it has to be cheap with minimal to no storage reads.
        - The unsigned tasks will be executed in a random order
        - Example in `substrate/frame/examples/tasks/`
        - Note: super useful for offchain workers! It might help us
    - [#4156](https://github.com/paritytech/polkadot-sdk/pull/4156) - Adds `AllowHrmpNotificationsFromRelayChain` barrier for safety (to ensure HRMP notifications originate solely from the Relay Chain)
        - Note: implemented in both the `xcm-simulator` and the actual runtime.
    - [#3721](https://github.com/paritytech/polkadot-sdk/pull/3721) - Adds `hrmp.establish_channel_with_system` call to allow parachains to establish a channel with a system parachain
    - [#4017](https://github.com/paritytech/polkadot-sdk/pull/4017) - Remove `TryRuntime` subcommand from node (we should use the standalone CLI https://github.com/paritytech/try-runtime-cli)
    - [#4266](https://github.com/paritytech/polkadot-sdk/pull/4266) - Reverts #3749, removing `execute_blob` and `send_blob` and restoring `execute` and `send`

- Extra changes:
    - [#4605](https://github.com/paritytech/polkadot-sdk/pull/4605) - Backport litep2p, locking it to 0.5.0 (since at release it was locked at the master branch which resulted in StorageHub not compiling because of the changes done since then to litep2p)
    - [#4584](https://github.com/paritytech/polkadot-sdk/pull/4584) - Backport `CheckMetadataHash` extension according to [RFC 0078](https://polkadot-fellows.github.io/RFCs/approved/0078-merkleized-metadata.html) (also check [#4274](https://github.com/paritytech/polkadot-sdk/pull/4274))


node/src/command.rs
```diff
@@ -273,53 +276,83 @@ pub fn run() -> Result<()> {
...
+								indexer_options,
+								id,
+								hwbench,
+							)
+							.await
+							.map(|r| r.0)
+							.map_err(Into::into)
+						}
+					},
+					sc_network::config::NetworkBackendType::Litep2p => {
```
snowmead, line 316: What is the reason behind supporting both network backends?

snowmead (approved):
All looks good, just wanted to clarify the use of having both rust-libp2p and litep2p network backends.

node/src/command.rs
TDemeco: Mostly because `litep2p` is not complete and might cause issues with the file transfers and everything we have to do appart from consensus and gossiping

client/file-transfer-service/src/handler.rs
```diff
@@ -339,7 +341,7 @@ impl FileTransferService {
     pub fn new(
         protocol_name: ProtocolName,
         request_receiver: async_channel::Receiver<IncomingRequest>,
-        network: Arc<ParachainNetworkService>,
+        network: Arc<dyn NetworkService>,
```
ffarall, line 344: What's up with this change?

client/file-transfer-service/src/lib.rs
```diff
@@ -50,18 +53,25 @@ pub fn configure_file_transfer_network(
...
 
-    let protocol_name = protocol_config.name.clone();
+    let request_response_config = Network::request_response_config(
+        protocol_config.name.clone(),
+        protocol_config.fallback_names.clone(),
+        protocol_config.max_request_size,
+        protocol_config.max_response_size,
+        protocol_config.request_timeout,
+        protocol_config.inbound_queue,
+    );
```
ffarall, line 63: And what about this?

node/src/command.rs
ffarall: Does the template node support both too?

node/src/service.rs
```diff
@@ -923,34 +934,24 @@ fn build_import_queue(
...
+        _,
+        _,
+        _,
+    >(
+        client,
+        block_import,
+        move |_, _| async move {
+            let timestamp = sp_timestamp::InherentDataProvider::from_system_time();
 
-                Ok((slot, timestamp))
```
ffarall, line 947: What are the consequences of removing this?

runtime/src/configs/xcm_config.rs
```diff
@@ -140,6 +140,8 @@ pub type Barrier = TrailingSetTopicAsId<
                     AllowExplicitUnpaidExecutionFrom<ParentOrParentsExecutivePlurality>,
                     // Subscriptions for version tracking are OK.
                     AllowSubscriptionsFrom<ParentRelayOrSiblingParachains>,
+                    // HRMP notifications from the relay chain are OK.
+                    AllowHrmpNotificationsFromRelayChain,
```
ffarall, line 144: Does this mean that HrmpNotifications from other origins will NOT be accepted?

test/util/constants.ts
```diff
@@ -1,3 +1,3 @@
 export const UNIT = 1_000_000_000_000n;
-export const ROUGH_TRANSFER_FEE = 1_472_378_298n; // This is just naive copypaste from explorer
+export const ROUGH_TRANSFER_FEE = 1_482_409_632n; // This is just naive copypaste from explorer
```
ffarall, line 2: What's up with this change?

ffarall (approved):
Again, outstanding PR docs.

node/src/command.rs
ffarall: Yes

PR #213, opened 2024-10-01
feat: adding payment stream creation event to indexer

undercover-cactus:
This PR add the logic to have the indexer processing the payment stream creation events.

We are adding a new table `PaymentStream` that aggregate the payment stream created (with fixed rate and dynamic rate). The table has an extra column `total_amount` that keep record of the total amount paid to the provider (not yet implemented).


client/indexer-db/src/models/paymentstream.rs
```diff
@@ -0,0 +1,31 @@
...
+#[derive(Debug, Queryable, Insertable, Selectable)]
+#[diesel(table_name = paymentstream)]
+pub struct PaymentStream {
+    // postgres attributed ID for this payment stream
+    pub id: i32,
+    // Account ID of the payer
+    pub account: String,
+    // ID of the payee (msp or bsp)
+    pub provider: String,
+    // Total amount already paid to this provider
```
links234, line 15: You mean to this provider from this account. If I want total amount for provider I need to get the sum.

client/indexer-service/src/handler.rs
```diff
@@ -238,17 +238,46 @@ impl IndexerService {
...
+            pallet_payment_streams::Event::FixedRatePaymentStreamCreated {
+                provider_id,
+                user_account,
+                rate: _rate,
+            } => {
+                PaymentStream::create(conn, provider_id.to_string(), user_account.to_string())
+                    .await?;
+            }
+            pallet_payment_streams::Event::FixedRatePaymentStreamUpdated { .. } => {
+                // Currently we are not treating the info of fixed rate update
```
ffarall, line 266:
```suggestion
                // TODO: Currently we are not treating the info of fixed rate update
```

ffarall (changes requested):
Great job! Neat and tidy code. Just some changes required mostly around naming and conventions.

pallets/payment-streams/src/lib.rs
```diff
@@ -486,8 +494,10 @@ pub mod pallet {
             // Check that the extrinsic was executed by the root origin
             ensure_root(origin)?;
 
+            let current_tick = Self::get_current_tick();
+
             // Execute checks and logic, update storage
-            Self::do_delete_fixed_rate_payment_stream(&provider_id, &user_account)?;
+            Self::do_delete_fixed_rate_payment_stream(&provider_id, &user_account, current_tick)?;
```
ffarall, line 500: `do_delete_fixed_rate_payment_stream` can lookup the `current_tick` on its own, it doesn't need to receive it as a parameter.

pallets/payment-streams/src/lib.rs
```diff
@@ -573,11 +583,14 @@ pub mod pallet {
...
             ensure_root(origin)?;
 
+            let current_tick = Self::get_current_tick();
+
             // Execute checks and logic, update storage
             Self::do_update_dynamic_rate_payment_stream(
                 &provider_id,
                 &user_account,
                 new_amount_provided,
+                current_tick,
```
ffarall, line 593: `do_update_dynamic_rate_payment_stream` can lookup the `current_tick` on its own, it doesn't need to receive it as a parameter.

pallets/payment-streams/src/lib.rs
```diff
@@ -616,8 +629,10 @@ pub mod pallet {
             // Check that the extrinsic was executed by the root origin
             ensure_root(origin)?;
 
+            let current_tick = Self::get_current_tick();
+
             // Execute checks and logic, update storage
-            Self::do_delete_dynamic_rate_payment_stream(&provider_id, &user_account)?;
+            Self::do_delete_dynamic_rate_payment_stream(&provider_id, &user_account, current_tick)?;
```
ffarall, line 635: Same

pallets/payment-streams/src/lib.rs
```diff
@@ -674,11 +689,15 @@ pub mod pallet {
...
 
+            // Get the last tick to add it to the event
+            let last_tick_charged = Self::get_current_tick();
+
             // Emit the corresponding event (we always emit it even if the charged amount was 0)
             Self::deposit_event(Event::<T>::PaymentStreamCharged {
                 user_account,
                 provider_id: provider_id,
                 amount: amount_charged,
+                last_tick_charged,
```
ffarall, line 700: Same

pallets/payment-streams/src/lib.rs
```diff
@@ -447,8 +448,15 @@ pub mod pallet {
...
+            let current_tick = Self::get_current_tick();
+
             // Execute checks and logic, update storage
-            Self::do_update_fixed_rate_payment_stream(&provider_id, &user_account, new_rate)?;
+            Self::do_update_fixed_rate_payment_stream(
+                &provider_id,
+                &user_account,
+                new_rate,
+                current_tick,
+            )?;
```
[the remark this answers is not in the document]
ffarall: Regardless, it shouldn't be the `current_tick` that is used here, but rather the tick up to which this payment stream was charged, which is not necessarily the same. In fact in most cases, it will differ. For BSPs for instance, it is the last tick for which they submitted a proof for.

pallets/payment-streams/src/lib.rs
ffarall: For example, here in `do_charge_payment_streams`, it would be the `last_chargeable_tick` variable used there.

ffarall (changes requested):

client/indexer-service/src/handler.rs
```diff
@@ -238,17 +238,53 @@ impl IndexerService {
...
+                // We want to handle this and update the payment stream total amount
+                let ps =
+                    PaymentStream::get(conn, user_account.to_string(), provider_id.to_string())
+                        .await?;
+                let new_total_amount = ps.total_amount_paid + amount;
+                PaymentStream::update_total_amount(
+                    conn,
+                    ps.id,
+                    new_total_amount,
+                    *last_tick_charged as i64,
```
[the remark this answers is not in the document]
undercover-cactus: I am using i64 here because postgres only allow for i32 and i64. i64 would fit all the values possible of u32 so it looked like a good solution.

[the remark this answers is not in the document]
ffarall: I agree, provided that we stick with `u32` as the `BlockNumber` type (which will most likely be the case). If we don't, and use `u64`, do we have any alternative? Given that postgres only allows `i32` and `i64`?

Anyway, if we stick to `i64`, I'd prefer it if we use `try_into` and log some error if it fails. The `as` syntax will truncate silently and we might not notice it until it's too late.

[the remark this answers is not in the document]
rllola: If we want bigger we can use `Numeric`. We are already using it for amount.

Ok I can change it for `try_into`

PR #215, opened 2024-10-04
feat: :sparkles:: NewStorageRequest handled in user send file tasks

snowmead:
Added `NewStorageRequest` handler event for `UserSendsFileTask`. It also modifies the MSP and user tasks to skip storage request events if it isn't intended for them.

Add `query_provider_multiaddresses` runtime api.

Common alias type to represent multiple multiaddresses for MSPs and BSPs.

An integration test is skipped since it would require an MSP to be part of the network to confirm files. It shall be moved in a subsequent change.

_This is a preliminary change before working on msp tasks and integration tests. It is separated to handle any errors that would occur in the old testing network setup for that only handled BSPs._


test/util/bspNet/helpers.ts
```diff
@@ -253,7 +253,8 @@ export const runSimpleBspNet = async (bspNetConfig: BspNetConfig, verbose = fals
           alice.address,
           ShConsts.DUMMY_MSP_ID,
           bspNetConfig.capacity || ShConsts.CAPACITY_512,
-          [multiAddressBsp],
+          // The peer ID has to be different from the BSP's since the user now attempts to send files to MSPs when new storage requests arrive.
+          ["/ip4/127.0.0.1/tcp/30350/p2p/12D3KooWNEZ8PGNydcdXTYy1SPHvkP9mbxdtTqGGFVrhorDzeTfA"],
```
ffarall, line 257: We might want to make this a constant

node/src/tasks/user_sends_file.rs
```diff
@@ -121,84 +239,26 @@ where
...
-                }
-            }
-            info!(target: LOG_TARGET, "Successfully sent file {:?} to peer {:?}", file_metadata.fingerprint, peer_id);
-            return Ok(());
+        match self.send_chunks_to_provider(peer_ids, &file_metadata).await {
+            Some(result) => result,
+            None => Err(anyhow::anyhow!(
+                "Failed to send file key {:?} to any of the peers",
+                file_key
+            )),
```
ffarall, line 261: Why not just forwarding the error from `send_chunks_to_provider`? They should be more accurate.

node/src/tasks/user_sends_file.rs
```diff
@@ -48,6 +51,81 @@ where
             storage_hub_handler,
         }
     }
+
+    async fn send_chunks_to_provider(
```
ffarall, line 55: I'd prefer to have these utility functions at the bottom of the file, so that it is easier to see what events this task handles, and therefore how is the flow.

client/common/src/blockchain_utils.rs
```diff
@@ -48,3 +50,28 @@ pub fn get_events_at_block(
         .transpose()?
         .ok_or(EventsRetrievalError::StorageNotFound)
 }
+
+/// Attempt to convert BoundedVec of BoundedVecs of bytes.
+///
+/// Returns a list of [`Multiaddr`] objects that have successfully been parsed from the raw bytes.
+pub fn convert_raw_multiaddresses_to_multiaddr(multiaddresses: Multiaddresses) -> Vec<Multiaddr> {
```
ffarall, line 57: Not sure this fits in `blockchain_utils`, but I don't mind much anyway.

ffarall (approved):
Approved with comments!

client/common/src/blockchain_utils.rs
snowmead: We were using this in the blockchain service and in the task, so I figured this is a common area to put it in.

node/src/tasks/user_sends_file.rs
snowmead: That means I would have to declare another impl block for `UserSendsFileTask`. Or would you like me to move the current impl to the bottom of the file?

PR #219, opened 2024-10-07
feat: Handle proofs events in the indexer

undercover-cactus:
In this PR we index the events related to accepted proof and the event related to slashing.

The slashing event modify the bsp table to change the total amount slashed and the latest proof accepted update the last proven tick.

NOTE: A small modification was made for the runtime providers arguments name to follow the convention (`who` for account `provider_id` for provider `bsp_id` for BSP provider id).


pallets/proofs-dealer/runtime-api/src/lib.rs
```diff
@@ -27,6 +28,7 @@ sp_api::decl_runtime_apis! {
         fn get_forest_challenges_from_seed(seed: &RandomnessOutput, provider_id: &ProviderId) -> Vec<Key>;
         fn get_current_tick() -> BlockNumber;
         fn get_next_deadline_tick(provider_id: &ProviderId) -> Result<BlockNumber, GetNextDeadlineTickError>;
+        fn get_stake(provider_id: &ProviderId) -> Option<Balance>;
```
undercover-cactus, line 31: I am not sure I have added to right pallet.

pallets/proofs-dealer/runtime-api/src/lib.rs
```diff
@@ -7,13 +7,14 @@ use sp_std::vec::Vec;
 
 sp_api::decl_runtime_apis! {
     #[api_version(1)]
-    pub trait ProofsDealerApi<ProviderId, BlockNumber, Key, RandomnessOutput, TrieRemoveMutation>
+    pub trait ProofsDealerApi<ProviderId, BlockNumber, Key, RandomnessOutput, TrieRemoveMutation, Balance>
```
ffarall, line 10: This seems to be unnecessary. I assume it's a leftover from when you were adding the runtime API here.

pallets/providers/src/lib.rs
```diff
@@ -525,6 +525,10 @@ pub mod pallet {
         InvalidEncodedAccountId,
         /// Error thrown when trying to update a payment stream that does not exist.
         PaymentStreamNotFound,
+
+        // Provider interface errors
+        /// The provider stake could not be found.
+        ProviderStakeNotFound,
```
ffarall, line 531: This doesnt seem to be used anywhere else.

pallets/proofs-dealer/src/lib.rs
```diff
@@ -386,6 +386,7 @@ pub mod pallet {
         ProofAccepted {
             provider: ProviderIdFor<T>,
             proof: Proof<T>,
+            last_tick_proof: BlockNumberFor<T>,
```
ffarall, line 389: I would rename this to `last_tick_proven` and set it to the last tick for which this `provider` submitted a proof. That tick is not the `current_tick` when the `provider` submits a proof, it's in fact a previous tick that in `do_submit_proof` is under the variable name `challenges_tick`.

pallets/proofs-dealer/src/types.rs
```diff
@@ -154,3 +154,8 @@ pub type BlockFullnessHeadroomFor<T> = <T as crate::Config>::BlockFullnessHeadro
 
 /// Syntactic sugar for MinNotFullBlocksRatio type used in the ProofsDealer pallet.
 pub type MinNotFullBlocksRatioFor<T> = <T as crate::Config>::MinNotFullBlocksRatio;
+
+/// BalanceOf is the balance type of the runtime.
+pub type BalanceOf<T> = <<T as crate::Config>::NativeBalance as fungible::Inspect<
+    <T as frame_system::Config>::AccountId,
+>>::Balance;
```
ffarall, line 161: This seems unnecessary

client/indexer-db/src/models/proofs.rs
```diff
@@ -0,0 +1,55 @@
...
+    pub id: i32,
+    pub provider_id: String,
+    pub last_tick_proof: i64,
+}
+
+impl Proofs {
+    pub async fn create<'a>(
+        conn: &mut DbConnection<'a>,
+        provider_id: String,
+        last_tick_proof: i64,
```
ffarall, line 19: I'd rename this to `last_tick_proven`, or else have two fields stored here: `last_tick_proven` and `proof_submitted_at_tick` (or similar name). The former would be the last tick for which this `provider` submitted a proof, and the latter would be the tick at which that proof was submitted.

client/indexer-service/src/handler.rs
```diff
@@ -345,6 +365,7 @@ impl IndexerService {
                     capacity.into(),
                     sql_multiaddresses,
                     bsp_id.to_string(),
+                    stake,
```
ffarall, line 368: The stake field should also be changed in the `CapacityChanged` event.

client/indexer-service/src/handler.rs
```diff
@@ -332,6 +342,16 @@ impl IndexerService {
                 multiaddresses,
                 capacity,
             } => {
+                let current_block_hash = self.client.info().best_hash;
+
+                let stake = self
+                    .client
+                    .runtime_api()
+                    .get_bsp_stake(current_block_hash, bsp_id)
+                    .expect("to have a stake")
```
ffarall, line 351: Up for discussion whether or not we want to leave the possibility of a panic here 🤔

ffarall (changes requested):

client/indexer-service/src/handler.rs
links234: Keep in mind that any error here will result in the retry to index on the next finality notification.

While I agree with "if something really bad and unrecoverable happened, let's abort" I'm not sure if there is any case where the result may be different later (in which case would be recoverable).

ffarall: Agreed, if this happened I wouldn't even know why it is happening. It would be an error in the runtime even. So panic is fine. This would be unrecoverable.

client/indexer-db/migrations/2024-09-26-145832_create_bsp/up.sql
```diff
@@ -4,7 +4,8 @@ CREATE TABLE bsp (
     account VARCHAR NOT NULL,
     capacity NUMERIC(20, 0) NOT NULL,
     created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
-    updated_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
+    updated_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    stake NUMERIC(38, 0) NOT NULL DEFAULT 0
```
links234, line 8: Maybe switch the order of fields so that `created_at` and `update_at` are the last ones?

client/indexer-db/src/models/proofs.rs
```diff
@@ -0,0 +1,55 @@
+use diesel::prelude::*;
+use diesel_async::RunQueryDsl;
+
+use crate::{schema::proofs, DbConnection};
+
+/// A single record table that holds the state/metadata of the indexer service.
+#[derive(Debug, Queryable, Insertable, Selectable)]
+#[diesel(table_name = proofs)]
+pub struct Proofs {
```
links234, line 9: I see we don't keep anything else besides the `last_tick_proven`. Why a new table?
If this is meant only to simplify the abstract `provider_id` (which can be BSP or MSP) - why keeping multiple entries?

client/indexer-db/src/models/proofs.rs
```diff
@@ -0,0 +1,55 @@
...
+    pub id: i32,
+    pub provider_id: String,
```
links234, line 11: What do you think of renaming it to `onchain_provider_id` so it's more clear that it is not a DB id?

client/indexer-service/src/handler.rs
```diff
@@ -361,8 +382,20 @@ impl IndexerService {
                 old_capacity: _old_capacity,
                 next_block_when_change_allowed: _next_block_when_change_allowed,
             } => match provider_id {
-                StorageProviderId::BackupStorageProvider(_) => {
+                StorageProviderId::BackupStorageProvider(bsp_id) => {
                     Bsp::update_capacity(conn, who.to_string(), new_capacity.into()).await?;
+
+                    // update also the stake
+                    let current_block_hash = self.client.info().best_hash;
```
links234, line 389: This most often will not be the correct `current_block_hash` - as in the block we are currently processing.

You should pass and use the `block_hash` parameter from `index_block` method.

links234 (changes requested):
Looks good overall! Left some minor nitpicks and problem that needs to be addressed.

client/indexer-service/src/handler.rs
undercover-cactus: The other option is to skip the update of the stake value and we just have an error message. But the database would be inconsistent with what is really happening in chain.

Or we retry until we have an answer which could block the indexing process if the issue cannot be solved on its own.

client/indexer-db/src/models/proofs.rs
undercover-cactus: I added a new table because at first I though we wanted to save all the proofs in the database but indeed now it doesn't make sense.

pallets/proofs-dealer/src/lib.rs
```diff
@@ -596,8 +597,18 @@ pub mod pallet {
 
             Self::do_submit_proof(&provider, &proof)?;
 
+            // Get last tick for which the submitter submitted a proof.
+            let last_tick_proven = match LastTickProviderSubmittedAProofFor::<T>::get(provider) {
+                Some(tick) => tick,
+                None => return Err(Error::<T>::NoRecordOfLastSubmittedProof.into()),
+            };
```
ffarall, line 604: I would prefer it if `do_submit_proof` returns the `last_tick_proven`. We avoid checking for an impossible error, and keep this logic out of `lib.rs`

ffarall (approved):
Great job, approved with just one minor comment

client/indexer-service/src/handler.rs
```diff
@@ -403,7 +441,22 @@ impl IndexerService {
             } => {
                 Msp::delete(conn, who.to_string()).await?;
             }
-            pallet_storage_providers::Event::Slashed { .. } => {}
+            pallet_storage_providers::Event::Slashed {
+                provider_id,
+                amount_slashed: _amount_slashed,
+            } => {
+                let current_block_hash = self.client.info().best_hash;
```
links234, line 448: Shouldn't this be `block_hash` instead of `self.client.info().best_hash` too?

links234 (changes requested):

links234 (approved):

PR #220, opened 2024-10-07
feat: index file requests, files and BSPs storing them 

links234:


undercover-cactus (approved):

client/indexer-db/src/models/file.rs
```diff
@@ -0,0 +1,115 @@
...
+
+impl File {
+    pub async fn create<'a>(
+        conn: &mut DbConnection<'a>,
+        account: impl Into<String>,
+        file_key: impl Into<Vec<u8>>,
+        bucket_id: i32,
+        location: impl Into<Vec<u8>>,
+        fingerprint: impl Into<Vec<u8>>,
+        size: i64,
```
snowmead, line 54: Can we use `FileStorageRequestStep` as we do in `update_step`?

client/indexer-service/src/handler.rs
```diff
@@ -212,16 +212,84 @@ impl IndexerService {
...
-            pallet_file_system::Event::BspConfirmStoppedStoring { .. } => {}
-            pallet_file_system::Event::BspConfirmedStoring { .. } => {}
+            pallet_file_system::Event::BspConfirmStoppedStoring {
+                bsp_id,
+                file_key,
+                new_root: _,
+            } => {
+                let bsp = Bsp::get_by_onchain_bsp_id(conn, bsp_id.to_string()).await?;
+                let file = File::get_by_file_key(conn, file_key.as_ref().to_vec()).await?;
+                BspFile::create(conn, bsp.id, file.id).await?;
```
snowmead, line 222: Why are we inserting instead of deleting?

snowmead (approved):
Nice, just a couple of comments.

client/indexer-db/src/models/file.rs
links234: Not sure what you are referring to.

client/indexer-db/src/models/file.rs
```diff
@@ -0,0 +1,127 @@
...
+pub struct File {
+    /// The ID of the file as stored in the database. For the runtime id, use `onchain_bsp_id`.
+    pub id: i32,
+    pub account: String,
+    pub file_key: Vec<u8>,
+    pub bucket_id: i32,
+    pub location: Vec<u8>,
+    pub fingerprint: Vec<u8>,
+    pub size: i64,
+    /// The step this file is at. 0 = requested, 1 = fullfilled.
```
ffarall, line 30:
```suggestion
    /// The step this file is at. 0 = requested, 1 = fulfilled.
```

client/indexer-db/src/models/peer_id.rs
```diff
@@ -0,0 +1,29 @@
...
+use crate::{schema::peer_id, DbConnection};
+
+/// Table that holds the list of peers that the indexer is interested in.
+#[derive(Debug, Queryable, Insertable, Selectable)]
+#[diesel(table_name = peer_id)]
+pub struct PeerId {
+    pub id: i32,
+    pub peer: Vec<u8>,
+    pub created_at: NaiveDateTime,
+    pub updated_at: NaiveDateTime,
```
ffarall, line 14: Why a separate table for this? I

client/indexer-db/src/models/bucket.rs
```diff
@@ -80,4 +80,15 @@ impl Bucket {
             .await?;
         Ok(bucket)
     }
+
+    pub async fn get_by_onchain_bucket_id<'a>(
```
ffarall, line 84: Not specific to this PR, but are we not storing the Bucket root?

client/indexer-db/src/models/file.rs
```diff
@@ -0,0 +1,127 @@
...
+#[derive(Debug, Queryable, Insertable, Associations)]
+#[diesel(table_name = file_peer_id)]
+#[diesel(belongs_to(File, foreign_key = file_id))]
+#[diesel(belongs_to(PeerId, foreign_key = peer_id))]
+pub struct FilePeerId {
+    pub file_id: i32,
+    pub peer_id: i32,
+}
+
+impl File {
```
ffarall, line 46: I think it is important to add a `get_by_bucket`. That is actually the main reason why we have an indexer in the first place.

client/indexer-service/src/handler.rs
```diff
@@ -212,16 +212,82 @@ impl IndexerService {
...
                 .await?;
             }
-            pallet_file_system::Event::BspConfirmStoppedStoring { .. } => {}
-            pallet_file_system::Event::BspConfirmedStoring { .. } => {}
+            pallet_file_system::Event::BspConfirmStoppedStoring {
+                bsp_id: _,
+                file_key,
+                new_root: _,
+            } => {
+                File::delete(conn, file_key.as_ref().to_vec()).await?;
```
ffarall, line 220: Isn't this deleting the file from the database if just one BSP stops storing it? That would be wrong.

client/indexer-db/src/models/bsp.rs
```diff
@@ -90,4 +92,47 @@ impl Bsp {
...
+}
+
+#[derive(Debug, Queryable, Insertable, Selectable)]
+#[diesel(table_name = bsp_file)]
+pub struct BspFile {
+    pub bsp_id: i32,
+    pub file_id: i32,
+}
+
+impl BspFile {
```
ffarall, line 126: From what I understand, this links a `File` to a `bsp_id`, so that we only store the `File` information once, but then we have several BSPs that reference it as being a part of their forests, right?

client/indexer-db/src/models/file.rs
```diff
@@ -0,0 +1,127 @@
...
+    Requested = 0,
+    Fullfilled = 1,
+    Expired = 2,
+    Revoked = 3,
+}
+
+/// Table that holds the Files (both ongoing requests and completed).
+#[derive(Debug, Queryable, Insertable, Selectable)]
+#[diesel(table_name = file)]
+pub struct File {
```
ffarall, line 21: Don't we want to store the BSPs storing this file? It's one of the main reasons for the indexer.

client/indexer-service/src/handler.rs
```diff
@@ -212,16 +212,82 @@ impl IndexerService {
...
+            pallet_file_system::Event::StorageRequestExpired { file_key } => {
+                File::update_step(
+                    conn,
+                    file_key.as_ref().to_vec(),
+                    FileStorageRequestStep::Expired,
+                )
+                .await?;
+            }
+            pallet_file_system::Event::StorageRequestRevoked { file_key } => {
+                File::update_step(
```
ffarall, line 284: Shouldn't we delete it here?

client/indexer-service/src/handler.rs
```diff
@@ -212,16 +212,82 @@ impl IndexerService {
...
+            pallet_file_system::Event::StorageRequestFulfilled { file_key } => {
+                File::update_step(
+                    conn,
+                    file_key.as_ref().to_vec(),
+                    FileStorageRequestStep::Fullfilled,
+                )
+                .await?;
+            }
+            pallet_file_system::Event::StorageRequestExpired { file_key } => {
+                File::update_step(
```
ffarall, line 276: If it has no BSPs here, we should delete it

ffarall (changes requested):

client/indexer-db/src/models/file.rs
links234: Agree. You want to add it in this PR when we will also use it?

client/indexer-db/src/models/bucket.rs
links234: Not in this PR, no. I was planning to add root for both buckets and BSPs in another PR.

Is this fine?

client/indexer-db/src/models/peer_id.rs
links234: Just the way relational databases best handle arrays. Otherwise would be hard (if not impossible) to query for example by `PeerId`.

client/indexer-db/src/models/bsp.rs
links234: Yes. 
We will have multiple entries here, 1 for each BSP storing the File. Also there are delete triggers setup in Postgres so that if any of the BSP or the File is deleted the relationship entry from this table is also deleted.

client/indexer-db/src/models/bucket.rs
ffarall: Yeah that's fine. But we have to be aware that until then we want be able to use the main functionality of our indexer.

client/indexer-db/src/models/file.rs
ffarall: Add it in this PR if possible

client/indexer-db/src/models/file.rs
links234: Yes, this is stored in the `BspFile` table.

ffarall (approved):

PR #221, opened 2024-10-07
feat: MSP respond storage requests tasks

snowmead:
Implements MSP tasks to receive files from users after storage requests have been issued and respond to storage requests in batches to the runtime.

The MSP will listen to the event contain the results after responding to them and insert only the file keys that have been successfully accepted by the runtime.

Implementation:

- File keys to respond to by the MSP are batched up to a maximum of `MaxBatchMspRespondStorageRequests` (runtime configuration) file keys which gets emitted and processed by the `MspUploadFileTask` via the `ProcessMspRespondStoringRequest` event.
- `ForestStorageHandler` is now implemented by each storage layer so when we call `insert`, the appropriate forest storage will be instantiated. This keeps the tasks still generic over the forest storage handler.
- User node now retries uploading the chunks to the peer. This was an issue when the MSP hasn't yet registered the user as a peer to accept the chunks before the user started establishing a connection.

Notable utilities:

- Added `watch_for_success_with_events` transaction function to retrieve the events emitted from the executed transaction. This enables the possibility for tasks to wait for a transaction to succeed and enact any addtional operations based on emitted events. In the case of the `ProcessMspRespondStoringRequest` event, the event containing the results of the the batch response is used to react to the accepted, rejected and failed file keys from the runtime.
- RPC methods including forest keys now accept an `Option<H256>` which would be the bucket id in the case of an MSP node storing multiple forests identified by their bucket ids. 
- `StorageLayerBuilder` impls are now generic over forest storage handlers.
- Added `test:mspnet` for MSP related integration tests.

TODOs:

- Add integration tests for file key rejections (e.g. when lack of storage capacity, or file key already exists).
- Handle cases where file metadata cannot be retrieved for some reason after having the file keys be accepted and applied to the runtime.
- Handle case where the file key failed to be inserted into bucket forest.
- Handle case when the local bucket root does not match runtime bucket root after applied delta.


test/util/bspNet/waits.ts
```diff
@@ -143,6 +143,53 @@ export const waitForBspStored = async (api: ApiPromise, checkQuantity?: number)
...
+ * 2. Checks for the presence of a 'mspRespondStorageRequestsMultipleBuckets' extrinsic in the transaction pool.
+ * 3. Seals a block and verifies the presence of an 'MspRespondedToStorageRequests' event.
+ *
+ * @param api - The ApiPromise instance to interact with the blockchain.
+ * @param checkQuantity - Optional param to specify the number of expected extrinsics.
+ * @returns A Promise that resolves when a MSP has sent a response to storage requests.
+ *
+ * @throws Will throw an error if the expected extrinsic or event is not found.
+ */
+export const waitForMspResponse = async (api: ApiPromise, checkQuantity?: number) => {
```
ffarall, line 161: It's kinda weird that this is under `bspNet`. A refactor is likely in order.

test/util/bspNet/testrunner.ts
```diff
@@ -110,6 +114,87 @@ export async function describeBspNet<
   }
 }
 
+/**
+ * Implementation of the describeBspNet function.
+ * @param title The title of the test suite.
+ * @param args Additional arguments (either tests function or options and tests function).
+ */
+export async function describeMspNet<
```
ffarall, line 122: Same here. This needs a refactor.

test/util/bspNet/testrunner.ts
```diff
@@ -110,6 +114,87 @@ export async function describeBspNet<
...
+        if (options?.keepAlive) {
+          if (fullNetConfigCases.length > 1) {
+            console.error(
+              `test run configured for multiple bspNetConfigs, only ${JSON.stringify(
+                fullNetConfig
+              )} will be kept alive`
+            );
+          }
+          console.log("🩺 Info:  Test run configured to keep BSPNet alive");
+          console.log("ℹ️ Hint: close network with:   pnpm docker:stop:bspnet  ");
```
ffarall, line 174: Shouldn't this say something different like fullnet?

ffarall (approved):

test/util/bspNet/waits.ts
snowmead: I put this here to avoid potential conflicts with Pablo.

PR #222, opened 2024-10-07
feat: ⏫ upgrade to Polkadot SDK stable2407

TDemeco:
### v1.14.0 to stable2407 -  Polkadot SDK Upgrade

- Changelog for `Node Dev`:
    - [#4848](https://github.com/paritytech/polkadot-sdk/pull/4848) - Optimized logic for gossiping assignments, reducing cpu usage by 15% to 20%
    - [#4922](https://github.com/paritytech/polkadot-sdk/pull/4922) - Optimized finalization performance with a quicker, more efficient, less complex version.
    - [#4755](https://github.com/paritytech/polkadot-sdk/pull/4755) - Send `PeerViewChange` message with high priority, updates `orchestra` to 0.4.0. This improves propagation times of gossiped messages.
    - [#4937](https://github.com/paritytech/polkadot-sdk/pull/4937) - `prospective-parachains` rework, again
        - Note: the change of removing support for parachain forks was reverted, since there were valid scenarios where a parachain wanted to fork.
    - [#5169](https://github.com/paritytech/polkadot-sdk/pull/5169) & [#5367](https://github.com/paritytech/polkadot-sdk/pull/5367) - Bump `serde_json` from 1.0.120 to 1.0.121
        - Note: bumped in StorageHub as well.
    - [#5393](https://github.com/paritytech/polkadot-sdk/pull/5393) - Add feature (`full-pov-size`) to allow Aura collators to use the full PoV size instead of half of it
        - Note: no changes needed from our side in the code, but we could experiment enabling this.
- Changelog for `Runtime Dev`:
    - [#5113](https://github.com/paritytech/polkadot-sdk/pull/5113) - Make the candidate relay parent progression check more strict for elastic scaling.
        - Note: already backported to v1.14.0 so it’s included in the previous update.
    - [#4863](https://github.com/paritytech/polkadot-sdk/pull/4863) - Make `tracing::log` work in the runtime
    - [#4777](https://github.com/paritytech/polkadot-sdk/pull/4777) - XCM builder pattern allows `clear_origin` before `buy_execution`.
        - Note: I did not use the builder pattern in tests since it wasn’t allowed to do this (which by all XCM standards was supossed to be allowed), so XCM’s from `sh-xcm-simulator` are built from vectors. I don’t think it’s worth to refactor them since it’s logically equivalent.
    - [#5581](https://github.com/paritytech/polkadot-sdk/pull/5581) - Clear the message queue before executing a dry-run of an XCM, so the only messages left are the ones the users of the API actually care about.
        - Note: implemented for `sh-xcm-simulator`.
    - [#5644](https://github.com/paritytech/polkadot-sdk/pull/5644) - Improve weight annotations, preventing allocations when calculating the weights.
    - [#5466](https://github.com/paritytech/polkadot-sdk/pull/5466) - `frame-omni-bencher` patch:
        - Note: main thing for us is the addition of a README with rust docs for usage.
    - [#5580](https://github.com/paritytech/polkadot-sdk/pull/5580) - Improve error message for pallet macro generated code, for calls, to ensure they return DispatchResult or DispatchResultWithPostInfo
- Changelog for `Node Operator`:
    - [#4097](https://github.com/paritytech/polkadot-sdk/pull/4097) - Introduce basic slot-based collator for elastic scaling. This is a new experimental collator variant that should only be used for testing and exploration.
        - Note: the differences with the `lookahead` collator are:
            - The new variant is slot-based (it waits for the next slot of the parachain to start authoring)
            - It uses the current best relay parent as anchor
            - It tends to be anchored to one relay parent earlier. While `lookahead` generally waits for a new relay block to arrive before it attemps to build a block, this new slot-based one starts authoring directly on the slot boundary, where a new relay chain block has probably not yet arrived.
    - [#5040](https://github.com/paritytech/polkadot-sdk/pull/5040) - Update libp2p-websocket to v0.42.2.
        - Note: This was backported to v1.14.0 so it was already included in the previous update.
    - [#4663](https://github.com/paritytech/polkadot-sdk/pull/4663) - Add elastic scaling MVP guide, available [here](https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_docs/guides/enable_elastic_scaling_mvp/index.html)
        - Note: since this is quite experimental it doesn’t make sense to implement it yet, but it’s nice to have a reference to the guide we will probably use in the future.

### Addition
This PR also includes a TON of fixes for the integration tests. It seems everything is working correctly now, BUT a few errors here and there are expected because of the nature of the tests and the limitations that arise with using GitHub workers.


test/suites/integration/bsp/automatic-tipping.test.ts
```diff
@@ -15,7 +15,7 @@ describeBspNet(
       await api.file.newStorageRequest("res/whatsup.jpg", "test/whatsup.jpg", "nothingmuch-2");
       await api.wait.bspVolunteer();
 
-      await sleep(1000); // wait for the bsp to download the files
+      await api.wait.bspStoredInTxPool();
```
snowmead, line 18: Can you remove the `extrinsicPresent` call below since this does the same thing?

test/suites/integration/bsp/volunteer.test.ts
```diff
@@ -233,26 +233,36 @@ describeBspNet("Multiple BSPs volunteer ", ({ before, createBspApi, createUserAp
...
-      module: "fileSystem",
-      method: "bspConfirmStoring",
-      checkTxPool: true,
-      assertLength: 1,
-      timeout: 10000
-    });
+    // Wait for the BSP to receive and store all files
+    for (let i = 0; i < source.length; i++) {
+      const fileKey = fileKeys[i];
+      await bspApi.assert.log({
```
snowmead, line 256: I don't think we should depend on logs. I would use the rpc method `isFileInFileStorage`, that way we don't need to maintain log outputs.

snowmead (approved):
Looks good, just a couple of comments

test/suites/integration/bsp/automatic-tipping.test.ts
TDemeco: Done!

test/suites/integration/bsp/volunteer.test.ts
TDemeco: Changed logs for a specific wait function that waits to get `FileFound` from the `isFileInFileStorage` RPC call

ffarall (approved):
Amazing job!

PR #223, opened 2024-10-08
feat: ✨ Update price of data according to system utilisation

ffarall:
This PR:
1. Adds the `UpdateStoragePrice` trait to `primitives/traits`, which defines a `update_storage_price` method.
2. Adds `UpdateStoragePrice` config type to File System pallet, which should implement the `UpdateStoragePrice` trait. This way, the File System pallet can be configured at will on how it updates the price for storage.
3. Adds the `MutatePricePerUnitPerTickInterface` trait to `primitives/traits`, which defines two functions: `get_price_per_unit_per_tick` and `set_price_per_unit_per_tick`.
4. Implements the `MutatePricePerUnitPerTickInterface` trait in the Payment Streams pallet, to be used by the File System pallet.
5. Implements the `on_poll` hook in the File System pallet, where it calls `update_storage_price` and sets that new price in the Payment Streams pallet.
6. Adds the `shp-data-price-updater` create which defines two implementations for `UpdateStoragePrice`:
    1. `NoUpdatePriceIndexUpdater`: Returns the same price. Useful for mock runtimes.
    2. `MostlyStablePriceIndexUpdater`: A highly configurable updater that keeps a price constant for the most part, but increases/decreases exponentially when the system utilisation is near 0% or 100%.
7. Configures the runtime to use `MostlyStablePriceIndexUpdater` with Pallet Param's parameters.
8. Adds `NumericalParam` trait that unifies many different traits that we normally use for numerical types configurable in trait inner types.
9. Fixes `get_stake` function in Storage Providers pallet to return MSP's stake and not held funds related to buckets.


pallets/file-system/src/lib.rs
```diff
@@ -1223,6 +1231,12 @@ pub mod pallet {
     where
         u32: TryFrom<BlockNumberFor<T>>,
     {
+        fn on_poll(_n: BlockNumberFor<T>, weight: &mut frame_support::weights::WeightMeter) {
+            // TODO: Benchmark computational weight cost of this hook.
```
undercover-cactus, line 1235: Should we create a task for this todo comment ?

ffarall: Good that you point it out. I believe we should have a task for benchmarking this pallet, generally speaking.

undercover-cactus (approved):

runtime/src/configs/runtime_params.rs
```diff
@@ -11,28 +11,84 @@ pub mod dynamic_params {
...
+        #[codec(index = 10)]
+        #[allow(non_upper_case_globals)]
+        /// l = [`LowerExponentFactor`]
+        /// system_utilisation = 0
+        ///
+        /// [`MinPrice`] = [`MostlyStablePrice`] - u * e ^ ( [`SystemUtilisationLowerThresholdPercentage`] - 0 )
+        ///
+        /// 9 = 48 - l * e ^ (0.3 - 0)
+        /// l = (48 - 9) * e ^ (-0.3) ≈ 29
+        pub static LowerExponentFactor: u32 = 12;
```
TDemeco, line 91: Why is this 12 if it should be 29 if we follow the same math as we did for the upper factor and in the docs for this factor?

TDemeco (approved):
Incredible job! The flexibility of havinga dedicated pallet-like primitive to handle price updates is definitely something we will cherish having in the future 💯 

The only thing that I don't fully like is the discontinuity of the price in the currently in use `MostlyStablePriceIndexUpdater`. I believe this can cause issues in the inflection points (30% and 95%) considering the immediate big changes in price getting to these means (from 48 to 19 picounits and from 48 to 459 picounits respectively).

At 30% issues might arise since users will want to pay as little as possible for the storage they are using, and they have the benefit of being able to change the used capacity of the network immediately by issuing new storage requests and issuing file deletions, contrary to BSPs that have a lock period before being able to change their provided capacity.
BSPs would not be incentivized to sign in or increase their capacity once near 30%, which is what we expect from the system, but users could temporary store files to try to trick BSPs into signing into the network or increasing their capacity, and once system capacity grows they could delete those files (with little to no penalty) and pay the considerably reduced rate of storage once system utilization drops under 30%. This works because they don't have to move the needle towards 0% utilization to get a favorable price (which wouldn't be in their interest since they actually want to use the system, and 30% could be enough for that purpose).

At 95% percent issues might arise because BSPs will want to make as much money as possible for the storage they are providing. If utilization is high enough (for example, 80%) but remains stable for long enough (balance between new storage requests and file deletions meaning no new net capacity is being used), the target of 95% is something the BSPs actually want to achieve, and could collectively decrease their total provided capacity to get to that point. Once they are there, they are earning almost 10 times as much while still leaving room (the extra 5%) for new storage requests, not halting the system. Sure, users can start to stop storing files to not pay the huge price, but the BSPs can wait for their capacity change locking period and reduce it again, since they are making a bigger profit than before as long as at least 10% of the files remain in the system compared to the initial state of the system.

I think both of those issues come from the fact that the price equation is not a continuum and most of the price reduction/increase comes into play in the thresholds (30% and 95%) instead of exponentially in the ranges (from 30% to 0% and from 95% to 100%).

I would love your (and the whole team's) opinion on the matter as I might be being too cautious with these remarks or I might even have gotten something completely wrong and the points I raised don't make sense 😬 

I also haven't come up with a better solution that avoids these issues and I believe both that the current solution is better than a fixed price, and that the code infrastructure you developed is superb and lets us easily update or add new price equations as needed and we could fix any issues later down the line, so I'm approving the PR 👍🏻

runtime/src/configs/runtime_params.rs
ffarall: You're absolutely correct, I must have missed it in one of the successive changes I did.

ffarall:
@TDemeco first of all thanks for the lengthy and on point review 👌🏼. I can tell you really payed attention to detail.

I've made an embarrassing mistake. The price equation is meant to be a CONTINUOUS function as you suggested, that exponentially increases after the `UpperThreshold`, and exponentially decreases below the `LowerThreshold`. It should have its minimum at `system_utilisation = 0` and its maximum at `system_utilisation = 1`. 

Here's the corrected formula:

 ```rust
if system_utilisation < LowerThreshold:
    price = stable_price - (e ^ (LowerThreshold - system_utilisation) - 1) * LowerExponentFactor
    price = max(price, MinPrice)
else if system_utilisation > UpperThreshold:
    price = stable_price + (e ^ (system_utilisation - UpperThreshold) - 1) * UpperExponentFactor
    price = min(price, MaxPrice)
else:
    price = stable_price
```

So,
- When `system_utilisation = LowerThreshold`: 
```
=> stable_price - (e ^ (LowerThreshold - LowerThreshold) - 1) * LowerExponentFactor = stable_price
``` 
- When `system_utilisation = UpperThreshold`:
```
stable_price + (e ^ (UpperThreshold - UpperThreshold) - 1) * UpperExponentFactor = stable_price
```

TDemeco (approved):

PR #226, opened 2024-10-11
feat: ✨ Add BSP lockup period

ffarall:
This PR:
1. Adds the `BspSignUpLockPeriod` configuration to Providers pallet.
2. Prevents BSPs from signing off before `BspSignUpLockPeriod` blocks have passed since their signup.


pallets/providers/src/utils.rs
```diff
@@ -449,6 +449,13 @@ where
             Error::<T>::StorageStillInUse
         );
 
+        // Check that the sign off period since the BSP signed up has passed
+        ensure!(
+            frame_system::Pallet::<T>::block_number()
+                >= bsp.sign_up_block + T::BspSignUpLockPeriod::get(),
+            Error::<T>::SignOffPeriodNotPassed
+        );
+
```
TDemeco, line 458: The lock for signing off is only checked for BSPs but the sign up block has been added to MSPs as well. Since we have this check for the former, shouldn't we have something similar for the latter?

pallets/providers/src/utils.rs
```diff
@@ -841,6 +848,7 @@ impl<T: Config> From<MainStorageProvider<T>> for BackupStorageProvider<T> {
             owner_account: msp.owner_account,
             payment_account: msp.payment_account,
             reputation_weight: T::StartingReputationWeight::get(),
+            sign_up_block: frame_system::Pallet::<T>::block_number(),
```
TDemeco, line 851: Continuing with having the `sign_up_block` for MSPs as well, this should be `msp.sign_up_block` instead of the current block.

runtime/src/configs/mod.rs
```diff
@@ -444,6 +444,7 @@ impl pallet_randomness::Config for Runtime {
 parameter_types! {
     pub const SpMinDeposit: Balance = 100 * UNIT;
     pub const BucketDeposit: Balance = 100 * UNIT;
+    pub const BspSignUpLockPeriod: BlockNumber = 90 * DAYS; // ~3 months
```
TDemeco, line 447: Maybe 3 months is a bit too much? Not opposed to it though

TDemeco (approved):
Left some comments for minor things but otherwise looks good 👍🏻

pallets/providers/src/utils.rs
ffarall: The reason for this lock is so that it is more expensive for an attacker to sign up a bunch of BSPs. There is no point in suddenly signing up a bunch of MSPs, and this kind of measure, increases the barrier of entry. Now you have to consider wether you're willing to lock these tokens for this amount of time, to venture into providing storage.

So that's why I didn't do the same for MSPs. I still record their signup to be more coherent and because it is somewhat useful information anyway.

pallets/providers/src/utils.rs
ffarall: You're right on that one. Originally I wasn't going to add `sign_up_block` to MSPs at all, but then I added it, so yes.

runtime/src/configs/mod.rs
ffarall: The auditor had suggested a year...

PR #232, opened 2024-10-18
feat: ✨ Enable benchmarking project-wide and benchmark `challenge` extrinsic

ffarall:


pallets/proofs-dealer/src/weights.rs
```diff
@@ -0,0 +1,76 @@
...
+#![cfg_attr(rustfmt, rustfmt_skip)]
+#![allow(unused_parens)]
+#![allow(unused_imports)]
+#![allow(missing_docs)]
+
+use frame_support::{traits::Get, weights::{Weight, constants::RocksDbWeight}};
+use core::marker::PhantomData;
+
+/// Weight functions needed for `pallet_proofs_dealer`.
+pub trait WeightInfo {
```
snowmead, line 39: I think you wrote this by hand if I am not mistaken, if that's the case, should we instead define the trait in the lib.rs of the pallet so we don't always have to write this trait after each generation?

snowmead (approved):

PR #234, opened 2024-10-22
feat: msp chargeable tick is current tick

undercover-cactus:
Last chargeable tick is set by default to be the current tick.

In case of dynamic payment we get the last chargeable info and get the last chargeable tick for it.


pallets/payment-streams/src/lib.rs
```diff
@@ -215,6 +215,10 @@ pub mod pallet {
     #[pallet::storage]
     pub type AccumulatedPriceIndex<T: Config> = StorageValue<_, BalanceOf<T>, ValueQuery>;
 
+    // List of priviledge providers
```
ffarall, line 218: Using `///` for rust docs makes it easier then to hover over the type where it's used, and see what it is about.

```suggestion
    /// Mapping of Privileged Providers.
    ///
    /// Privileged Providers are those who are allowed to charge up to the current tick in
    /// fixed rate payment streams, regardless of their [`LastChargeableInfo`].
```

ffarall (changes requested):

snowmead (approved):
lgtm modulo @ffarall comment about the tests

ffarall (approved):

PR #235, opened 2024-10-22
refactor: storage request flow and MSP batch accept

links234:
In this PR:
- MSP is no longer required for a bucket
- Remove unused/obsolete DataServer extrinsics
- Change the MSP batch accept files to be all-or-nothing
- Update the storage request flow events such that:
  - An expired storage request now can be either Rejected if the MSP does not respond or Expired (but still accepted, as before)
  - When the MSP rejects the storage request, an event is emitted immediately


test/suites/integration/msp/respond-multi-requests.test.ts
```diff
@@ -103,81 +102,6 @@ describeMspNet(
...
-
-      strictEqual(response2.newBucketRoot.toString(), local_bucket_root2.toString());
-
-      for (const fileKey of response2.fileKeys) {
-        const isFileInForest = await mspApi.rpc.storagehubclient.isFileInForest(
-          response2.bucketId.toString(),
-          fileKey
-        );
-        invariant(isFileInForest.isTrue, "File is not in forest");
-      }
```
snowmead, line 180: why is this all deleted? shouldn't we refactor the test?

snowmead (approved):
overall lgtm

pallets/file-system/src/lib.rs
```diff
@@ -528,10 +517,16 @@ pub mod pallet {
...
-        },
+        /// Notifies that a Main Storage Provider (MSP) has accepted a storage request for a specific file key.
+        ///
+        /// This event is emitted when an MSP agrees to store a file, but the storage request
+        /// is not yet fully fulfilled (i.e., the required number of Backup Storage Providers
+        /// have not yet confirmed storage).
+        ///
+        /// # Note
+        /// This event is not emitted when the storage request is immediately fulfilled upon
+        /// MSP acceptance. In such cases, a [`StorageRequestFulfilled`] event is emitted instead.
```
ffarall, line 528: Why not emitting both?

pallets/file-system/src/lib.rs
```diff
@@ -550,12 +545,26 @@ pub mod pallet {
...
         StorageRequestRevoked { file_key: MerkleHash<T> },
-        /// Notifies that a BSP has opened a request to stop storing a file.
+        /// Notifies that a storage request has either been directly rejected by the MSP or
+        /// the MSP did not respond to the storage request in time.
+        /// Note: There might be BSPs that have volunteered and confirmed the file already, for
+        /// which a priority challenge to delete the file will be issued.
+        StorageRequestRejected {
+            file_key: MerkleHash<T>,
+            reason: RejectedStorageRequestReason,
+        },
```
ffarall, line 567: Great job at clearly documenting this.

pallets/file-system/src/lib.rs
```diff
@@ -928,7 +932,7 @@ pub mod pallet {
             location: FileLocation<T>,
             fingerprint: Fingerprint<T>,
             size: StorageData<T>,
-            msp_id: ProviderIdFor<T>,
+            msp_id: Option<ProviderIdFor<T>>,
```
ffarall, line 935: You know, this makes me think that we don't really need this MSP argument here, having already passed the bucket. It's not like you can choose a different MSP than the one storing the bucket.

Worth setting up a task with @santikaplan to remove this argument.

pallets/file-system/src/utils.rs
```diff
@@ -691,8 +658,8 @@ where
     /// caller to verify the final result and apply only the file keys that have been successfully accepted.
     pub(crate) fn do_msp_respond_storage_request(
```
ffarall, line 659: Really like how simple this turned out after this refactor.

pallets/file-system/src/utils.rs
```diff
@@ -721,143 +683,56 @@ where
...
     ///
     /// The implementation follows this sequence:
     /// 1. Verify the non-inclusion proof.
-    /// 2. For each file key: Record a successful acceptance or a failure. Any failed operation while processing a file key
-    /// will not result in the function failing, but the file key will be marked as failed and the function will continue processing the rest.
-    /// 3. Apply the delta with all the keys that were successfully accepted to the root of the bucket.
+    /// 2. For each file key: Verify and process the acceptance. If any operation fails during the processing of a file key,
+    /// the entire function will fail and no changes will be applied.
+    /// 3. If all file keys are successfully processed, apply the delta with all the accepted keys to the root of the bucket.
+    /// 4. If any step fails, the function will return an error and no changes will be made to the storage state.
```
ffarall, line 726: There's a caveat to this sentence. Actually, no changes will be applied to storage, so long as this function is called within an extrinsic that propagates the error and makes the extrinsic fail. Only if the overarching extrinsic fails, then the storage is not updated.

pallets/file-system/src/utils.rs
```diff
@@ -2245,13 +2074,14 @@ where
 }
 
 mod hooks {
+    use crate::types::RejectedStorageRequestReason;
```
ffarall, line 2077: Merge this import

pallets/file-system/src/types.rs
```diff
@@ -134,90 +117,74 @@ impl<T: Config> Debug for MspStorageRequestResponse<T> {
...
-pub struct MspFailedBatchStorageRequests<T: Config> {
-    pub file_keys: BoundedVec<(MerkleHash<T>, DispatchError), MaxBatchMspRespondStorageRequests<T>>,
-    pub bucket_id: BucketIdFor<T>,
-    pub owner: T::AccountId,
-}
+/// Input for MSPs to respond to storage request(s).
+///
+/// The input is a list of ([BucketIdFor], [MspStorageRequestResponse]) elements,
+/// where the [MspStorageRequestResponse] contains the file keys that are accepted
+/// or rejected by the MSP.
```
ffarall, line 185: I think this comment is outdated.

test/util/bspNet/fileHelpers.ts
```diff
@@ -13,16 +13,19 @@ export const sendNewStorageRequest = async (
   source: string,
   location: string,
   bucketName: string,
-  valuePropId?: HexString,
-  mspId?: HexString,
+  valuePropId?: HexString | null,
+  mspId?: HexString | null,
```
ffarall, line 17: These parameters are optional (can be `undefined`) but you also make it possible to send `null`. It's a bit confusing.

test/suites/integration/msp/reject-request.test.ts
```diff
@@ -76,30 +74,6 @@ describeMspNet(
...
-      // Check that the MSP has not updated the local forest root of the bucket
-      strictEqual(
-        local_bucket_root.toString(),
-        (await mspApi.rpc.storagehubclient.getForestRoot(response.bucketId.toString())).toString()
-      );
-
-      strictEqual(response.bucketId.toString(), bucketId.toString());
-
-      strictEqual(response.fileKeys[0][0].toString(), newStorageRequestDataBlob.fileKey.toString());
-      strictEqual(response.fileKeys[0][1].toString(), "FileKeyAlreadyStored");
```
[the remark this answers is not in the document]
ffarall: I don't fully follow why we should delete these checks.

ffarall (approved):
Honestly, remarkable work with a refactor that didn't seem that complicated at first, but escalated quickly. Very sensible choices here.

Aside from the minor comments I left, I'm curious about the chunks of test checks that were removed from the MSP integration suite.

pallets/file-system/src/lib.rs
links234: I agree. I also think it's worth merging the MSP param (`ProviderIdFor<T>`) with the value proposition id in a new struct so that where we use them both, we only have 1 `Option` instead of 2.

ffarall: Well, removing the MSP param would just leave the value prop param, so there you go.

pallets/file-system/src/lib.rs
links234: Only to reduce the onchain size, since it's not adding any additional information in that case (fulfilled request is a superset).

test/util/bspNet/fileHelpers.ts
links234: We are using `undefined` in many places as "this has not been provided, I will retrieve a default value (first value prop of an MSP)" while in this case `null` is an actual "value" and I wanted to keep the same behaviour.

pallets/file-system/src/lib.rs
ffarall: I mean, it's not that big of an optimisation, but I get it 👌🏼

test/util/bspNet/fileHelpers.ts
ffarall: Fair point

test/suites/integration/msp/reject-request.test.ts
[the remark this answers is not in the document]
links234: Rewrote the tests :)

PR #236, opened 2024-10-22
test: Add a second MSP to fullnet tests (describeMspNet)

pLabarta:
This PR adds a second MSP node to the fullnet testrunner. It also adds node keys and updates constants related to both the first MSP and the new one.

Extraction of common components used by both BspNet and MspNet testrunners will be included in a following PR.


docker/local-dev-full-compose.yml
```diff
@@ -42,7 +42,35 @@ services:
...
         "--max-storage-capacity=4294967295",
         "--jump-capacity=1073741824",
-        "--name=sh-msp",
+        "--name=sh-msp-1",
+        "--no-hardware-benchmarks",
+        "--unsafe-rpc-external",
+        "--rpc-methods=unsafe",
+        "--port=30350",
+        "--rpc-cors=all",
+        "--node-key=${NODE_KEY}",
```
ffarall, line 51: Does this mean that they will both have the same `NODE_KEY`? Shouldn't they have different ones?

docker/local-dev-full-rocksdb-compose.yml
```diff
@@ -46,7 +46,37 @@ services:
...
         "--storage-layer=rocks-db",
         "--storage-path=/tmp/msp/${MSP_ID:-default_msp_id}",
-        "--name=sh-msp",
+        "--name=sh-msp-1",
+        "--no-hardware-benchmarks",
+        "--unsafe-rpc-external",
+        "--rpc-methods=unsafe",
+        "--port=30350",
+        "--rpc-cors=all",
+        "--node-key=${NODE_KEY}",
```
ffarall, line 55: Same here

ffarall (approved):
Approved with minor comment. Good job!

docker/local-dev-full-compose.yml
pLabarta: No, on `test/util/fullNet/helpers.ts`, when containers are started, these are passed their respective `NODE_KEY`. Like so:

```
await compose.upOne("sh-msp-1", {
      cwd: cwd,
      configAsString: updatedCompose,
      log: true,
      env: {
        ...process.env,
        NODE_KEY: ShConsts.NODE_INFOS.msp1.nodeKey,
        BSP_IP: bspIp,
        BSP_PEER_ID: bspPeerId,
        MSP_ID: ShConsts.DUMMY_MSP_ID
      }
    });
```

PR #238, opened 2024-10-29
test: added BSP delete file test

undercover-cactus:
This PR test the process for a BSP to request to stop storing a file. The previously needed `generateForestProof` function was missing but is now available.


test/util/bspNet/waits.ts
```diff
@@ -128,13 +135,34 @@ export const waitForBspStored = async (api: ApiPromise, checkQuantity?: number)
...
+
+      if ( bspId ) {
+        try {
+          // In cases the bsp is submitting a proof at the same time is trying to confirmStoring
+          await assertExtrinsicPresent(api, {
+            module: "proofsDealer",
+            method: "submitProof",
+            checkTxPool: true,
+            timeout: 100
+          });
```
undercover-cactus, line 154: This is working but not what we discussed. The `assertExtrinsicPresent` function doesn't return the actual extrinsic so I cannot check the bspId (also not the actual ID but the adddress).

I could add an optional address value to `assertExtrinsicPresent` to verify it is indeed the one that submitted the proof but it would be pretty specific to `submitProof` method.

For now this is working just to check when we have one BSP trying to store.

test/suites/integration/bsp/submit-proofs.test.ts
```diff
@@ -5,13 +5,14 @@ import {
...
+  ShConsts,
+  bspThreeKey
 } from "../../../util";
-import { BSP_THREE_ID, BSP_TWO_ID, DUMMY_BSP_ID } from "../../../util/bspNet/consts";
+import { BSP_THREE_ID, BSP_TWO_ID, DUMMY_BSP_ID, NODE_INFOS } from "../../../util/bspNet/consts";
 
 describeBspNet(
   "BSP: Many BSPs Submit Proofs",
-  { initialised: "multi", networkConfig: "standard" },
+  { initialised: "multi", networkConfig: "standard", only: true },
```
TDemeco, line 15: This `only: true` shouldn't be here, right?

test/suites/integration/bsp/submit-proofs.test.ts
```diff
@@ -164,28 +165,31 @@ describeBspNet(
...
+      await userApi.sealBlock(
+        bspThreeApi.tx.fileSystem.bspRequestStopStoring(
+          fileData.fileKey,
+          fileData.bucketId,
+          fileData.location,
+          fileData.owner,
+          fileData.fingerprint,
+          fileData.fileSize,
+          false,
+          inclusionForestProof.toString()
```
TDemeco, line 183: I didn't know sending the proof as a string works!

test/util/bspNet/waits.ts
```diff
@@ -112,10 +112,19 @@ export const waitForBspVolunteerWithoutSealing = async (
...
+export const waitForBspStored = async (api: ApiPromise, checkQuantity?: number, bspId?: string) => {
+  // TODO: add bsp ID to check if is not confirming storing
   // To allow time for local file transfer to complete (10s)
   const iterations = 100;
   const delay = 100;
+
+  if (bspId && checkQuantity && checkQuantity > 1) {
+    // We do this because `bspConfirmStoring` cannot happened in the same block in which a BSP submit a proof.
+    throw new Error(
+      "Cannot use `waitForBspStored` with a bspId and an amount of extrinsec bigger than 1."
```
TDemeco, line 124:
```suggestion
    // We do this because a BSP cannot call `bspConfirmStoring` in the same block in which it has to submit a proof, since it can only send one root-changing transaction per block and proof submission is prioritized.
    throw new Error(
      "Invalid parameters: `waitForBspStored` cannot be used with an amount of extrinsics to wait for bigger than 1 if a BSP ID was specified."
```

TDemeco (changes requested):
Looks good! I requested some minor changes

test/suites/integration/bsp/submit-proofs.test.ts
```diff
@@ -355,8 +359,14 @@ describeBspNet(
...
-      // Finally, advance to the next challenge tick.
-      await userApi.advanceToBlock(nextChallengeTick);
+
+      const currentBlock = await userApi.rpc.chain.getBlock();
+      const currentBlockNumber = currentBlock.block.header.number.toNumber();
+
+      if (nextChallengeTick < currentBlockNumber) {
+        // Advance to the next challenge tick if needed
+        await userApi.advanceToBlock(nextChallengeTick);
+      }
```
ffarall, line 369: This should be the other way around. We should advance if `nextChallengeTick > currentBlockNumber`.

test/util/bspNet/waits.ts
```diff
@@ -112,10 +112,19 @@ export const waitForBspVolunteerWithoutSealing = async (
  *
  * @throws Will throw an error if the expected extrinsic or event is not found.
  */
-export const waitForBspStored = async (api: ApiPromise, checkQuantity?: number) => {
+export const waitForBspStored = async (api: ApiPromise, checkQuantity?: number, bspId?: string) => {
+  // TODO: add bsp ID to check if is not confirming storing
```
ffarall, line 116: This was left dangling right?

test/util/bspNet/waits.ts
```diff
@@ -128,13 +137,33 @@ export const waitForBspStored = async (api: ApiPromise, checkQuantity?: number)
...
+        try {
+          // In cases the bsp is submitting a proof at the same time is trying to confirmStoring
+          await assertExtrinsicPresent(api, {
+            module: "proofsDealer",
+            method: "submitProof",
+            checkTxPool: true,
+            timeout: 100
+          });
+
+          // if we have found one we go one block forward
```
ffarall, line 157:
```suggestion
          // If we have found one we go one block forward
```

test/util/bspNet/waits.ts
ffarall: And actually, it should be the BSP Account ID (the address)

test/util/bspNet/waits.ts
ffarall: @TDemeco the BSP ID in the transaction will likely be `None`, because the BSP is sending it with it's address that is linked to his BSP ID. So she would have to get the address that is the signer of the extrinsic, and compare that.

test/util/bspNet/waits.ts
```diff
@@ -128,13 +137,33 @@ export const waitForBspStored = async (api: ApiPromise, checkQuantity?: number)
...
+          const { events } = await sealBlock(api);
+          assertEventPresent(api, "fileSystem", "BspConfirmedStoring", events);
+
+          continue;
+        } catch {
+          // Nothing here
```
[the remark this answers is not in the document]
ffarall: Agree, but I would add that to the log message printed at the end of the attempts, below. Other wise you'd be printing that error at every polling iteration.

ffarall (changes requested):

test/suites/integration/bsp/submit-proofs.test.ts
```diff
@@ -168,28 +169,31 @@ describeBspNet(
...
-        //     fileMetadata.owner,
-        //     fileMetadata.fingerprint,
-        //     .fileSize,
-        //     false
-        //   ),
-        //   bspThreeKey
-        // );
-      }
-    );
+    it("BSP stops storing last file", async () => {
```
timbrinded, line 172: This test could benefit from better assertions. Currently it only checks that a blockchain event has been fired, but we don't actually know what the effects on the physical node are. 

I don't know how much of the delete feature has been impl but if relevant - consider adding checks to:
- Use an RPC call to verify the file can not longer be found
- Check that an expecting logging line is

test/util/bspNet/waits.ts
```diff
@@ -112,10 +112,19 @@ export const waitForBspVolunteerWithoutSealing = async (
  *
  * @throws Will throw an error if the expected extrinsic or event is not found.
  */
-export const waitForBspStored = async (api: ApiPromise, checkQuantity?: number) => {
+export const waitForBspStored = async (api: ApiPromise, checkQuantity?: number, bspId?: string) => {
+  // TODO: add bsp ID to check if is not confirming storing
   // To allow time for local file transfer to complete (10s)
   const iterations = 100;
   const delay = 200;
+
```
timbrinded, line 120: use `invariant(<condition>,<error message>)` as not only is it neater but also provides type-narrowing on subsequent code using vars in the condition

test/util/bspNet/waits.ts
undercover-cactus: I changed it for `bspAccount`

test/suites/integration/bsp/submit-proofs.test.ts
undercover-cactus: There is more tests that are being skipped that are executed after this one. I believe we should indeed re-activate them to make sure the full scenario works as expected. This could be in a new PR.

test/suites/integration/bsp/submit-proofs.test.ts
```diff
@@ -5,14 +5,15 @@ import {
...
+  bspThreeKey
 } from "../../../util";
-import { BSP_THREE_ID, BSP_TWO_ID, DUMMY_BSP_ID } from "../../../util/bspNet/consts";
+import { BSP_THREE_ID, BSP_TWO_ID, DUMMY_BSP_ID, NODE_INFOS } from "../../../util/bspNet/consts";
 import invariant from "tiny-invariant";
 
 describeBspNet(
   "BSP: Many BSPs Submit Proofs",
-  { initialised: "multi", networkConfig: "standard" },
+  { initialised: "multi", networkConfig: "standard", only: true },
```
TDemeco: This `only: true` is still here. Should it?

test/suites/integration/bsp/submit-proofs.test.ts
TDemeco, line 172: I would rename this tests to something like "BSP requests to stop storing the last file", since we're not actually checking anything else

test/suites/integration/bsp/submit-proofs.test.ts
```diff
@@ -359,8 +363,14 @@ describeBspNet(
...
-      // Finally, advance to the next challenge tick.
-      await userApi.advanceToBlock(nextChallengeTick);
+
+      const currentBlock = await userApi.rpc.chain.getBlock();
+      const currentBlockNumber = currentBlock.block.header.number.toNumber();
+
+      if (nextChallengeTick < currentBlockNumber) {
+        // Advance to the next challenge tick if needed
+        await userApi.advanceToBlock(nextChallengeTick);
+      }
```
TDemeco, line 373: This is still the other way around, right? The condition to advance to the next challenge tick should be `nextChallengeTick > currentBlockNumber`

test/util/bspNet/waits.ts
```diff
@@ -112,13 +112,41 @@ export const waitForBspVolunteerWithoutSealing = async (
...
+    throw new Error(
+      "Invalid parameters: `waitForBspStored` cannot be used with an amount of extrinsics to wait for bigger than 1 if a BSP ID was specified."
+    );
+  }
+
   for (let i = 0; i < iterations + 1; i++) {
     try {
       await sleep(delay);
+
+      // check if we have a submitProo extrinsic
```
TDemeco, line 136:
```suggestion
      // check if we have a submitProof extrinsic
```

TDemeco (changes requested):

test/suites/integration/bsp/submit-proofs.test.ts
undercover-cactus: yep... I am 100% sure I had fix it.

test/util/bspNet/test-api.ts
```diff
@@ -258,7 +258,8 @@ export class BspNetTestApi implements AsyncDisposable {
        * @param expectedExts - Optional param to specify the number of expected extrinsics.
        * @returns A promise that resolves when a BSP has confirmed storing a file.
        */
-      bspStored: (expectedExts?: number) => Waits.waitForBspStored(this._api, expectedExts),
+      bspStored: (expectedExts?: number, bspAccount?: string) =>
```
ffarall, line 261: I'd suggest using the `Address` type from `@polkadot/types/interfaces` here, instead of a plain `string`.

test/util/bspNet/waits.ts
```diff
@@ -112,13 +112,41 @@ export const waitForBspVolunteerWithoutSealing = async (
...
  * @throws Will throw an error if the expected extrinsic or event is not found.
  */
-export const waitForBspStored = async (api: ApiPromise, checkQuantity?: number) => {
+export const waitForBspStored = async (
+  api: ApiPromise,
+  checkQuantity?: number,
+  bspAccount?: string
+) => {
+  // TODO: add bsp ID to check if is not confirming storing
   // To allow time for local file transfer to complete (10s)
```
ffarall, line 121: Remove leftover `TODO`.

ffarall (approved):

TDemeco (approved):

PR #241, opened 2024-10-29
feat: :art: refactor data servers functionality to only use dedicated storage

TDemeco:
~This PR changes the naming of `MaxBspsPerStorageRequest` to `MaxDataServers`, corrects some docs that were not precise and also makes it so when a BSP confirms a file, it gets added to the data servers of that storage request (which could prove useful if a BSP wants to confirm storing a file but the user is not currently available to send it).~

This PR refactors the data servers functionality that was in storage requests' metadata. Instead of having it there AND in its own dedicated storage element (`StorageRequestBsps`), which served no redundancy purpose, now we only have them in the storage element, simplifying access to it.


pallets/file-system/src/types.rs
```diff
@@ -57,11 +57,11 @@ pub struct StorageRequestMetadata<T: Config> {
...
     pub user_peer_ids: PeerIds<T>,
 
-    /// List of storage providers that can serve the data that is requested to be stored.
+    /// List of backup storage providers that can act as a data server for a storage request, serving the data that is requested to be stored.
     ///
     /// This is useful when a BSP stops serving data and automatically creates a new storage request with no user multiaddresses, since
-    /// SPs can prove and serve the data to be replicated to other BSPs without the user having this stored on their local machine.
-    pub data_server_sps: BoundedVec<ProviderIdFor<T>, MaxBspsPerStorageRequest<T>>, // TODO: Change the Maximum data servers to be the maximum SPs allowed
+    /// BSPs can prove and serve the data to be replicated to other BSPs without the user having this stored on their local machine.
+    pub data_server_sps: BoundedVec<ProviderIdFor<T>, MaxDataServers<T>>,
```
snowmead, line 64: I think this should be bounded by the `ReplicationTarget` to ensure all BSPs who confirmed storing this file can register themselves as a data server. The alternative solution would be to add an equality check for `MaxDataServers` and `ReplicationTarget` in the `integrity_test` if we wish to keep these types distinct. I would just use the former imo.

pallets/file-system/src/utils.rs
```diff
@@ -1556,6 +1564,10 @@ where
                             .bsps_volunteered
                             .saturating_sub(ReplicationTargetType::<T>::one());
 
+                        storage_request_metadata
+                            .data_server_sps
+                            .retain(|&x| x != bsp_id);
```
snowmead, line 1569: I think we should instead remove this field entirely from the metadata and rely solely on `StorageRequestBsps` which holds all the confirmed BSPs of a storage request. Nodes can query this storage to find any potential confirmed BSPs and request the data. It would reduce complexity, additional storage and iteration as shown here. This was a premature addition to the metadata. 

https://github.com/Moonsong-Labs/storage-hub/blob/7c42d80fa25ac6e381b2a36f8597b96e75273571/pallets/file-system/src/lib.rs#L295-L303

snowmead (changes requested):
I think we should remove `data_server_sps` in favor of querying `StorageRequestBsps` for confirmed BSPs. It's redundant imo and I had added this field to the metadata prematurely.

pallets/file-system/src/utils.rs
TDemeco: I agree with this and your proposed solution, as it's something I thought about when implementing the changes ("why do we have a variable in the request metadata for data servers when we have a separate storage for them"), so if @ffarall agrees I'll do the changes in this PR

ffarall: I agree with both.

ffarall (approved):
Approving and subscribing to @snowmead's comment

links234 (approved):
I agree with @snowmead & @ffarall. I also think we should remove everything related to `data_server`s.

I was going to do it on this PR: https://github.com/Moonsong-Labs/storage-hub/pull/235. There I remove the extrinsic and tests related to it.

TDemeco:
> I agree with @snowmead & @ffarall. I also think we should remove everything related to `data_server`s.
> 
> I was going to do it on this PR: #235. There I remove the extrinsic and tests related to it.

Nice work there! I saw that your changes target mainly the data servers used to move buckets from one MSP to another, so it shouldn't conflict with the ones in this PR, right?

snowmead (approved):

PR #244, opened 2024-10-31
feat: :sparkles: add extrinsic to allow users to delete empty buckets

TDemeco:
This PR adds an extrinsic that allows the user owner of a bucket to delete it and recover its deposit if the bucket is empty.


santikaplan (approved):
LGTM! 👍

snowmead (approved):
Nice!

PR #245, opened 2024-10-31
feat: ✨ Proofs generator testing script and RPC

ffarall:
This PR:
1. Adds `generateProof` RPC method to generate a full proof (forest and keys) for a given `provider_id`, `seed` and `challenged_keys`.
2. Exposes runtime APIs to all RPC methods in `storagehubclient`.
3. Adds testing scenario meant specifically for generating proofs and challenges, and writes them to a template rust file.
4. Organises some dependencies alphabetically.


client/rpc/src/lib.rs
```diff
@@ -417,6 +446,86 @@ where
...
+                        proven_keys.push(right.key);
+                    }
+                    (Some(left), None) => proven_keys.push(left.key),
+                    (None, Some(right)) => proven_keys.push(right.key),
+                    (None, None) => {
+                        return Err(into_rpc_error("Both left and right leaves in forest proof are None. This should not be possible."));
+                    }
+                },
+                Proven::Empty => {
+                    return Err(into_rpc_error("Forest proof generated with empty forest. This should not be possible, as this provider shouldn't have been challenged with an empty forest."));
```
TDemeco, line 498: This is an RPC method, right? So it might be called without the need of a challenge which means the error message might not be correct. I'd leave it as "Tried to generate forest proof with an empty forest" or something along those lines

client/rpc/src/lib.rs
```diff
@@ -1,25 +1,26 @@
...
-use sp_keystore::{Keystore, KeystorePtr};
-use sp_runtime::{AccountId32, Deserialize, KeyTypeId, Serialize};
-
 use shc_file_manager::traits::{FileDataTrie, FileStorage, FileStorageError};
-
-use log::{debug, error};
-use tokio::fs;
-
-use std::{fmt::Debug, fs::File, io::Read, io::Write, path::PathBuf, sync::Arc};
-use tokio::{fs::create_dir_all, sync::RwLock};
```
TDemeco, line 22: Nice, this was a bit messy

TDemeco (approved):
Great work!! 💯

client/rpc/src/lib.rs
ffarall: Yeah, I agree. I'll change that in the next PR to speed up this one getting merged.

PR #246, opened 2024-11-01
fix: 🩹 correctly set last valid block for epoch randomness

TDemeco:
This PR changes the way we set the last valid block when updating the epoch randomness.

Previously, whenever we detected than an epoch changed happened in the relay chain, we got the randomness from the previous epoch and set it as valid for any commitments that were done on the previous block or any block before that. That could have caused issues since the previous block could have been anchored to the previous relay chain block, and the validator of that relay chain block would have been able to influence the obtained epoch randomness (further explained [here](https://github.com/paritytech/polkadot-sdk/blob/2700dbf2dda8b7f593447c939e1a26dacdb8ce45/substrate/frame/babe/src/randomness.rs#L82)).

Now, we make it so the randomness from the previous epoch is only valid for commitments done in parachain blocks that were anchored at least before the last relay chain block of the previous epoch. This way neither the last collator nor the last validator of the epoch can influence the randomness AND make commitments knowing what their randomness will be after their influence.


snowmead (approved):
lgtm!

ffarall (approved):
Great job. I went through the implementation of `RelaychainDataProvider` and everything checks out the way you described it.

PR #247, opened 2024-11-01
feat: Penalize BSP for stop storing a file and transfer to Treasury account

snowmead:
Penalize the BSP for stopping to store a file. The extrinsic will fail if they do not have enough funds to pay for the fee. The fee is then transfered to the treasury account.


pallets/file-system/src/utils.rs
```diff
@@ -1540,7 +1540,17 @@ where
             Error::<T>::NotABsp
         );
 
-        // TODO: charge SP for this action.
+        let bsp_account_id =
+            <T::Providers as shp_traits::ReadProvidersInterface>::get_owner_account(bsp_id)
+                .ok_or(Error::<T>::FailedToGetOwnerAccount)?;
```
TDemeco, line 1545: I would change the error for a `expect_or_err!` macro invocation since this should never error, as we are checking that the provider exists and is a BSP before, so it MUST have an owner account.

TDemeco (approved):
Looks good! Just a minor nitpick

runtime/src/configs/runtime_params.rs
```diff
@@ -110,6 +110,11 @@ pub mod dynamic_params {
         #[allow(non_upper_case_globals)]
         /// The maximum treasury cut that can be taken from the amount charged from a payment stream.
         pub static MaximumTreasuryCut: Perbill = Perbill::from_percent(5);
+
+        #[codec(index = 15)]
+        #[allow(non_upper_case_globals)]
+        /// The penalty a BSP must pay when they forcefully stop storing a file.
+        pub static BspStopStoringFilePenalty: Balance = 1 * UNIT;
```
ffarall, line 117: This should be much higher. Add a TODO to set it to something higher at least.

ffarall (approved):

PR #248, opened 2024-11-04
fix: Trigger NewStorageRequest event when BSP stops storing

undercover-cactus:
This PR add a test to verify that the `NewStorageRequest` event is being triggered and move the event `NewStorageRequest` inside the utility function to request storage to make sure it is being triggered when needed.


test/suites/integration/bsp/storage-delete.test.ts
```diff
@@ -0,0 +1,163 @@
...
+      );
+
+      // unpause bsp three
+      await userApi.docker.resumeBspContainer({ containerName: "sh-bsp-three" });
+
+      const inclusionForestProof = await bspApi.rpc.storagehubclient.generateForestProof(null, [
+        fileKey
+      ]);
+      await userApi.sealBlock(
+        bspApi.tx.fileSystem.bspRequestStopStoring(
```
ffarall, line 117: It's fine to use the `bspApi` here, but you don't necessarily have to.

test/suites/integration/bsp/storage-delete.test.ts
```diff
@@ -0,0 +1,163 @@
...
+      );
+
+      // when requested to stop storing a file we should also receive an event new storage request
+      // to replace the bsp leaving
+      userApi.assert.fetchEventData(
+        userApi.events.fileSystem.NewStorageRequest,
+        await userApi.query.system.events()
+      );
+
+      // wait for the irght moment to confirm stop storing
```
ffarall, line 143:
```suggestion
      // Wait for the right moment to confirm stop storing
```

ffarall:
Oh by the way, we should change the PR title to something a bit more descriptive. Something like `fix: 🐛 Trigger NewStorageRequest event when BSP stops storing`.

test/suites/integration/bsp/storage-delete.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    });
+
+    it("Network launches and can be queried", async () => {
+      const userNodePeerId = await userApi.rpc.system.localPeerId();
+      strictEqual(userNodePeerId.toString(), userApi.shConsts.NODE_INFOS.user.expectedPeerId);
+      const bspNodePeerId = await bspApi.rpc.system.localPeerId();
+      strictEqual(bspNodePeerId.toString(), userApi.shConsts.NODE_INFOS.bsp.expectedPeerId);
+    });
+
+    it("bsp one stop storing and bsp two volunteer", async () => {
```
ffarall, line 38: Isn't it BSP Three the one who volunteers to replace BSP One?

test/suites/integration/bsp/storage-delete.test.ts
```diff
@@ -0,0 +1,106 @@
...
+          userApi.shConsts.NODE_INFOS.user.AddressId,
+          fingerprint,
+          fileSize,
+          false,
+          inclusionForestProof.toString()
+        ),
+        bspKey
+      );
+
+      await sleep(500);
```
ffarall, line 79: There's no need to wait between a sealed block and checking if there's an event in that block. Specially if you're sealing the block and checking for the event with the same `userApi`. If you were sealing the block with `userApi` and checking, let's say, with `bspApi`, then yes, because you need to wait for the block to be gossipped.

test/suites/integration/bsp/storage-delete.test.ts
```diff
@@ -0,0 +1,106 @@
...
+        currentBlockNumber + userApi.consts.fileSystem.minWaitForStopStoring.toNumber();
+      await userApi.advanceToBlock(cooldown);
+
+      // Confirm stop storing
+      await userApi.sealBlock(
+        userApi.tx.fileSystem.bspConfirmStopStoring(fileKey, inclusionForestProof),
+        bspKey
+      );
+
+      await sleep(500);
```
ffarall, line 99: Also here, no need to wait.

test/suites/integration/bsp/storage-delete.test.ts
```diff
@@ -0,0 +1,106 @@
...
+        await userApi.file.createBucketAndSendNewStorageRequest(source, destination, bucketName);
+
+      // Wait for the two BSP to volunteer
+      await userApi.wait.bspVolunteer(2);
+      await userApi.wait.bspStored(2);
+
+      // Revoke the storage request otherwise the new storage request event is not being triggered
+      await userApi.sealBlock(userApi.tx.fileSystem.revokeStorageRequest(fileKey), shUser);
+
+      await sleep(500);
```
ffarall, line 56: No need to wait here too.

ffarall (approved):
The test is looking much cleaner! 👏🏼  I've left a couple of comments that hopefully help too. Anyway, I trust you'll take care of them.

And watch out for the failing test in CI too.

test/suites/integration/bsp/storage-delete.test.ts
```diff
@@ -0,0 +1,163 @@
...
+      // Wait for the two BSP to volunteer
+      await userApi.wait.bspVolunteer(2);
+      await userApi.wait.bspStored(2);
+
+      // Wait for the BSP to receive and store all files
+      const fileKey = fileKeys[0];
+      await bspApi.wait.bspFileStorageComplete(fileKey);
+
+      // Revoke the storage request otherwise the new storage request event is not being triggered
+      await userApi.sealBlock(userApi.tx.fileSystem.revokeStorageRequest(fileKey), shUser);
```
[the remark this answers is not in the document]
undercover-cactus: I tried to do that but I couldn't create the storage request with MSP `null`. I tried also to trigger the StorageRequestExperiration but I couldn't triggered it. I used the rpc api to know how long I was supposed to wait.

If you have a quick way to have MSP ´null´ I will modify it. I agree revoking the storage request is not great.

ffarall (approved):

PR #249, opened 2024-11-06
fix: 🩹 Execute `verifyContainerFreshness` in `before` block of each test suite

ffarall:
Before it was being executed at the beginning of the `describeBspNet` which caused unexpected errors when a test suite was executed with `only: true` and `keepAlive: true`. Now it is executed in the `before` block of each test suite meaning two things:
1. Test suites without `only: true` are filtered out and don't execute `verifyContainerFreshness` when running `pnpm test:bspnet:only`.
2. `verifyContainerFreshness` is run on every test suite, including same test suites but with different network configs (like memory DB and Rocks DB).

### DIFF
--- a/test/suites/integration/bsp/storage-capacity.test.ts
+++ b/test/suites/integration/bsp/storage-capacity.test.ts
@@ -1,12 +1,5 @@
 import assert from "node:assert";
-import {
-  bspKey,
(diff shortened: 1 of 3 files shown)


timbrinded (approved):
LGTM!

PR #255, opened 2024-11-12
feat: Slash provider with automatic/manual top up

snowmead:
Heavily modifies the `do_slash` implementation to support the following cases:
- slash what we can from the provider's held deposit
- attempt to automatically top up the provider's held deposit if there is a capacity deficit (i.e. the provider's capacity falls below their used capacity) if they have enough free balance

Providers can manually top up their deposit by calling the new `top_up_deposit` extrinsic which will attempt to hold the required deposit needed to cover the capacity deficit. This extrinsic doesn't execute if there is no capacity deficit.

Two more subsequent PRs will cover the cases listed below:
- a grace period will be allocated to a provider before they are slashed for them to provide a proof of the files they would have submitted for the challenges missed in order to reduce the slash amount. right now the slash amount is calculated on a worst case scenario basis.
- a hook will handle end grace period blocks for awaited providers who haven't topped up yet. this will mark the providers as inactive, signaling to the network that they issue redundancy requests for files stored under inactive providers.


pallets/providers/src/lib.rs
```diff
@@ -427,6 +439,34 @@ pub mod pallet {
...
+        ProviderIdFor<T>,
+        (),
+    >;
+
+    /// Storage providers currently awaited for to top up their deposit. This storage holds the current amount that the provider was
+    /// slashed for.
+    ///
+    /// This is primarily used to lookup providers, restrict certain operations while they are in this state.
+    #[pallet::storage]
+    pub type AwaitingTopUpFromProviders<T: Config> =
```
ffarall, line 467: This storage element is never cleared if the provider doesn't voluntarily top up. I'd suggest removing it altogether. I'd let Providers increment their stake immediately after they've been slashed, by setting their "change capacity periods" so that they can immediately do it now. I wouldn't worry about storing how much they have to top up to.

pallets/providers/src/lib.rs
```diff
@@ -271,6 +277,12 @@ pub mod pallet {
         /// fee for a bucket that doesn't have any files yet)
         #[pallet::constant]
         type ZeroSizeBucketFixedRate: Get<BalanceOf<Self>>;
+
+        /// Period of time for a provider to top up their deposit after being slashed.
+        ///
+        /// If the provider does not top up their deposit within this period, they will be marked as insolvent.
+        #[pallet::constant]
+        type TopUpGracePeriod: Get<u32>;
```
ffarall, line 285: This grace period is not enforced. I'd just remove it.

pallets/providers/src/types.rs
```diff
@@ -5,11 +5,24 @@ use codec::{Decode, Encode, MaxEncodedLen};
...
 use scale_info::TypeInfo;
 use sp_runtime::BoundedVec;
 
 pub type Multiaddresses<T> = BoundedVec<MultiAddress<T>, MaxMultiAddressAmount<T>>;
 
+pub type ValuePropId<T> = HashId<T>;
+
+/// Top up metadata for a provider tracked in storage.
+#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, RuntimeDebugNoBound, PartialEq, Eq, Clone)]
+pub struct TopUpMetadata {
```
ffarall, line 18: I'd remove this.

pallets/providers/src/utils.rs
```diff
@@ -662,16 +664,7 @@ where
...
-        let capacity_over_minimum = new_capacity
-            .checked_sub(&T::SpMinCapacity::get())
-            .ok_or(Error::<T>::StorageTooLow)?;
-        let deposit_for_capacity_over_minimum = T::DepositPerData::get()
-            .checked_mul(&capacity_over_minimum.into())
-            .ok_or(DispatchError::Arithmetic(ArithmeticError::Overflow))?;
-        let new_deposit = T::SpMinDeposit::get()
-            .checked_add(&deposit_for_capacity_over_minimum)
-            .ok_or(DispatchError::Arithmetic(ArithmeticError::Overflow))?;
+        let new_deposit = Self::compute_deposit_needed_for_capacity(new_capacity)?;
```
ffarall, line 667: There are 3 other places where this calculation is done. We'd want to unify that.

pallets/providers/src/utils.rs
```diff
@@ -992,7 +1195,9 @@ where
     /// being an exact match to a file key stored by the Storage Provider. The StorageHub protocol requires the Storage Provider to
     /// submit a proof of storage for the neighbouring file keys of the missing challenged file key.
     ///
-    /// The slashing amount is calculated based on an assumption that every file is the maximum size allowed by the protocol.
+    /// The slashing amount is calculated as the product of the [`SlashAmountPerMaxFileSize`](Config::SlashAmountPerMaxFileSize) (an assumption
+    /// that every file is the maximum size allowed by the protocol) and the accrued failed proof submissions multiplied by `2` to
```
ffarall, line 1199: I don't think we enforce a max file size. This just a "very large file size" that we consider. Meaning that most files should be below that.

pallets/providers/src/lib.rs
```diff
@@ -191,6 +194,9 @@ pub mod pallet {
             + Ord
             + Bounded;
 
+        /// Interface to get the relay chain block number which was used as an anchor for the last block in the parachain.
+        type RelayBlockGetter: BlockNumberProvider<BlockNumber = RelayChainBlockNumber>;
```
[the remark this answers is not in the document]
snowmead: so we can deterministically set the average time we allocate to a provider top up grace period. in the near future we will probably adjust our runtime to use coretime/elastic scaling. this grace period should not be using the parachains block number.

pallets/providers/src/lib.rs
```diff
@@ -427,6 +439,34 @@ pub mod pallet {
...
+    ///
+    /// The `on_pool` hook will process every grace period's slashed providers and attempt to top up their required deposit before
+    /// marking them as insolvent. If a provider is marked as insolvent, the network (e.g users, other providers) can issue
+    /// `add_redundancy` requests to replicate the data loss if it was a BSP. If it was an MSP, the user can decide to move their
+    /// buckets to another MSP or delete their buckets.
+    ///
+    /// The relay chain block is used to ensure we have a predictive way to determine how much time we allocate to the provider to
+    /// top up their deposit.
+    #[pallet::storage]
+    pub type GracePeriodToSlashedProviders<T: Config> = StorageDoubleMap<
```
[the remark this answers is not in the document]
snowmead: we discussed previously about splitting this task into 3 separate PRs. I precised in the description of the PR the subsequent PRs I will be working on which one of them includes executing in `on_poll` or `on_idle` the automatic clean up of the storage elements you mentioned in your review. 

>Two more subsequent PRs will cover the cases listed below:
>- a grace period will be allocated to a provider before they are slashed for them to provide a proof of the files they would have submitted for the challenges missed in order to reduce the slash amount. right now the slash amount is calculated on a worst case scenario basis.
>- a hook will handle end grace period blocks for awaited providers who haven't topped up yet. this will mark the providers as inactive, signaling to the network that they issue redundancy requests for files stored under inactive providers.

pallets/providers/src/lib.rs
```diff
@@ -427,6 +439,34 @@ pub mod pallet {
         OptionQuery,
     >;
 
+    /// Providers that have been slashed and are required to top up their deposit to the required amount given their current
+    /// max capacity.
```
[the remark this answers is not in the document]
snowmead: agreed, will update these docs so it is clear

pallets/providers/src/utils.rs
```diff
@@ -831,48 +824,258 @@ where
...
+        );
+        AwaitingTopUpFromProviders::<T>::remove(provider_id);
+
+        // Signal that the slashed amount has been topped up
+        Self::deposit_event(Event::<T>::TopUpFulfilled {
+            provider_id,
+            amount: required_delta_hold_amount,
         });
 
         Ok(Pays::No.into())
```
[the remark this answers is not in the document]
snowmead: agreed

pallets/providers/src/lib.rs
```diff
@@ -1265,6 +1323,20 @@ pub mod pallet {
 
             Self::do_slash(&provider_id)
         }
+
+        /// Dispatchable extrinsic to top-up the deposit of a Storage Provider.
+        ///
+        /// The dispatch origin for this call must be signed.
+        ///
+        /// This is a free transaction if the user successfully tops up their deposit.
```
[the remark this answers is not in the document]
snowmead: I think this operation should be free so we increase the probabiliy of providers being slashed.

pallets/providers/src/lib.rs
```diff
@@ -496,10 +536,24 @@ pub mod pallet {
...
 
-        /// Event emitted when an SP has been slashed.
+        /// Event emitted when a SP has been slashed.
         Slashed {
             provider_id: ProviderIdFor<T>,
-            amount_slashed: BalanceOf<T>,
+            amount: BalanceOf<T>,
+        },
+
+        /// Event emitted when a provider has been slashed optionally signaling the end of the grace
```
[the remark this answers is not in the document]
snowmead: https://www.merriam-webster.com/thesaurus/signaling

pallets/providers/src/lib.rs
```diff
@@ -427,6 +439,34 @@ pub mod pallet {
         OptionQuery,
     >;
 
+    /// Providers that have been slashed and are required to top up their deposit to the required amount given their current
+    /// max capacity.
+    ///
+    /// The `on_pool` hook will process every grace period's slashed providers and attempt to top up their required deposit before
+    /// marking them as insolvent. If a provider is marked as insolvent, the network (e.g users, other providers) can issue
+    /// `add_redundancy` requests to replicate the data loss if it was a BSP. If it was an MSP, the user can decide to move their
+    /// buckets to another MSP or delete their buckets.
```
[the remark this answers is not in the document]
snowmead: this was here initially because the scope of this pr would cover the hook implementation. 

I am inclined to leave this here for documentation sake when it comes time to implement it in a subsequent pr

pallets/providers/src/lib.rs
```diff
@@ -496,10 +536,24 @@ pub mod pallet {
...
-        /// Event emitted when an SP has been slashed.
+        /// Event emitted when a SP has been slashed.
         Slashed {
             provider_id: ProviderIdFor<T>,
-            amount_slashed: BalanceOf<T>,
+            amount: BalanceOf<T>,
+        },
+
+        /// Event emitted when a provider has been slashed optionally signaling the end of the grace
+        /// period if an automatic top up could not be performed.
```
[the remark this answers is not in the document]
snowmead: no it isn't thanks for spotting, will update

pallets/providers/src/lib.rs
snowmead: replied in comment above ^

pallets/providers/src/types.rs
```diff
@@ -5,11 +5,24 @@ use codec::{Decode, Encode, MaxEncodedLen};
...
+    /// The last block at which the provider will either forcibly top up their deposit or be marked as
+    /// insolvent.
+    ///
+    /// This is the relay chain block number which the parachain is anchored to.
+    pub end_block_grace_period: RelayChainBlockNumber,
+}
```
ffarall, line 24: Is it worth having a struct for just this field?

pallets/providers/src/utils.rs
```diff
@@ -830,53 +795,200 @@ where
...
+            MainStorageProviders::<T>::mutate(provider_id, |p| *p = Some(provider));
+        } else if let Some(mut provider) = BackupStorageProviders::<T>::get(provider_id) {
+            provider.capacity = capacity;
+            BackupStorageProviders::<T>::mutate(provider_id, |p| *p = Some(provider));
+        }
+
+        Ok(())
+    }
+
+    /// Allows a storage provider to manually top up their held deposit to restore capacity up to their currentlt used capacity.
```
ffarall, line 924:
```suggestion
    /// Allows a storage provider to manually top up their held deposit to restore capacity up to their currently used capacity.
```

pallets/providers/src/utils.rs
```diff
@@ -830,53 +795,200 @@ where
...
+    ///
+    /// The provider must be within a grace period due to insufficient capacity.
+    /// Holds the required amount from the provider's free balance to match their used capacity.
+    ///
+    /// This will short circuit if the provider is not registered, doesn't need a top-up, or lacks sufficient balance.
```
ffarall, line 929:
```suggestion
    /// This will error out if the provider is not registered, doesn't need a top-up, or lacks sufficient balance.
```

pallets/providers/src/utils.rs
```diff
@@ -830,53 +795,200 @@ where
...
+            required_delta_hold_amount,
+        )?;
+
+        // Set the provider's capacity to the used capacity
+        capacity = used_capacity;
+
+        // Update the provider's capacity in storage
+        if let Some(mut provider) = MainStorageProviders::<T>::get(provider_id) {
+            provider.capacity = capacity;
+            MainStorageProviders::<T>::mutate(provider_id, |p| *p = Some(provider));
```
ffarall, line 966: I understand that you're just setting the value, why use `mutate` then?

pallets/providers/src/utils.rs
ffarall: Also, it doesn't seem to error out if the amount needed to top up is zero. Which is fine btw, just the comment would be wrong.

pallets/providers/src/utils.rs
```diff
@@ -1135,6 +1249,44 @@ where
...
+            .checked_mul(&capacity_over_minimum.into())
+            .ok_or(ArithmeticError::Overflow)?;
+        T::SpMinDeposit::get()
+            .checked_add(&deposit_for_capacity_over_minimum)
+            .ok_or(ArithmeticError::Overflow.into())
+    }
+
+    fn get_provider_details(
+        provider_id: ProviderIdFor<T>,
+    ) -> Result<(T::AccountId, StorageDataUnit<T>, StorageDataUnit<T>), DispatchError>
```
ffarall, line 1270: Not a huge fan of returning a tuple here, but ok.

pallets/providers/Cargo.toml
```diff
@@ -16,6 +16,7 @@ targets = ["x86_64-unknown-linux-gnu"]
 
 [dependencies]
 codec = { workspace = true }
+log = { workspace = true }
```
ffarall, line 19: Is this still needed?

pallets/providers/src/tests.rs
```diff
@@ -4808,166 +4814,426 @@ mod slash {
...
+                        )
+                        .unwrap()
+                        .into();
+                    deposit_needed_for_capacity_used.saturating_sub(NativeBalance::balance_on_hold(
+                        &StorageProvidersHoldReason::get(),
+                        &self.account,
+                    ))
+                };
+
+                // Set proofs-dealer storage to have a slashable provider with a accrued slashes based on the test setup
```
ffarall, line 4931:
```suggestion
                // Set proofs-dealer storage to have a slashable provider with an accrued slashes based on the test setup
```

pallets/providers/src/tests.rs
```diff
@@ -4808,166 +4814,426 @@ mod slash {
...
                 pallet_proofs_dealer::SlashableProviders::<Test>::insert(
-                    &bob_provider_id,
-                    bob_accrued_failed_proof_submissions,
+                    &self.provider_id,
+                    if self.slash_held_deposit { u32::MAX } else { 1 },
```
ffarall, line 4934: Why the fact that `slash_held_deposit` is `true` means the accrued failed proof submissions should be `u32::MAX`?

ffarall: Is this maybe left from when you were first slashing from the free balance?

pallets/providers/src/tests.rs
```diff
@@ -4808,166 +4814,426 @@ mod slash {
...
-                let actual_slashed_amount = slash_factor.min(deposit_on_hold);
+        use super::*;
 
-                // Check that the Treasury has received the slash amount
-                assert_eq!(
-                    NativeBalance::free_balance(&<Test as crate::Config>::Treasury::get()),
-                    treasury_balance + actual_slashed_amount
-                );
-            });
+        struct TestSetup {
```
ffarall, line 4878: Personally, I'm not a fan of having one "master" logic that handles all the test cases, based on some configurations. I see the benefit of having one piece of code to update if you want to change something, but this is tests code, not implementation code. So it is fine to have repeated code with slightly different conditions that make up a specific scenario.

I personally prioritise looking at the test and understanding step by step how the scenario is setup, then what action is taken, and how the result is checked. Here I have to go to the test, remember which flags were set, and understand how that shapes the behaviour of this one-for-all implementation. Which is taking me some minutes of looking at the screen 😂.

Anyway, up for debate.

pallets/providers/src/tests.rs
```diff
@@ -4808,166 +4814,426 @@ mod slash {
...
+            ExtBuilder::build().execute_with(|| {
+                let mut test_setup = TestSetup::default();
+
+                test_setup.slash_held_deposit = true;
+                test_setup.slash_and_verify();
+
+                test_setup.automatic_top_up = true;
+                test_setup.slash_and_verify();
+            });
+        }
```
ffarall, line 5166: Took me more time than I'm willing to admit, but... I understood why any of your tests fail if the BSP can top up immediately after having a capacity deficit. None of your tests have `automatic_top_up` enabled the first time you `slash_and_verify`. If you do that, slashing fails.

pallets/providers/src/tests.rs
ffarall: As a counter argument to my own argument, I see that this helped you to have more testing setups and scenarios, which is always appreciated.

runtime/src/configs/mod.rs
```diff
@@ -502,6 +516,7 @@ impl pallet_storage_providers::Config for Runtime {
     type MaxCommitmentSize = ConstU32<1000>;
     type ZeroSizeBucketFixedRate =
         runtime_params::dynamic_params::runtime_config::ZeroSizeBucketFixedRate;
+    type TopUpGracePeriod = ConstU32<600>; // 1 hour with 6 second timeslots.
```
ffarall, line 519: Wasn't this meant to be in the order of days?

test/suites/integration/bsp/submit-proofs.test.ts
```diff
@@ -595,10 +597,6 @@ describeBspNet(
         dummyBspNextChallengeTick < bspTwoNextChallengeTick
           ? dummyBspNextChallengeTick
           : bspTwoNextChallengeTick;
-      const secondBlockToAdvance =
-        dummyBspNextChallengeTick < bspTwoNextChallengeTick
-          ? bspTwoNextChallengeTick
-          : dummyBspNextChallengeTick;
```
ffarall, line 601: What's up with this deletion here?

test/suites/integration/bsp/slash-provider.test.ts
```diff
@@ -1,65 +1,69 @@
 import { strictEqual } from "node:assert";
 import { describeBspNet, type EnrichedBspApi } from "../../../util";
 
-describeBspNet("BSPNet: Slash Provider", ({ before, createUserApi, createBspApi, it }) => {
-  let userApi: EnrichedBspApi;
-  let bspApi: EnrichedBspApi;
+describeBspNet(
+  "BSPNet: Slash Provider",
+  { only: true },
```
ffarall, line 6: Are these `only` leftover from debugging?

ffarall (changes requested):

pallets/providers/src/utils.rs
```diff
@@ -830,53 +795,200 @@ where
...
             provider_id: *provider_id,
-            amount_slashed,
+            amount: actual_slashed,
         });
 
-        Ok(Pays::No.into())
+        // No need to top up if the provider's used capacity is below the minimum capacity
+        if used_capacity < T::SpMinCapacity::get() {
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
snowmead: yes thanks for pointing this out

pallets/providers/src/utils.rs
```diff
@@ -830,53 +795,200 @@ where
...
+            )?;
+
+            // Increase capacity up to the used capacity
+            capacity = used_capacity;
+
+            let top_up_metadata = expect_or_err!(
+                AwaitingTopUpFromProviders::<T>::take(provider_id),
+                "Top up metadata should exist when there is a capacity deficit",
+                Error::<T>::TopUpNotRequired
+            );
```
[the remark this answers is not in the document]
snowmead: yes definitely agree with this, missed a test case which would have resulted in this error you mentioned

pallets/providers/src/tests.rs
snowmead: changed it so that we calculate the accrued slashes needed to slash the entire held deposit

pallets/providers/src/utils.rs
snowmead: I use `set` now

pallets/providers/src/utils.rs
snowmead: ya it's just a convenience function for extracting the exact values I need for the slashing and top up operations. I don't want to create another struct or return an enum just for this.

pallets/providers/Cargo.toml
snowmead: no thanks for spotting

test/suites/integration/bsp/submit-proofs.test.ts
snowmead: both BSPs were responding in the same block. this test was failing because we were expecting the second BSP to respond in a subsequent block to the first one being responded to by the first BSP.

I couldn't see any part of my code that would affect this test.

test/suites/integration/bsp/slash-provider.test.ts
snowmead: yes thanks for spotting

runtime/src/configs/mod.rs
snowmead: thanks for spotting, I had put this initially

pallets/providers/src/tests.rs
snowmead: yes exactly. I added this test to cover this case :)

ffarall (approved):

PR #256, opened 2024-11-13
feat: ✨ Bound execution of slashable providers loop in Proofs Dealer's `on_poll`

ffarall:
In this PR:
1. Bound the execution of the `on_poll` hook in Proofs Dealer pallet.
    1.  In the `do_new_challenges_round` function, the `while` loop that iterated through all the Providers that had to be marked as slashable was unbounded.
    2. While it is expected that little to no Providers should be left to loop through, it is still technically possible that all Providers match their challenge cycles and fail to submit proofs. Given that the number of Providers is unbounded, this can result in an unbounded execution.
    3. This is now limited by the `MaxSlashableProvidersPerTick` config parameter, which should be set to a reasonably high value after benchmarking the hook.
    4. The function can now catch up in subsequent blocks if this limit is reached at a given block.
2. Add unit test for newly added functionality.
3. Remove manual weight metering from `on_poll` hook (in preparation for benchmarking)


TDemeco (approved):
Went through the logic step by step to make sure I wasn't missing anything and it all seems good 👍🏻 just some nitpicks about clarity

PR #257, opened 2024-11-14
fix: :bug: add checks to avoid panicking when decoding proofs with `sp-trie`

TDemeco:
This happens because we accept arbitrary proofs from users, and `sp-trie` has an issue where a purposely crafted bad proof makes it panic when decoding it (instead of returning an error).

I've already opened a [PR](https://github.com/paritytech/polkadot-sdk/pull/6486) to fix it upstream but in the meantime we need this fix to allow the auditors to keep testing.


TDemeco:
Closed as this is fixed in #258

PR #260, opened 2024-11-18
Feat: index Merkle roots for BSPs and Buckets

links234:


pallets/file-system/src/lib.rs
```diff
@@ -819,6 +820,7 @@ pub mod pallet {
                 msp_id,
                 bucket_id,
                 name,
+                root: <T::ProofDealer as shp_traits::ProofsDealerInterface>::MerkleHash::default(),
```
ffarall, line 823: If everytime there's a new bucket, it's root is the default, why would you need to pass that information on the event? You can access this type from the client, and get its default.

pallets/providers/src/lib.rs
```diff
@@ -460,10 +460,12 @@ pub mod pallet {
         },
 
         /// Event emitted when a Backup Storage Provider has confirmed its sign up successfully. Provides information about
-        /// that BSP's account id, the total data it can store according to its stake, and its multiaddress.
+        /// that BSP's account id, the initial root of the Merkle Patricia Trie that it stores, the total data it can store
+        /// according to its stake, and its multiaddress.
         BspSignUpSuccess {
             who: T::AccountId,
             bsp_id: BackupStorageProviderId<T>,
+            root: MerklePatriciaRoot<T>,
```
ffarall, line 468: Here as well, it would be the default root.

ffarall (approved):
Approved with comment to optimise.

pallets/file-system/src/lib.rs
links234: I know, I specifically wanted to make it fool-proof for someone that might want to write an indexer in another language - this way I don't have to care about which hashing scheme runtime is using - I already have all the info.

PR #262, opened 2024-11-18
Processing expired file deletion requests reduce bucket size

snowmead:
Reduce the bucket size when we process expired file deletion requests.

The user which initiates a deletion request might not provider a forest proof of inclusion and so it is possible that the user sets a file size greater than in actuality. The runtime cannot verify this but we have no choice but to delete the file anyways and reduce the bucket size by the specified amount since we cannot gatekeep users from deleting their files. 

This will incentivize the MSPs to submit proofs of inclusion for the file deletion requests if ever the size of the file is false.


pallets/file-system/src/tests.rs
```diff
@@ -6690,6 +6690,287 @@ mod delete_file_and_pending_deletions_tests {
...
+                    vec![]
+                );
+
+                // Asser that the pending file deletion request was removed from storage
+                assert_eq!(
+                    file_system::PendingFileDeletionRequests::<Test>::get(owner_account_id.clone()),
+                    BoundedVec::<_, <Test as file_system::Config>::MaxUserPendingDeletionRequests>::default()
+                );
+
+                // Assert that the bucket size was not decreased because the bucket was not found
```
ffarall, line 6957: I think this comment is wrong

pallets/file-system/src/tests.rs
```diff
@@ -6690,6 +6690,287 @@ mod delete_file_and_pending_deletions_tests {
...
+
+                // Assert that there was no queued priority challenge for file key in proofs dealer pallet
+                assert!(!pallet_proofs_dealer::PriorityChallengesQueue::<Test>::get()
+                .iter()
+                .any(|x| *x == (file_key, Some(TrieRemoveMutation))),);
+            });
+        }
+
+        #[test]
+        fn delete_file_expired_pending_file_deletion_request_with_insolvent_user() {
```
ffarall, line 6835: Good that you added this test. It will be helpful then to check if we change the behaviour of this.

ffarall (approved):
Great job! Fix the tests in the CI and all good.

PR #263, opened 2024-11-19
test: :test_tube: New re-org tests and helpers

timbrinded:
Changes:
- Adds new `extendFork()` to the block api, to allow the system
  -  NB this current causes a wasm panic on forks longer than 3 blocks
- CI sharding changes to improve execution times
- New disabled tests prior to the reorg functionality that is soon to be added


test/suites/integration/bsp/submit-proofs.test.ts
```diff
@@ -242,7 +242,49 @@ describeBspNet(
       // TODO: Check that BSP-Three no longer has a challenge deadline.
     });
 
-    it("BSP submits proof, transaction gets dropped, BSP-resubmits and succeeds", async () => {
+    it(
+      "BSP submits proof, transaction gets dropped, BSP-resubmits and succeeds",
+      { skip: "Not implemented yet." },
```
ffarall, line 247: Why skipping this? Wasn't this passing?

test/suites/integration/bsp/submit-proofs.test.ts
```diff
@@ -242,7 +242,49 @@ describeBspNet(
...
+
+        await userApi.block.seal();
+        await userApi.assert.extrinsicPresent({
+          module: "proofsDealer",
+          method: "submitProof"
+        });
+      }
+    );
+
+    it("Proofs not submitted in unfinalised block", { skip: "Not implemented yet." }, async () => {
```
ffarall, line 287: I think there's been a confusion, and I'm likely to blame. BSPs _should_ submit proofs even if blocks are not finalised. Otherwise, if finality is lagging and they don't submit proofs, once finality catches up, they might already be late to submit it and be slashed.

ffarall (changes requested):
Requesting changes but mostly to have a quick chat and clarify what the testing case that we wanted to test was about.

timbrinded:
Looks live my refactor has broken something, so need to do a fix next week- but in the meantime have a look at the new test cases ive written. Three failing test cases which should be useful for your feature branch next week @ffarall

test/util/bspNet/block.ts
```diff
@@ -25,6 +25,39 @@ export interface SealedBlock {
   extSuccess?: boolean;
 }
 
+export const extendFork = async (
```
ffarall, line 28: Amazing helper. Could use some docs on what it does and what to pass it. Or at least the wrapper in `test-api.ts`, which is the one we'll be using in tests.

test/util/bspNet/block.ts
```diff
@@ -25,6 +25,39 @@ export interface SealedBlock {
...
+  api: ApiPromise,
+  options: { parentBlockHash: string; amountToExtend: number; verbose: boolean }
```
ffarall, line 30: How about making the verbose patameter optional?

```suggestion
  options: { parentBlockHash: string; amountToExtend: number; verbose?: boolean }
```

ffarall (approved):
Just som eminor comments, but great work!

ffarall:
Also, please change the title and description of the PR, they don't really fit now.

test/util/bspNet/block.ts
timbrinded: Wrapper is already there, it's how it's being called in the `reorg-proof.test` file. Added docs now though

PR #264, opened 2024-11-19
refactor: `pay_outstanding_debt` extrinsic of `pallet-payment-streams`

TDemeco:
This PR:
- Adds a runtime API `get_providers_with_payment_streams_with_user`, which receives a user and returns the list of Providers that have at least one payment stream with it. This is useful for users that were deemed insolvent and want to recoup as much of their paid deposits as possible calling the `pay_outstanding_debt` extrinsic of the payment streams pallet.
- Refactors the `pay_outstanding_debt` extrinsic to, instead of receiving the number of payment streams to pay, it receives a list of Providers. This way, it checks if that Provider has payment streams with the user and pays it the remaining debt, without having unbounded iterations. Weight is then proportional to the amount of Providers that the user wants to repay.


ffarall (approved):

pallets/payment-streams/src/lib.rs
```diff
@@ -801,36 +801,38 @@ pub mod pallet {
...
+        /// which can be obtained by calling the `get_providers_with_payment_streams_with_user` runtime API.
+        /// There was an idea to limit the amount of Providers that can be received by this extrinsic using a constant in the configuration of this pallet,
+        /// but the correct benchmarking of this extrinsic should be enough to avoid any potential abuse.
         #[pallet::call_index(8)]
-        #[pallet::weight(T::WeightInfo::pay_outstanding_debt(*amount_of_streams_to_pay))]
+        #[pallet::weight(T::WeightInfo::pay_outstanding_debt(providers.len().try_into().unwrap_or(u32::MAX)))]
         pub fn pay_outstanding_debt(
             origin: OriginFor<T>,
-            amount_of_streams_to_pay: u32,
+            providers: Vec<ProviderIdFor<T>>,
```
snowmead, line 829: unbounded 😱😆

snowmead (approved):

pallets/payment-streams/src/lib.rs
TDemeco: It's bounded by the weight of the extrinsic! Basically the caller of this will pay for its execution proportional to the amount of Providers, so it shouldn't be an issue. I thought about making it a bounded vec and bounding it with a constant of the Config of the pallet, maybe that's better? @ffarall cc'ing you just in case

ffarall: Worst case scenario, the user sends a transaction that can take up A LOT of weight, but we know that in advance and will charge him for it. Whether or not this could result in another issue... unclear to me.

PR #265, opened 2024-11-20
test: :white_check_mark: Implement TODOs and Fix Tests

timbrinded:
- Update `fetchEventData()` -> `fetchEvent()` to provide typed event data object instead of tuple
- Re-added volunteer tests to the top level suite
- Removed some TODOs and refactor


ffarall (approved):
I like what you did here! Let me know if you need a hand with the failing tests. Might be a carried over bug from recent changes where the test has not been running.

PR #266, opened 2024-11-20
refactor: ♻️ change the price updater to return the price per gigabyte

TDemeco:
This PR refactors the traits in charge of updating the price per unit of data per tick to instead use the price per gigaunit (in our case, gigabyte) per tick. This allows this price to fit within the 10 decimal places of DOT, but comes with quite a few changes in the logic. Most importantly, any logic that interacted with dynamic-rate payment streams (be it to charge them or update their deposit) now has a division to account for this increase.

- Important note 1: because of this division the minimum amount to charge for dynamic-rate payment streams (after which the division is no longer 0) is when the price index difference (between the time when it was last charged and the time in which the provider submitted its last proof) times the amount provided is greater than 1 gigaunit (1_073_741_824). Considering our worst case scenario (where the system is at its minimum price of 10 nanounits), a Provider that holds around 10MB from a user will be able to charge him every block.

- Important note 2: following the same logic, when creating a new dynamic-rate payment stream the equation `current_price_per_giga_unit_per_tick * amount_provided * NewStreamDeposit::get()` has to be greater than 1 gigaunit. This means that, when a BSP volunteers and accepts a file from a user for which it didn't had any files before, it has to be bigger than 107KB, which is quite limiting. My proposed solution (which I'd implement in this PR) would be to just not have a deposit for payment streams that small, and as it gets bigger the deposit starts to grow.
  - Note: this was fixed by having a minimum deposit and saturating to it whenever the payment stream would otherwise have an empty deposit


snowmead:
I agree with your take that payment streams should only be created after some minimum threshold has been reached.

snowmead (approved):
Nice job! Just a couple of comments.

Thanks for fixing the fixed rate payment stream updates. It was not clear to me that the function was expected a delta instead of a new set value.

primitives/data-price-updater/src/lib.rs
```diff
@@ -1,5 +1,11 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
+//! This crate holds the logic for updating the price index of storage based on the system utilisation.
+//! It provides two structs that implement the `UpdateStoragePrice` trait:
+//! - `NoUpdatePriceIndexUpdater`: A struct that does not update the price index.
+//! - `MostlyStablePriceIndexUpdater`: A struct that updates the price index based on the system utilisation,
+//! but keeps a mostly stable price.
```
ffarall, line 7: Nice addition

pallets/providers/src/utils.rs
```diff
@@ -1035,6 +1035,9 @@ where
 
         match delta {
             RateDeltaParam::NewBucket => {
+                // Get the rate over that of a zero sized bucket of the bucket to add.
+                // If its size is less than the minimum for which rate starts to increase (which depends on the value proposition),
+                // then the rate over the zero size bucket is 0.
                 let bucket_rate = value_prop
                     .price_per_unit_of_data_per_block
```
ffarall, line 1042: Why don't we make value propositions also be in "price per giga-unit of data, per block"?

pallets/providers/src/utils.rs
```diff
@@ -1086,6 +1092,15 @@ where
                 }
             }
             RateDeltaParam::Increase(delta) => {
+                // Get the current bucket rate, which is the rate of a zero sized bucket plus the rate according to the bucket size.
+                let bucket_rate = value_prop
+                    .price_per_unit_of_data_per_block
+                    .checked_mul(&bucket.size.into())
+                    .ok_or(ArithmeticError::Overflow)?
+                    .checked_add(&T::ZeroSizeBucketFixedRate::get())
+                    .ok_or(ArithmeticError::Overflow)?;
```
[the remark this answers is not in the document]
ffarall: Yeah this is a bit confusing, because here we calculate the `bucket_rate` as:
```rust
                let bucket_rate = value_prop
                    .price_per_unit_of_data_per_block
                    .checked_mul(&bucket.size.into())
                    .ok_or(ArithmeticError::Overflow)?
                    .checked_add(&T::ZeroSizeBucketFixedRate::get())
                    .ok_or(ArithmeticError::Overflow)?;
```

But in `RateDeltaParam::NewBucket` or `RateDeltaParam::RemoveBucket` we calculate it as:
```rust
                let bucket_rate = value_prop
                    .price_per_unit_of_data_per_block
                    .checked_mul(&bucket.size.into())
                    .ok_or(ArithmeticError::Overflow)?;
```

And only then add the zero size bucket rate to the `new_rate`. We could unify it. But no big deal.

pallets/providers/src/utils.rs
```diff
@@ -1097,38 +1112,67 @@ where
...
+                    .checked_sub(&new_bucket_rate)
+                    .ok_or(ArithmeticError::Underflow)?;
+
+                // If the rate has changed, update the payment stream.
+                if !delta_rate.is_zero() {
+                    // Since this is a decrease, subtract the delta rate from the current rate.
+                    let new_rate = current_rate.saturating_sub(delta_rate);
 
-                if new_rate < T::ZeroSizeBucketFixedRate::get() {
-                    <T::PaymentStreams as PaymentStreamsInterface>::delete_fixed_rate_payment_stream(
```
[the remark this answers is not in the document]
ffarall: I think what had been suggested was to make the deposit be 0. Anyway, I wouldn't make anything free. If anything, I'd provide minimums.

pallets/payment-streams/src/utils.rs
```diff
@@ -12,17 +12,19 @@ use frame_support::traits::{
...
 };
 use sp_runtime::{
-    traits::{Convert, One},
+    traits::{CheckedDiv, Convert, One},
     Saturating,
 };
 
 use crate::{weights::WeightInfo, *};
 
+const GIGAUNIT: u32 = 1_073_741_824;
```
ffarall, line 26: This could go in `primitives/constants`

ffarall (approved):

pallets/providers/src/utils.rs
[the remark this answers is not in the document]
TDemeco: I unified this logic using `multiply_rational` so it should be consistent and more concise now

pallets/providers/src/utils.rs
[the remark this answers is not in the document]
TDemeco: Payment streams should only be deleted when deleting the last bucket of it, otherwise there could be empty buckets lying around that are not charged even though we have a zero-sized bucket rate

pallets/providers/src/utils.rs
TDemeco: Good call, done!

pallets/payment-streams/src/utils.rs
TDemeco: Yep, refactored.

test/suites/integration/bsp/debt-collection.test.ts
```diff
@@ -301,20 +312,24 @@ describeBspNet(
...
       const freeBalance = (await userApi.query.system.account(userAddress)).data.free;
 
       // To make the user insolvent, we need to update the payment stream with a very high amount
-      // and advance new stream deposit blocks
-      // To do this, the new amount provided should be equal to the free balance of the user divided by
-      // the current price of storage multiplied by the new stream deposit
+      // of amount provided so when the BSP tries to charge it the user cannot pay its debt.
+      // We set the new provided amount to be as much as possible considering the deposit
+      // that the user is going to have to pay. We leave 10 * ED in the account to allow
+      // the user to pay its other debts.
```
santikaplan, line 327:
```suggestion
      // so that when the BSP attempts to charge it, the user cannot cover the debt.
      // We set the new provided amount to be as much as possible, considering the deposit
      // that the user is required to pay. We leave 10 * ED in the account to ensure
      // the user can still pay other debts.
```

ffarall (approved):

PR #268, opened 2024-11-21
feat: add task to let MSP charge user every X blocks

undercover-cactus:
This PR add a `NotifyPeriod` event that would be triggered every X blocks. It helps notifyng the MSP task that X blocks has passed and that it can charge users for their service.

The amount of blocks (aka the msp charging frequency) is a configurable value that can be passed through the CLI.

### DIFF
--- a/client/blockchain-service/src/handler.rs
+++ b/client/blockchain-service/src/handler.rs
@@ -52,7 +52,7 @@ use crate::{
     events::{
         AcceptedBspVolunteer, BlockchainServiceEventBusProvider,
         FinalisedTrieRemoveMutationsApplied, LastChargeableInfoUpdated, NewStorageRequest,
-        SlashableProvider, SpStopStoringInsolventUser, UserWithoutFunds,
+        NotifyPeriod, SlashableProvider, SpStopStoringInsolventUser, UserWithoutFunds,
     },
     state::{
         BlockchainServiceStateStore, LastProcessedBlockNumberCf,
@@ -109,6 +109,8 @@ pub struct BlockchainService {
     /// various edge cases when restarting the node, all originating from the "dynamic" way of
     /// computing the next challenges tick. This case is handled separately.
     pub(crate) pending_submit_proof_requests: BTreeSet<SubmitProofRequest>,
+    /// Notify period value to know when to trigger the NotifyPeriod event.
+    notify_period: Option<u32>,
 }
 
 /// Event loop for the BlockchainService actor.
@@ -931,6 +933,7 @@ impl BlockchainService {
         rpc_handlers: Arc<RpcHandlers>,
         keystore: KeystorePtr,
         rocksdb_root_path: impl Into<PathBuf>,
+        notify_period: Option<u32>,
     ) -> Self {
         Self {
             client,
@@ -945,6 +948,7 @@ impl BlockchainService {
             last_block_processed: Zero::zero(),
             persistent_state: BlockchainServiceStateStore::new(rocksdb_root_path.into()),
             pending_submit_proof_requests: BTreeSet::new(),
+            notify_period,
         }
     }
 
@@ -1063,6 +1067,9 @@ impl BlockchainService {
 
         // Process pending requests that update the forest root.
         self.check_pending_forest_root_writes();
+
+        // Check that trigger an event every X amount of blocks (specified in config).
+        self.check_for_notify(&block_number);
 
         let state_store_context = self.persistent_state.open_rw_context_with_overlay();
         // Get events from storage.
@@ -1295,6 +1302,14 @@ impl BlockchainService {
                 // TODO: a node that has a newer version of the runtime, therefore the EventsVec type is different.
                 // TODO: Consider using runtime APIs for getting old data of previous blocks, and this just for current blocks.
                 error!(target: LOG_TARGET, "Failed to get events storage element: {:?}", e);
+            }
+        }
+    }
+
+    fn check_for_notify(&self, block_number: &BlockNumber) {
+        if let Some(np) = self.notify_period {
+            if block_number % np == 0 {
+                self.emit(NotifyPeriod {});
             }
         }
     }
--- a/node/src/cli.rs
+++ b/node/src/cli.rs
@@ -109,6 +109,12 @@ pub struct ProviderConfigurations {
     /// Extrinsic retry timeout in seconds.
     #[clap(long, default_value = "60")]
     pub extrinsic_retry_timeout: u64,
+
+    /// MSP charging fees frequency.
+    #[clap(long, required_if_eq_any([
+        ("provider_type", "msp"),
+    ]))]
+    pub msp_charging_freq: Option<u32>,
 }
 
 impl ProviderConfigurations {
@@ -128,6 +134,7 @@ impl ProviderConfigurations {
             max_storage_capacity: self.max_storage_capacity,
             jump_capacity: self.jump_capacity,
             extrinsic_retry_timeout: self.extrinsic_retry_timeout,
+            msp_charging_freq: self.msp_charging_freq,
         }
     }
 }
--- a/node/src/services/builder.rs
+++ b/node/src/services/builder.rs
@@ -107,6 +107,7 @@ where
     max_storage_capacity: Option<StorageDataUnit>,
     jump_capacity: Option<StorageDataUnit>,
     extrinsic_retry_timeout: u64,
+    notify_period: Option<u32>,
 }
 
 /// Common components to build for any given configuration of [`RoleSupport`] and [`StorageLayerSupport`].
@@ -125,6 +126,7 @@ where
             max_storage_capacity: None,
             jump_capacity: None,
             extrinsic_retry_timeout: DEFAULT_EXTRINSIC_RETRY_TIMEOUT_SECONDS,
+            notify_period: None,
         }
     }
 
@@ -166,6 +168,18 @@ where
         self
     }
 
+    // TODO: add a function ´with_notify_period´ to be called in the service.
+    // but check that ´blockchain´ is None (so we don't have called with_blockchain before)
+    pub fn with_notify_period(&mut self, notify_period: u32) -> &mut Self {
+        if self.blockchain.is_some() {
+            panic!(
+                "`with_notify_period`should never be called after starting the blockchain service."
+            );
+        }
+        self.notify_period = Some(notify_period);
+        self
+    }
+
     pub async fn with_blockchain(
         &mut self,
         client: Arc<ParachainClient>,
@@ -181,6 +195,7 @@ where
             rpc_handlers.clone(),
             keystore.clone(),
             rocksdb_root_path,
+            self.notify_period,
         )
         .await;
 
@@ -410,6 +425,7 @@ pub trait RequiredStorageProviderSetup {
         max_storage_capacity: Option<StorageDataUnit>,
         jump_capacity: Option<StorageDataUnit>,
         extrinsic_retry_timeout: u64,
+        msp_charging_freq: Option<u32>,
     );
 }
 
@@ -424,11 +440,16 @@ where
         max_storage_capacity: Option<StorageDataUnit>,
         jump_capacity: Option<StorageDataUnit>,
         extrinsic_retry_timeout: u64,
+        msp_charging_freq: Option<u32>,
     ) {
         self.setup_storage_layer(storage_path);
         if max_storage_capacity.is_none() {
             panic!("Max storage capacity not set");
         }
+        if let Some(notify_period) = msp_charging_freq {
+            self.with_notify_period(notify_period.clone());
+        }
+
         self.with_max_storage_capacity(max_storage_capacity);
         self.with_jump_capacity(jump_capacity);
         self.with_retry_timeout(extrinsic_retry_timeout);
@@ -446,6 +467,7 @@ where
         max_storage_capacity: Option<StorageDataUnit>,
         jump_capacity: Option<StorageDataUnit>,
         extrinsic_retry_timeout: u64,
+        msp_charging_freq: Option<u32>,
     ) {
         if storage_path.is_none() {
             panic!("Storage path not set");
@@ -453,6 +475,9 @@ where
         self.setup_storage_layer(storage_path);
         if max_storage_capacity.is_none() {
             panic!("Max storage capacity not set");
+        }
+        if let Some(notify_period) = msp_charging_freq {
+            self.with_notify_period(notify_period.clone());
         }
         self.with_max_storage_capacity(max_storage_capacity);
         self.with_jump_capacity(jump_capacity);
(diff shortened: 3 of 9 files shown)


node/src/services/builder.rs
```diff
@@ -424,11 +440,16 @@ where
...
+        msp_charging_freq: Option<u32>,
     ) {
         self.setup_storage_layer(storage_path);
         if max_storage_capacity.is_none() {
             panic!("Max storage capacity not set");
         }
+        if let Some(notify_period) = msp_charging_freq {
+            self.with_notify_period(notify_period.clone());
+        }
+
```
undercover-cactus, line 452: I am not super happy for having to do this with all different `setup` function. It is a bit redundant with the builder strategy.

node/src/tasks/msp_charge_fees.rs
```diff
@@ -0,0 +1,129 @@
...
+        info!(
+            target: LOG_TARGET,
+            "Charging users",
+        );
+
+        let own_provider_id = self
+            .storage_hub_handler
+            .blockchain
+            .query_storage_provider_id(None)
+            .await?;
```
undercover-cactus, line 70: Shouldn't MSP task have access to their own msp id ?

test/suites/integration/msp/debt-collection.test.ts
```diff
@@ -0,0 +1,254 @@
...
+      const currentBlock = await userApi.rpc.chain.getBlock();
+      const currentBlockNumber = currentBlock.block.header.number.toNumber();
+
+      const blocksToAdvance = currentBlockNumber + (12 - (currentBlockNumber % 12)) + 1; // 12 is the msp_charging_freq that we setup
+      if (blocksToAdvance > currentBlockNumber) {
+        await userApi.advanceToBlock(blocksToAdvance);
+      }
+
+      // Verify that the MSP charged the users after the notified
+      await userApi.assert.eventPresent("paymentStreams", "PaymentStreamCharged");
```
undercover-cactus, line 251: We should probably verify the amount charged to make sure it is correct.

test/suites/integration/msp/debt-collection.test.ts
```diff
@@ -0,0 +1,250 @@
...
+        "StorageRequestFulfilled"
+      );
+      storageRequestFulfilledDataBlob =
+        userApi.events.fileSystem.StorageRequestFulfilled.is(storageRequestFulfilledEvent) &&
+        storageRequestFulfilledEvent.data;
+    } catch {
+      // Event not found, continue
+    }
+
+    let acceptedFileKey: string | null = null;
```
timbrinded, line 136: try not to use null, use undefined instead

test/suites/integration/msp/debt-collection.test.ts
```diff
@@ -0,0 +1,250 @@
...
+      }
+
+      issuedFileKeys.push(newStorageRequestDataBlob.fileKey);
+    }
+
+    // Seal block containing the MSP's transaction response to the storage request
+    await userApi.wait.mspResponseInTxPool();
+    await userApi.sealBlock();
+
+    let mspAcceptedStorageRequestDataBlob: any = undefined;
```
timbrinded, line 109: whats going on with these empty try catch blocks and any data blobs?

if this fails in future it will be hard to debug and its not obvious what this test is trying to do.

I would instead have some branching logic that is determined by query of last block to see which of the conditions is fulfilled (and reject if none). 

Typescript doesn't have `match e { ... }` sadly, but we do have:

```ts
switch (true) {
   case e : {...} 
}
```

Instead of use the asserts to find the events like this, might more straight forward to search for the event manually by querying e.g. `(await api.query.system.events()).filter(x=> x.event.section.toString() === "fileSystem")`

and from there you can limit the amount of lets and untyped state hanging around  :cold_sweat:

test/suites/integration/msp/debt-collection.test.ts
```diff
@@ -0,0 +1,250 @@
...
+    // There is only a single key being accepted since it is the first file key to be processed and there is nothing to batch.
+    strictEqual(
+      issuedFileKeys.some((key) => key.toString() === acceptedFileKey),
+      true
+    );
+
+    // Allow time for the MSP to update the local forest root
+    await sleep(3000);
+
+    const local_bucket_root = await mspApi.rpc.storagehubclient.getForestRoot(
```
timbrinded, line 159: ts uses camel case soz

test/suites/integration/msp/debt-collection.test.ts
```diff
@@ -0,0 +1,250 @@
...
+    const { event: bucketRootChangedEvent } = await userApi.assert.eventPresent(
+      "providers",
+      "BucketRootChanged"
+    );
+
+    const bucketRootChangedDataBlob =
+      userApi.events.providers.BucketRootChanged.is(bucketRootChangedEvent) &&
+      bucketRootChangedEvent.data;
+
+    if (!bucketRootChangedDataBlob) {
```
timbrinded, line 172: use invariant

test/suites/integration/msp/debt-collection.test.ts
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    // There is only a single key being accepted since it is the first file key to be processed and there is nothing to batch.
+    strictEqual(
+      issuedFileKeys.some((key) => key.toString() === acceptedFileKey),
+      true
+    );
+
+    // Allow time for the MSP to update the local forest root
+    await sleep(3000);
```
timbrinded, line 157: is there an associated log line we can wait on?

test/suites/integration/msp/debt-collection.test.ts
```diff
@@ -0,0 +1,250 @@
...
+    const newBucketEventEvent = await userApi.createBucket(bucketName, valuePropId);
+    const newBucketEventDataBlob =
+      userApi.events.fileSystem.NewBucket.is(newBucketEventEvent) && newBucketEventEvent.data;
+
+    if (!newBucketEventDataBlob) {
+      throw new Error("NewBucket event data does not match expected type");
+    }
+
+    const txs = [];
+    for (let i = 0; i < source.length; i++) {
```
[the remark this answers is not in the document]
undercover-cactus: We use ´i´ as the index for the destination too. In order to use a for of loop we would need to modify a bit more.

test/suites/integration/msp/debt-collection.test.ts
undercover-cactus: I have replaced it by searching manually. I also think it is better than `try and catch`

test/suites/integration/msp/debt-collection.test.ts
undercover-cactus: I don't know about that. I agree that we should move away from `sleep`

timbrinded (commented):
Thanks for the updates, tests look lots better. Had some follow up comments but low priority

node/src/tasks/msp_charge_fees.rs
links234: It would be nicer to not have to do this, indeed.

But when starting up, the only known thing is the account - the provider id associated is known by the runtime.

client/blockchain-service/src/handler.rs
```diff
@@ -1298,4 +1305,12 @@ impl BlockchainService {
...
         }
     }
+
+    fn check_for_notify(&self, block_number: &BlockNumber) {
+        if let Some(np) = self.notify_period {
+            if block_number % np == 0 {
+                self.emit(NotifyPeriod {});
+            }
+        }
+    }
```
ffarall, line 1315: This fits better in `utils.rs`

node/src/cli.rs
```diff
@@ -109,6 +109,12 @@ pub struct ProviderConfigurations {
     /// Extrinsic retry timeout in seconds.
     #[clap(long, default_value = "60")]
     pub extrinsic_retry_timeout: u64,
+
+    /// MSP charging fees frequency.
```
ffarall, line 113: You might want to add more info like the unit of this parameter. For example, is it in number of blocks? And if so, could be helpful to give some parameter like "Setting it to 600, with 6 seconds blocks, means charging every 1 hour".

Also, wouldn't this be a _period_ instead of a frequency?

test/suites/integration/bsp/debt-collection.test.ts
```diff
@@ -99,11 +99,9 @@ describeBspNet(
...
       let currentBlock = await userApi.rpc.chain.getBlock();
       let currentBlockNumber = currentBlock.block.header.number.toNumber();
-      const blocksToAdvance = nextChallengeTick - currentBlockNumber;
-
-      // Advance blocksToAdvance blocks.
-      for (let i = 0; i < blocksToAdvance; i++) {
-        await userApi.sealBlock();
+      if (nextChallengeTick > currentBlockNumber) {
+        // Advance to the next challenge tick if needed
+        await userApi.advanceToBlock(nextChallengeTick);
```
ffarall, line 104: Nice fix/update

test/suites/integration/msp/debt-collection.test.ts
```diff
@@ -0,0 +1,246 @@
+import { strictEqual } from "node:assert";
+import { describeMspNet, shUser, sleep, type EnrichedBspApi } from "../../../util";
+import invariant from "tiny-invariant";
+
+describeMspNet(
+  "Single MSP collecting debt",
+  { only: true },
```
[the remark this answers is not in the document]
ffarall: Yep, same here

test/suites/integration/msp/debt-collection.test.ts
ffarall: Agreed

node/src/services/handler.rs
```diff
@@ -2,10 +2,19 @@ use std::sync::Arc;
...
+    msp_upload_file::MspUploadFileTask, sp_slash_provider::SlashProviderTask,
+    user_sends_file::UserSendsFileTask, BspForestStorageHandlerT, FileStorageT,
+    MspForestStorageHandlerT,
+};
 use shc_actors_framework::{
     actor::{ActorHandle, TaskSpawner},
     event_bus::{EventBusListener, EventHandler},
 };
+use shc_blockchain_service::events::NotifyPeriod;
 use shc_blockchain_service::{
```
ffarall, line 18: Merge these two imports

node/src/services/builder.rs
```diff
@@ -166,6 +168,18 @@ where
         self
     }
 
+    // TODO: add a function ´with_notify_period´ to be called in the service.
+    // but check that ´blockchain´ is None (so we don't have called with_blockchain before)
```
ffarall, line 172: Is this a leftover comment?

node/src/services/builder.rs
ffarall: Agreed. In fact all the functions that are not storage-layer-or-provider-type-specific (i.e. that they don't need to have a special implementation for a provider type and storage layer), I'd leave them out of this `setup`.

In other words, to make proper use of the builder pattern I would:
1. Remove the use of `with_notify_period` from here, and call it in `service.rs` (in `init_sh_builder()`), in case the CLI parameter was passed.
2. Do the same with `with_max_storage_capacity`, `with_jump_capacity`, `with_retry_timeout`, as they seem to be called in the same way, regarless of what the provider type and storage layer is. Well almost, in the User is different. But anyway, these tiny initialisation differences are exactly what the builder pattern is meant to be good for.

ffarall: Anyway, I'd like to get @snowmead involved since he was heavily involved in this code.

node/src/tasks/msp_charge_fees.rs
```diff
@@ -0,0 +1,129 @@
+use anyhow::anyhow;
+use sc_tracing::tracing::*;
+use shc_actors_framework::event_bus::EventHandler;
+use shc_blockchain_service::commands::BlockchainServiceInterface;
+use shc_blockchain_service::events::NotifyPeriod;
+use shc_blockchain_service::types::Tip;
+use shc_common::types::MaxUsersToCharge;
+use shc_common::types::StorageProviderId;
```
ffarall, line 8: I prefer merging these imports.

node/src/services/handler.rs
```diff
@@ -2,10 +2,19 @@ use std::sync::Arc;
 use storage_hub_runtime::StorageDataUnit;
 use tokio::sync::RwLock;
 
+use crate::tasks::{
```
ffarall, line 5: I prefer having the `crate` import below all the others as it was.

node/src/tasks/msp_charge_fees.rs
```diff
@@ -0,0 +1,129 @@
...
+
+            let charging_result = self
+                .storage_hub_handler
+                .blockchain
+                .send_extrinsic(call, Tip::from(0))
+                .await;
+
+            match charging_result {
+                Ok(submitted_transaction) => {
+                    info!(target: LOG_TARGET, "Submitted extrinsic to charge users with debt: {}", submitted_transaction.hash());
```
ffarall, line 119: We might want to make this a `debug` log.

node/src/tasks/msp_charge_fees.rs
```diff
@@ -0,0 +1,129 @@
...
+            let call = storage_hub_runtime::RuntimeCall::PaymentStreams(
+                pallet_payment_streams::Call::charge_multiple_users_payment_streams {
+                    user_accounts: users_chunk.to_vec().try_into().expect("Chunk size is the same as MaxUsersToCharge, it has to fit in the BoundedVec"),
+                },
+            );
+
+            let charging_result = self
+                .storage_hub_handler
+                .blockchain
+                .send_extrinsic(call, Tip::from(0))
```
ffarall, line 114: We may want to use `watch_for_success` here.

node/src/tasks/msp_charge_fees.rs
```diff
@@ -0,0 +1,129 @@
...
+use shc_blockchain_service::types::Tip;
+use shc_common::types::MaxUsersToCharge;
+use shc_common::types::StorageProviderId;
+use sp_core::Get;
+use storage_hub_runtime::Balance;
+
+use crate::tasks::{FileStorageT, MspForestStorageHandlerT, StorageHubHandler};
+
+const LOG_TARGET: &str = "msp-charge-fees-task";
+const MIN_DEBT: Balance = 0;
```
ffarall, line 15: These constants and the growing number of configuration parameters for BSPs/MSPs makes it increasingly compelling to have a way to configure them through a configuration file (like a .yml/.toml or similar). If you feel like that's something you'd like to do, setup a task with @santikaplan.

ffarall (changes requested):
Honestly great job! Just a handful of minor suggestions and comments.

node/src/cli.rs
undercover-cactus: Yes period is better. I changed every 'freq' with 'period'.

node/src/tasks/msp_charge_fees.rs
undercover-cactus: For BSP, we are not watching for success. Do we want to have it there too ?

test/suites/integration/msp/debt-collection.test.ts
```diff
@@ -0,0 +1,249 @@
...
+  });
+
+  it("MSP is charging user", async () => {
+    // Calculate how many blocks to advance until next challenge tick.
+    const currentBlock = await userApi.rpc.chain.getBlock();
+    const currentBlockNumber = currentBlock.block.header.number.toNumber();
+
+    const blocksToAdvance = currentBlockNumber + (12 - (currentBlockNumber % 12)) + 1; // 12 is the msp_charging_period that we setup
+    if (blocksToAdvance > currentBlockNumber) {
+      await userApi.advanceToBlock(blocksToAdvance);
```
undercover-cactus, line 234: change this to .block.skipTo()

ffarall (approved):

PR #272, opened 2024-11-25
fix: 🩹 Create a collection only when updating to a private bucket

santikaplan:
When updating a bucket's privacy, we were previously creating and associating a collection when changing the privacy to public (if a collection ID existed but the collection did not exist in Storage). Now, this behavior is restricted to cases where the privacy is changed to private.

Additionally, some minor comment corrections were made to clarify that the `do_create_and_associate_collection_with_bucket` function does not validate the bucket's privacy.

### DIFF
--- a/pallets/file-system/src/utils.rs
+++ b/pallets/file-system/src/utils.rs
@@ -430,8 +430,8 @@ where
             (true, None) => {
                 Some(Self::do_create_and_associate_collection_with_bucket(sender.clone(), bucket_id)?)
             }
-            // Handle case where the bucket has an existing collection.
-            (_, Some(current_collection_id))
+            // Handle case where the bucket has an existing collection, but the collection is not in stroage.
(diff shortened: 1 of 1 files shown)


ffarall (approved):

pallets/file-system/src/utils.rs
```diff
@@ -430,8 +430,8 @@ where
             (true, None) => {
                 Some(Self::do_create_and_associate_collection_with_bucket(sender.clone(), bucket_id)?)
             }
-            // Handle case where the bucket has an existing collection.
-            (_, Some(current_collection_id))
+            // Handle case where the bucket has an existing collection, but the collection is not in stroage.
```
TDemeco, line 433:
```suggestion
            // Handle case where the bucket has an existing collection, but the collection is not in storage.
```

TDemeco (approved):

PR #273, opened 2024-11-26
fix: :bug: change payment stream update logic to delete if user is insolvent

TDemeco:
This PR changes the logic of updating payment streams. Now, whenever a call to update a payment stream is done, we check beforehand if the user is currently insolvent and if so, call the delete payment stream function instead.


pallets/file-system/src/tests.rs
```diff
@@ -7393,6 +7242,149 @@ mod delete_file_and_pending_deletions_tests {
...
+                );
+
+                // Assert that the pending file deletion request was removed from storage
+                assert_eq!(
+                    file_system::PendingFileDeletionRequests::<Test>::get(owner_account_id.clone()),
+                    BoundedVec::<_, <Test as file_system::Config>::MaxUserPendingDeletionRequests>::default()
+                );
+
+                // Assert that the payment stream was correctly deleted since the user is without funds
+                assert!(pallet_payment_streams::FixedRatePaymentStreams::<Test>::get(msp_id, owner_account_id.clone()).is_none());
```
snowmead, line 7380: how did the payment stream get automatically deleted?

TDemeco: The idea is that when a provider tries to charge a user that is deemed as without funds, the `manage_user_without_funds` logic gets executed, which basically charges the debt of the payment stream from the user deposit and then clears it. After this, the provider is free to delete without penalty any files from the user since the runtime knows that it is not getting paid for them

pallets/file-system/src/tests.rs
```diff
@@ -7393,6 +7242,149 @@ mod delete_file_and_pending_deletions_tests {
                 );
             });
         }
+
+        #[test]
+        fn delete_file_expired_pending_file_deletion_request_with_insolvent_user() {
```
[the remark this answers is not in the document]
TDemeco: Thanks! It's just the previous test with some changes hehe

pallets/file-system/src/tests.rs
snowmead: ok I see, it's not part of a separate process instead of having it automatically being deleted when we update the payment stream

pallets/payment-streams/src/utils.rs
```diff
@@ -202,25 +196,29 @@ where
...
-                "Payment stream should exist if it was found before.",
-                Error::<T>::PaymentStreamNotFound
-            );
-            payment_stream.rate = new_rate;
-            payment_stream.user_deposit = new_deposit;
-            Ok::<(), DispatchError>(())
-        })?;
+        // The payment stream could have been deleted when charged if the user was out of funds.
+        // If that didn't happen, we update it with the new rate here.
+        if FixedRatePaymentStreams::<T>::get(provider_id, user_account).is_some() {
```
ffarall, line 201: Before, in this same function, we have the check:
```rust
        // Check that a fixed-rate payment stream between the Provider and User exists
        ensure!(
            FixedRatePaymentStreams::<T>::contains_key(provider_id, user_account),
            Error::<T>::PaymentStreamNotFound
        );
```

Doesn't that mean that if the user doesn't have a payment stream with the provider, the function will fail before even reaching this point? And if so, if an MSP is trying to remove a file from an insolvent user for a second time, he won't be able to do it, because the check will fail.

If I'm correct, please add a test case for this scenario. Also if this is the case, one thing we could do is handle this specific situation in the File System pallet, where we would check if the payment stream exists before reducing the bucket's size. If it doesn't, we handle it and accept it.

ffarall (approved):
Good job!!

PR #275, opened 2024-11-27
feat: ✨ Detect re-orgs in block import pipeline of Blockchain Service

ffarall:


client/blockchain-service/src/utils.rs
```diff
@@ -107,6 +107,45 @@ impl BlockchainService {
         }
     }
 
+    pub(crate) fn register_best_block_and_check_reorg<Block>(
```
snowmead, line 110: is is it possible to write unit tests for this?
I think it would be worth it

snowmead (approved):

test/suites/integration/bsp/reorg-proof.test.ts
```diff
@@ -1,9 +1,9 @@
 import { notStrictEqual, strictEqual } from "node:assert";
-import { describeBspNet, type EnrichedBspApi } from "../../../util";
+import { ShConsts, describeBspNet, type EnrichedBspApi } from "../../../util";
 
 describeBspNet(
   "BSP proofs resubmitted on chain re-org ♻️",
-  { initialised: true, networkConfig: "standard" },
+  { initialised: true, networkConfig: "standard", only: true },
```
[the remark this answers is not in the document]
ffarall: Yes! Thanks!

client/blockchain-service/src/utils.rs
ffarall: Although I could write a unit test _specifically_ for this function, we don't have unit tests for Blockchain Service functions in general. Anyway, my goal is to test this alongside the full re-org behaviour in the integration tests. I include some tests for this already in this PR, but more will come as this feature progresses.

PR #276, opened 2024-11-28
Feat: MSP move bucket

links234:


node/src/tasks/msp_move_bucket.rs
```diff
@@ -0,0 +1,306 @@
...
+
+            let file_metadata = file.to_file_metadata(bucket.clone());
+            let file_key = file_metadata.file_key::<HashT<StorageProofsMerkleTrieLayout>>();
+
+            self.storage_hub_handler
+                .file_storage
+                .write()
+                .await
+                .insert_file(file_key.clone(), file_metadata.clone())
+                .unwrap();
```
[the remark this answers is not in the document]
snowmead: and I think you can remove the `.clone()` since `file_key` impls `Copy`

node/src/tasks/msp_move_bucket.rs
```diff
@@ -0,0 +1,306 @@
...
+                continue;
+            }
+
+            // We will cycle through all the BSP peer IDs for each chunk until we successfully
+            // download the file.
+            let mut bsp_peer_ids_iter = bsp_peer_ids.iter().cycle();
+
+            for chunk in 0..chunks_count {
+                for _ in 0..DOWNLOAD_REQUEST_RETRY_COUNT {
+                    let peer_id = bsp_peer_ids_iter.next().unwrap();
```
snowmead, line 191: Shouldn't we cycle on files instead of chunks. I think this would be more efficient to request all the chunks from a single BSP.

node/src/tasks/msp_move_bucket.rs
```diff
@@ -0,0 +1,306 @@
...
+                        Ok(download_request) => download_request,
+                        Err(error) => {
+                            error!(
+                                target: LOG_TARGET,
+                                "Failed to download chunk {:?} of file {:?} from peer {:?}: {:?}",
+                                chunk, file_key, peer_id, error
+                            );
+                            continue;
+                        }
+                    };
```
snowmead, line 214: Won't this fail quite often if we are requesting random BSPs for some data. Shouldn't we identify the BSPs that have this file first via the indexer and then request it from them?

node/src/tasks/msp_move_bucket.rs
```diff
@@ -0,0 +1,306 @@
...
+                target: LOG_TARGET,
+                "MSP: inserting downloaded file {:?} of bucket {:?} to forest storage",
+                file_key,
+                event.bucket_id,
+            );
+
+            if let Err(error) = forest_storage
+                .write()
+                .await
+                .insert_files_metadata(&[file_metadata])
```
snowmead, line 293: We should check that we downloaded successfully downloaded and inserted all chunks in file storage before adding the file to our forest no?

snowmead (changes requested):

client/file-transfer-service/src/handler.rs
```diff
@@ -52,6 +52,44 @@ use super::{
...
+}
+
+impl Ord for BucketIdWithExpiration {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        // Order by expiration first, then by bucket_id
+        self.expiration
+            .cmp(&other.expiration)
+            .then(self.bucket_id.cmp(&other.bucket_id))
+    }
+}
```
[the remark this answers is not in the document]
links234: I was not sure about the order of comparisons, I wanted it to be explicit by expiration first.

client/file-transfer-service/src/handler.rs
```diff
@@ -271,6 +319,52 @@ impl Actor for FileTransferService {
...
+
+                    match callback.send(result) {
+                        Ok(()) => {}
+                        Err(_) => error!(
+                            target: LOG_TARGET,
+                            "Failed to send the response back. Looks like the requester task is gone."
+                        ),
+                    }
+                }
+                FileTransferServiceCommand::UnregisterBucket {
```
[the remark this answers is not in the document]
links234: Sound good, but if we do, `grace_period_seconds` would probably need a rename too (?).

node/src/service.rs
```diff
@@ -338,9 +340,16 @@ where
...
 
-    if indexer_options.is_some() {
+    if indexer_config.indexer {
         let task_spawner = TaskSpawner::new(task_manager.spawn_handle(), "indexer-service");
-        spawn_indexer_service(&task_spawner, client.clone(), maybe_db_pool.unwrap()).await;
+        spawn_indexer_service(
+            &task_spawner,
+            client.clone(),
+            maybe_db_pool.clone().expect(
+                "Indexer is enabled but no database URL is provided (via CLI using --database-url or setting DATABASE_URL environment variable)",
```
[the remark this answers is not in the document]
links234: No because we allow to provide the database URL via environment variable DATABASE_URL.

node/src/tasks/msp_move_bucket.rs
```diff
@@ -0,0 +1,306 @@
...
+            .blockchain
+            .send_extrinsic(call, Tip::from(0))
+            .await?
+            .with_timeout(Duration::from_secs(
+                self.storage_hub_handler
+                    .provider_config
+                    .extrinsic_retry_timeout,
+            ))
+            .watch_for_success(&self.storage_hub_handler.blockchain)
+            .await?;
```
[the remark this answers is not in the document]
links234: Agree. I left this and other not-happy paths for the next PR!

node/src/tasks/msp_move_bucket.rs
links234: I was thinking more about distributing the load to the BSPs than getting the file as fast as possible. Anyway - the fastest transfer speeds would be achieved if we parallelize this (have multiple inflight download requests at the same time). This way you can actually saturate the download bandwidth with multiple uploaders.

node/src/tasks/msp_move_bucket.rs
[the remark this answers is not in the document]
links234: I can change it to an `expect` but I don't really know how you can recover from that error otherwise.

node/src/tasks/msp_move_bucket.rs
links234: That's what we are doing. We query the indexer for all the files in a bucket, than for each file we query all the peer ids of BSPs storing that file and we cycle through them.

client/file-transfer-service/src/handler.rs
```diff
@@ -503,4 +652,33 @@ impl FileTransferService {
...
+    fn handle_expired_buckets(&mut self) {
+        let now = chrono::Utc::now();
+        while let Some(bucket) = self.bucket_allow_list_grace_period_time.pop_first() {
+            if bucket.expiration < now {
+                let _ = self.unregister_bucket(bucket.bucket_id).map_err(|e| {
+                    error!(target: LOG_TARGET, "Failed to unregister expired bucket {:?}: {:?}", bucket.bucket_id, e);
+                });
+            } else {
+                // Re-insert the bucket back into the set since it hasn't expired yet.
+                self.bucket_allow_list_grace_period_time.insert(bucket);
```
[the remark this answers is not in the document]
links234: No. In lack of a "`peek_first`" functionality we use `pop_first` which removes the item from the set everytime - including the last item which is supposed to fail the expiration check (meaning it should be kept).

snowmead (approved):
beast

client/common/src/blockchain_utils.rs
```diff
@@ -57,21 +57,25 @@ pub fn get_events_at_block(
 pub fn convert_raw_multiaddresses_to_multiaddr(multiaddresses: Multiaddresses) -> Vec<Multiaddr> {
```
ffarall, line 57:
```suggestion
pub fn convert_raw_multiaddresses_to_multiaddr_vec(multiaddresses: Multiaddresses) -> Vec<Multiaddr> {
```

test/util/netLaunch/index.ts
```diff
@@ -153,6 +153,14 @@ export class NetworkLauncher {
...
         "--database-url=postgresql://postgres:postgres@docker-sh-postgres-1:5432/storage_hub"
       );
+      if (this.type === "fullnet") {
+        composeYaml.services["sh-msp-1"].command.push(
+          "--database-url=postgresql://postgres:postgres@docker-sh-postgres-1:5432/storage_hub"
+        );
+        composeYaml.services["sh-msp-2"].command.push(
+          "--database-url=postgresql://postgres:postgres@docker-sh-postgres-1:5432/storage_hub"
+        );
+      }
```
ffarall, line 163: We may want to add this just in the yml files directly. I don't think we need to do it dynamically, as MSPs will always need that indexer connection.

test/util/bspNet/waits.ts
```diff
@@ -323,30 +310,47 @@ export const waitForBspToCatchUpToChainTip = async (
...
+  lambda: () => Promise<boolean>;
+  iterations?: number;
+  delay?: number;
+}
+
+/**
+ * Waits for an arbitrary condition to be true. It keeps polling the condition until it is true or
+ * a timeout is reached.
+ */
+export const waitFor = async (options: WaitForOptions) => {
```
ffarall, line 338: REALLY LIKE THIS. Much needed.

node/src/tasks/bsp_download_file.rs
```diff
@@ -51,8 +51,31 @@ where
...
+        if let Some(bucket_id) = bucket_id {
+            if file_metadata.bucket_id != bucket_id.as_ref().to_vec() {
+                error!(
+                    target: LOG_TARGET,
+                    "File bucket mismatch for file {:?}: expected {:?}, got {:?}",
+                    event.file_key, file_metadata.bucket_id, bucket_id
+                );
+                return Err(anyhow::anyhow!("File bucket mismatch"));
+            }
+        }
```
ffarall, line 77: This task could use some docs.

node/src/tasks/bsp_move_bucket.rs
```diff
@@ -0,0 +1,184 @@
...
+
+use shc_actors_framework::event_bus::EventHandler;
+use shc_blockchain_service::{
+    commands::BlockchainServiceInterface,
+    events::{MoveBucketAccepted, MoveBucketExpired, MoveBucketRejected, MoveBucketRequested},
+};
+use shc_file_transfer_service::commands::FileTransferServiceInterface;
+
+use crate::services::handler::StorageHubHandler;
+use crate::tasks::{BspForestStorageHandlerT, FileStorageT};
```
ffarall, line 12: Merge

node/src/tasks/msp_move_bucket.rs
```diff
@@ -0,0 +1,312 @@
...
+use shc_blockchain_service::{
+    commands::BlockchainServiceInterface, events::MoveBucketRequestedForNewMsp,
+};
+use shc_common::types::{FileKeyProof, HashT, StorageProofsMerkleTrieLayout};
+use shc_file_transfer_service::commands::FileTransferServiceInterface;
+use shc_forest_manager::traits::ForestStorage;
+use shp_file_metadata::ChunkId;
+
+use crate::services::handler::StorageHubHandler;
+use crate::tasks::{FileStorageT, MspForestStorageHandlerT};
```
ffarall, line 16: merge and move to the bottom.

node/src/tasks/bsp_move_bucket.rs
```diff
@@ -0,0 +1,184 @@
...
+/// Handles the [`MoveBucketRequested`] event.
+///
+/// This event is triggered when an user requests to move a bucket to a new MSP.
+/// As a BSP, we need to allow the new MSP to download the files we have from the bucket.
+impl<FL, FSH> EventHandler<MoveBucketRequested> for BspMoveBucketTask<FL, FSH>
+where
+    FL: FileStorageT,
+    FSH: BspForestStorageHandlerT,
+{
+    async fn handle_event(&mut self, event: MoveBucketRequested) -> anyhow::Result<()> {
```
ffarall, line 61: Shouldn't this check if this BSP even has files from that bucket?

ffarall: I mean it's not mandatory, but it would be a nice to have so that we don't clutter the allow list. Anyway, there shouldn't be that many bucket changes ideally.

node/src/tasks/msp_move_bucket.rs
```diff
@@ -0,0 +1,312 @@
...
+                target: LOG_TARGET,
+                "Indexer is disabled but a move bucket event was received. Please provide a database URL (and enable indexer) for it to use this feature."
+            );
+
+            let call = storage_hub_runtime::RuntimeCall::FileSystem(
+                pallet_file_system::Call::msp_respond_move_bucket_request {
+                    bucket_id: event.bucket_id,
+                    response: BucketMoveRequestResponse::Rejected,
+                },
+            );
```
ffarall, line 94: This could use a bit more comments

node/src/tasks/msp_move_bucket.rs
```diff
@@ -0,0 +1,312 @@
...
+                        .extrinsic_retry_timeout,
+                ))
+                .watch_for_success(&self.storage_hub_handler.blockchain)
+                .await?;
+
+            return Ok(());
+        };
+
+        // TODO: check that we have enough space to accept the bucket and reject if not (+test)
+        // TODO: check that are know all the BSPs for the files in the bucket and reject if not (+test)
```
ffarall, line 112: @santikaplan let's map these tasks, likely for @undercover-cactus.

node/src/tasks/msp_move_bucket.rs
```diff
@@ -0,0 +1,312 @@
...
+
+        let mut indexer_connection = indexer_db_pool.get().await?;
+
+        let forest_storage = self
+            .storage_hub_handler
+            .forest_storage_handler
+            .get_or_create(&event.bucket_id.as_ref().to_vec())
+            .await;
+
+        for file in shc_indexer_db::models::File::get_by_onchain_bucket_id(
```
ffarall, line 149: Is this downloading one chunk per fail at a time? Shouldn't we optimise this so that it sends download requests to many BSPs for all the chunks of all the files at once?

ffarall: I know that could get cumbersome to handle errors then, but it would be infinitely faster.

ffarall: Anyway, if this works for now, it's fine, but we should consider this as an important improvement.

test/suites/integration/msp/move-bucket.test.ts
```diff
@@ -0,0 +1,312 @@
...
+      }
+    });
+
+    it("User moves bucket to second MSP", async () => {
+      const requestMoveBucketResult = await userApi.sealBlock(
+        userApi.tx.fileSystem.requestMoveBucket(bucketId, msp2Api.shConsts.DUMMY_MSP_ID_2),
+        shUser
+      );
+
+      await assertEventPresent(
```
ffarall, line 277:
```suggestion
      assertEventPresent(
```

test/suites/integration/msp/move-bucket.test.ts
```diff
@@ -0,0 +1,312 @@
...
+      );
+
+      await userApi.wait.waitForTxInPool({
+        module: "fileSystem",
+        method: "mspRespondMoveBucketRequest"
+      });
+
+      const { events } = await userApi.sealBlock();
+
+      await assertEventPresent(userApi, "fileSystem", "MoveBucketAccepted", events);
```
ffarall, line 291:
```suggestion
      assertEventPresent(userApi, "fileSystem", "MoveBucketAccepted", events);
```

ffarall (approved):
Remarkable work. I left many comments but they're mostly nitpicks

client/file-transfer-service/src/handler.rs
```diff
@@ -503,4 +650,33 @@ impl FileTransferService {
...
+                break;
+            }
+        }
```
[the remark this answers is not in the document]
ffarall:
```suggestion
        // Return early if there are no buckets to unregister.
        if self.bucket_allow_list_grace_period_time.is_empty() {
            return;
        }

        // Get the current time.
        let now = chrono::Utc::now();

        // At this point we know there must be at least one bucket in the allow list.
        let mut bucket_to_check = self.bucket_allow_list_grace_period_time.first();
        while bucket_to_check.map_or(false, |bucket| bucket.expiration < now) {
            // Remove the bucket from the allow list.
            let bucket_to_remove = self
                .bucket_allow_list_grace_period_time
                .pop_first()
                .expect("Bucket allow list is not empty; qed");

            // Try to unregister the bucket.
            if let Err(e) = self.unregister_bucket(bucket_to_remove.bucket_id) {
                error!(target: LOG_TARGET, "Failed to unregister expired bucket {:?}: {:?}", bucket_to_remove.bucket_id, e);
            }

            // Update the expiration to check.
            bucket_to_check = self.bucket_allow_list_grace_period_time.first();
        }
```

client/indexer-db/migrations/2024-09-26-132546_create_multiaddress/up.sql
```diff
@@ -1,6 +1,6 @@
 CREATE TABLE multiaddress (
     id SERIAL PRIMARY KEY,
-    address VARCHAR NOT NULL,
+    address BYTEA NOT NULL,
```
undercover-cactus, line 3: I like this change

test/suites/integration/msp/move-bucket.test.ts
```diff
@@ -0,0 +1,312 @@
...
+    let bucketId: string;
+    const allBucketFiles: string[] = [];
+
+    before(async () => {
+      userApi = await createUserApi();
+      const maybeMsp1Api = await createMsp1Api();
+      if (maybeMsp1Api) {
+        msp1Api = maybeMsp1Api;
+      } else {
+        throw new Error("MSP API for first MSP not available");
```
undercover-cactus, line 36: We could use `assert` here instead of throw

undercover-cactus (approved):

client/common/src/blockchain_utils.rs
```diff
@@ -57,21 +57,25 @@ pub fn get_events_at_block(
...
-                }
-            },
+        if let Some(multiaddress) = convert_raw_multiaddress_to_multiaddr(&raw_multiaddr) {
+            multiaddress_vec.push(multiaddress);
+        }
+    }
+    multiaddress_vec
+}
+
+pub fn convert_raw_multiaddress_to_multiaddr(raw_multiaddr: &[u8]) -> Option<Multiaddr> {
```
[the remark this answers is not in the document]
links234: What would the behaviour be, when would there be an error? Right now `convert_raw_multiaddresses_to_multiaddr` returns back whatever is valid. I agree `convert_raw_multiaddress_to_multiaddr` could be changed from `Option` to `Result` but not the other (which instead could be refactored out somehow).

client/blockchain-service/src/handler.rs
```diff
@@ -1248,6 +1250,39 @@ impl BlockchainService {
...
+                                new_msp_id,
+                            });
+                            if self.provider_ids.contains(&new_msp_id) {
+                                self.emit(MoveBucketRequestedForNewMsp { bucket_id });
+                            }
+                        }
+                        RuntimeEvent::FileSystem(
+                            pallet_file_system::Event::MoveBucketRejected { bucket_id, msp_id },
+                        ) => {
+                            self.emit(MoveBucketRejected { bucket_id, msp_id });
```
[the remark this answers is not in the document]
links234: If there is no subscriber - nothing really, they get dropped.

node/src/tasks/msp_move_bucket.rs
links234: Yes, that was the plan. Wrote it so that we can change it easily later, but didn't want to throw it in right now. Let's map this into a task as well.

client/indexer-db/migrations/2024-09-26-132546_create_multiaddress/up.sql
links234: We should do it everywhere tbh.

client/indexer-db/src/models/file.rs
```diff
@@ -2,8 +2,12 @@ use chrono::NaiveDateTime;
 use diesel::prelude::*;
 use diesel_async::RunQueryDsl;
 
+use sc_network::{Multiaddr, PeerId};
+
+use shc_common::types::{FileMetadata, Fingerprint};
```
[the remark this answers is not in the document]
links234: To be super extra pedantic :D, I left the separation between substrate specific (`sc_network`) and internal crates intentional - if this is not the convention we want let me know and I'll change it.

client/indexer-db/Cargo.toml
```diff
@@ -25,3 +25,7 @@ diesel = { workspace = true }
 diesel-async = { workspace = true }
 thiserror = { workspace = true }
 bigdecimal = { workspace = true }
+
+sc-network = { workspace = true }
+
+shc-common = { workspace = true }
```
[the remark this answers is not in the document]
links234: Same here.

test/util/netLaunch/index.ts
links234: This would break for tests where we don't need the indexer but use MSPs - we can not support that case and always have an indexer instead.

test/util/bspNet/waits.ts
links234: Thanks. Eventually I did a drive-by refactor in the `waits.ts` since it was bugging me too much.

node/src/tasks/bsp_move_bucket.rs
links234: In order to do that we need to have an index from `BucketId` to `FileKey`s. This is 1 of the reasons why I opted to have allowlists for buckets and let the BSP check the `FileKey,BucketId` requested pair. To fix just this we can also add an index just for all `BucketId`s stored.

PR #277, opened 2024-11-28
feat: delete file task

undercover-cactus:
This PR add the file deletion task so that the calculated root is correct.

In this PR we also removed ´tiny-invariant´ called from the test and remove completly the dependencies in favor of the native assert.


test/suites/integration/msp/stop-storing-bucket.test.ts
```diff
@@ -1,6 +1,5 @@
-import { strictEqual } from "node:assert";
+import assert, { strictEqual } from "node:assert";
 import { describeMspNet, mspKey, sleep, type EnrichedBspApi } from "../../../util";
-import invariant from "tiny-invariant";
```
ffarall, line 3: Much appreciated change! Should we also remove the dependency from `package.json`?

test/suites/integration/bsp/submit-proofs.test.ts
```diff
@@ -13,7 +13,7 @@ import { BSP_THREE_ID, BSP_TWO_ID, DUMMY_BSP_ID, NODE_INFOS } from "../../../uti
 
 describeBspNet(
   "BSP: Many BSPs Submit Proofs",
-  { initialised: "multi", networkConfig: "standard" },
+  { initialised: "multi", networkConfig: "standard", only: true },
```
ffarall, line 16: Remove. Probably leftover from debugging.

test/suites/integration/bsp/submit-proofs.test.ts
```diff
@@ -195,48 +195,45 @@ describeBspNet(
...
-            "fileSystem",
-              "BspConfirmStoppedStoring",
-            fileDeletionConfirmResult.events
-          );
+      const confirmStopStoringEvent = await userApi.assert.eventPresent(
+        "fileSystem",
+        "BspConfirmStoppedStoring"
+      );
+      // wait for line in docker logs
+      await sleep(2000);
```
ffarall, line 224: Did you mean to change this for a wait for some log in docker logs?

test/suites/integration/bsp/submit-proofs.test.ts
```diff
@@ -195,48 +195,45 @@ describeBspNet(
...
+        null,
+        [fileMetadata.fileKey]
+      );
+      // Wait enough blocks for the deletion to be allowed.
+      const currentBlock = await userApi.rpc.chain.getBlock();
+      const currentBlockNumber = currentBlock.block.header.number.toNumber();
+      const cooldown =
+        currentBlockNumber + bspThreeApi.consts.fileSystem.minWaitForStopStoring.toNumber();
+      await userApi.block.skipTo(cooldown);
+      // TODO: Confirm the request of deletion. Make sure the extrinsic doesn't fail and the root is updated correctly.
```
ffarall, line 210:
```suggestion
      // Confirm the request of deletion. Make sure the extrinsic doesn't fail and the root is updated correctly.
```

test/suites/integration/bsp/submit-proofs.test.ts
```diff
@@ -195,48 +195,45 @@ describeBspNet(
...
-      const newRoot = bspThreeApi.rpc.storagehubclient.getForestRoot();
+      const newRoot = (await bspThreeApi.rpc.storagehubclient.getForestRoot(null)).unwrap();
+      assert(userApi.events.fileSystem.BspConfirmStoppedStoring.is(confirmStopStoringEvent.event));
       const newRootInRuntime = confirmStopStoringEvent.event.data.newRoot;
-      assert(newRoot === newRootInRuntime, "The new root should be updated correctly");
-    */
-      }
-    );
+
+      // Important! Keep the string conversion to avoid a recursive call that lead to a crash in javascript.
```
ffarall, line 230: Good comment

test/suites/integration/bsp/submit-proofs.test.ts
```diff
@@ -195,48 +195,45 @@ describeBspNet(
...
-    const currentBlockNumber = currentBlock.block.header.number.toNumber();
-    const cooldown = currentBlockNumber + bspThreeApi.consts.fileSystem.minWaitForStopStoring.toNumber();
-    await bspThreeApi.block.skipTo(cooldown); */
-        // TODO: Confirm the request of deletion. Make sure the extrinsic doesn't fail and the root is updated correctly.
-        /*  const fileDeletionConfirmResult = bspThreeApi.sealBlock(bspThreeApi.tx.fileSystem.bspConfirmStopStoring(
-        fileMetadata.fileKey,
-        inclusionForestProof,
-      ));
+    it("BSP can correctly delete a file from its forest and runtime correctly updates its root", async () => {
+      // TODO: Setup a BSP that has two files which lie under the same NibbledBranch in the forest.
```
ffarall, line 199: This comment was probably leftover from something else we wanted to test. We should probably remove it or put it somewhere else so that it doesn't confuse future readers.

node/src/tasks/bsp_delete_file.rs
```diff
@@ -0,0 +1,150 @@
...
+            event.file_key,
+            event.bsp_id
+        );
+
+        // Remove the file from the forest.
+        self.remove_file_from_forest(&event.file_key.into()).await?;
+
+        info!(
+            target: LOG_TARGET,
+            "File {:?} successfuly removed from forest",
```
ffarall, line 102:
```suggestion
            "File {:?} successfully removed from forest",
```

node/src/tasks/bsp_delete_file.rs
```diff
@@ -0,0 +1,150 @@
...
+            .await
+            .ok_or_else(|| anyhow!("Failed to get forest storage."))?;
+        if read_fs
+            .read()
+            .await
+            .contains_file_key(&event.file_key.into())?
+        {
+            warn!(
+                target: LOG_TARGET,
+                "FinalisedBspConfirmStoppedStoring applied and finalised for file key {:?}, but file key is still in Forest. This can only happen if the same file key was added again after deleted by the user.",
```
ffarall, line 140:
```suggestion
                "FinalisedBspConfirmStoppedStoring applied and finalised for file key {:?}, but file key is still in Forest. This can only happen if the same file key was added again after deleted by this BSP.",
```

ffarall: In this case is this same BSP who's deleting the file.

node/src/services/handler.rs
```diff
@@ -291,5 +292,22 @@ where
             .clone()
             .subscribe_to(&self.task_spawner, &self.blockchain);
         sp_stop_storing_insolvent_user_event_bus_listener.start();
+
+        // Delete file and update the root.
```
ffarall, line 296: Probably worth a bit more description here. Like the fact that this is the task that removes a file as a consequence of this BSP requesting to stop storing it.

ffarall (approved):
Great job! Just a few nitpicks that I trust you'll take care of.

test/suites/integration/bsp/submit-proofs.test.ts
undercover-cactus: Oh yes I forgot!

node/src/tasks/bsp_delete_file.rs
```diff
@@ -0,0 +1,150 @@
...
+{
+    pub fn new(storage_hub_handler: StorageHubHandler<FL, FSH>) -> Self {
+        Self {
+            storage_hub_handler,
+        }
+    }
+
+    async fn remove_file_from_forest(&self, file_key: &H256) -> anyhow::Result<()> {
+        // Remove the file key from the Forest.
+        // Check that the new Forest root matches the one on-chain.
```
snowmead, line 46: I believe you don't check that the forest root matches the one on chain.

But since this task is async to removing the file from the forest in the runtime, there could be a scenario where there is multiple files being removed but this task is executed in a different order from the order on-chain. 

For example:
- on-chain > delete file 1
- on-chain > delete file 2
- task > remove file 1 from forest > fails check since forest root on chain is the result of removing both files
- task > remove file 2 from forest > succeeds

node/src/tasks/bsp_delete_file.rs
```diff
@@ -0,0 +1,150 @@
...
+        {
+            let fs = self
+                .storage_hub_handler
+                .forest_storage_handler
+                .get(&NoKey)
+                .await
+                .ok_or_else(|| anyhow!("Failed to get forest storage."))?;
+
+            fs.write().await.delete_file_key(file_key).map_err(|e| {
+                error!(target: LOG_TARGET, "CRITICAL❗️❗️ Failed to apply mutation to Forest storage. This may result in a mismatch between the Forest root on-chain and in this node. \nThis is a critical bug. Please report it to the StorageHub team. \nError: {:?}", e);
```
snowmead, line 56: This is not a critical error due to the example explained in previous comment.

test/suites/integration/msp/stop-storing-bucket.test.ts
undercover-cactus: Not yet. I think there is still some in the utils functions.

node/src/tasks/bsp_delete_file.rs
undercover-cactus: I have removed the comment and indeed I don't check the Forest root.

I am working on an other PR for an integration test that test multiple file deletion for a BSP. I will check those scenarios.

PR #280, opened 2024-11-29
feat: ✨ Add `snapshot` functionality to `ForestStorageHandler`

ffarall:
This PR:
1. Adds the `snapshot` function to the `ForestStorageHandler` trait, allowing to create copies of existing Forests.
2. Implements it for the existing structs that implement `ForestStorageHandler`.
    1. To be precise, this is only relevant for `ForestStorageCaching`, because `ForestStorageSingle` wouldn't be able to manage more than the one original Forest.

Testing is not included in this PR. This functionality will be tested and debug in subsequent PRs that use this new feature.

### DIFF
--- a/client/forest-manager/src/rocksdb.rs
+++ b/client/forest-manager/src/rocksdb.rs
@@ -7,7 +7,11 @@ use sp_state_machine::{warn, Storage};
 use sp_trie::{
     prefixed_key, recorder::Recorder, PrefixedMemoryDB, TrieDBBuilder, TrieLayout, TrieMut,
 };
-use std::{io, path::PathBuf, sync::Arc};
+use std::{
+    fs, io,
+    path::{Path, PathBuf},
+    sync::Arc,
+};
 use trie_db::{DBValue, Trie, TrieDBMutBuilder};
 
 use crate::{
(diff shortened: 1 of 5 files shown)


client/forest-manager/src/rocksdb.rs
```diff
@@ -26,6 +30,44 @@ pub(crate) fn other_io_error(err: String) -> io::Error {
...
+pub fn copy_db<T>(
+    src: String,
+    dest: String,
+) -> Result<StorageDb<T, kvdb_rocksdb::Database>, ErrorT<T>>
+where
+    T: TrieLayout,
+    HasherOutT<T>: TryFrom<[u8; 32]>,
+{
+    let src_path: &Path = Path::new(&src);
+    let dest_path: &Path = Path::new(&dest);
```
undercover-cactus, line 59: It is a bit of an unusual way to initiate.

Do this instead:
```rust
 let src_path: Path = Path::new(&src);
 let dest_path: Path = Path::new(&dest);
 
 copy_dir_all(&src_path, &dest_path).map_err(|e| {
 ....
```

snowmead (approved):
nice job! 

I think we should remove the `ForestStorageSingle` implementation since we no longer use it for any specific role in the network.

node/src/services/forest_storage.rs
```diff
@@ -256,4 +302,42 @@ where
...
+        }
+
+        let storage_path = self
+            .storage_path
+            .clone()
+            .expect("Storage path should be set");
+        let src = format!("{}_{:?}", storage_path, src_key);
+        let dest = format!("{}_{:?}", storage_path, dest_key);
+
+        let underlying_db = match rocksdb::copy_db(src, dest) {
```
links234, line 326: I'm pretty sure this is not 100% correct and that it will be very hard for it to surface. I suggest we guard lock operations on the `src` in order to guarantee no writes from other threads.

links234 (changes requested):
While this can definitely be better integrated with RocksDB native snapshots (not sure if/how this is exposed) I think this is good enough, not a big performance issue and easy to change.
However I left a comment which should be addressed (either double check that it is not the case or fix).

links234 (approved):

PR #285, opened 2024-12-12
feat: Pallet file system benchmarking setup

snowmead:
Added benchmarking setup for the file system pallet and benchmarked `issue_storage_request` and `create_bucket`


pallets/file-system/src/benchmarking.rs
```diff
@@ -0,0 +1,165 @@
...
+            .try_into()
+            .unwrap();
+        let bucket_id = <<T as crate::Config>::Providers as ReadBucketsInterface>::derive_bucket_id(
+            &user,
+            name.clone(),
+        );
+        let location = vec![1; MaxFilePathSize::<T>::get().try_into().unwrap()]
+            .try_into()
+            .unwrap();
+        let fingerprint = <<T as frame_system::Config>::Hashing as Hasher>::hash(b"tes");
```
TDemeco, line 73:
```suggestion
        let fingerprint = <<T as frame_system::Config>::Hashing as Hasher>::hash(b"benchmark_fingerprint");
```

pallets/file-system/src/benchmarking.rs
```diff
@@ -0,0 +1,165 @@
...
+        #[extrinsic_call]
+        _(
+            signed_origin,
+            bucket_id,
+            location,
+            fingerprint,
+            size,
+            None,
+            peer_ids,
+        );
```
TDemeco, line 89: Worst-case scenario I believe is:
- With a MSP (since there are extra checks that the pallet has to do)
- With a full list of peer IDs (since the logic clones the list)

TDemeco (approved):
Setup looks good! Just some nitpicks on one benchmark

PR #287, opened 2024-12-12
fix: :bug: change update order of payment stream's rate and bucket size

TDemeco:
This PR fixes a bug where sometimes when increasing the size of a bucket (because the MSP that stores it accepts to store a new file for it from the user that owns it) the update of the rate of the fixed-rate payment stream between that MSP and that user was done incorrectly and ended up being greater than expected.

This was happening because the function that updates the rate of a fixed-rate payment stream (when updating the size of a bucket) expects that bucket's size as recorded in storage to not have been updated yet, since it uses the size of the bucket (previous to the update) and its delta to calculate the difference in rate, which then uses to update the rate of the payment stream to its new value. This expectation was not met, since the logic that increases a bucket's size was first updating it on the buckets' storage and then calling the aforementioned funcion.

This resulted in a scenario where the new rate of the payment stream was calculated as:
`new_payment_stream_rate = previous_rate_of_bucket + (2 * rate_delta_because_of_new_file - rate_delta_because_of_new_file)`

To the unsuspecting eye this calculation seems correct, and that's why everything was working correctly, BUT in the specific scenario where the calculation of `rate_delta_because_of_new_file` had a fractional part that was over 0.25 but lower than 0.5, the rounding of `rate_delta_because_of_new_file` would be down (since its fractional part was lower than 0.5) and the rounding of `2 * rate_delta_because_of_new_file` would be up (since its fractional part would be over 0.5)! This made it so in some scenarios the resulting new rate was one higher than the amount expected.

With the introduced changes, this calculation has now been changed to:
`new_payment_stream_rate = previous_rate_of_bucket + rate_delta_because_of_new_file`
avoiding the error completely.

As a bonus I did some changes to the test that was checking this to make it a bit more clear and explicit, avoiding the issue we were having where the issue was hard to reproduce.

### DIFF
--- a/pallets/providers/src/utils.rs
+++ b/pallets/providers/src/utils.rs
@@ -1732,27 +1732,29 @@ impl<T: pallet::Config> MutateBucketsInterface for pallet::Pallet<T> {
         bucket_id: &Self::BucketId,
         delta: Self::StorageDataUnit,
     ) -> DispatchResult {
-        let (msp_id, user_id) = Buckets::<T>::try_mutate(&bucket_id, |maybe_bucket| {
+        Buckets::<T>::try_mutate(&bucket_id, |maybe_bucket| {
             let bucket = maybe_bucket.as_mut().ok_or(Error::<T>::BucketNotFound)?;
 
+            // Get the MSP ID and the user owner of the bucket
+            let msp_id = bucket
+                .msp_id
+                .ok_or(Error::<T>::BucketMustHaveMspForOperation)?;
+            let user_id = bucket.user_id.clone();
+
+            // First, try to update the fixed rate payment stream with the new rate, since
+            // this function uses the current bucket size to calculate it
+            Self::apply_delta_fixed_rate_payment_stream(
+                &msp_id,
+                bucket_id,
+                &user_id,
+                RateDeltaParam::Increase(delta),
+            )?;
+
+            // Then, if that was successful, update the bucket size
             bucket.size = bucket.size.saturating_add(delta);
 
-            Ok::<_, DispatchError>((
-                bucket
-                    .msp_id
-                    .ok_or(Error::<T>::BucketMustHaveMspForOperation)?,
-                bucket.user_id.clone(),
-            ))
-        })?;
-
-        Self::apply_delta_fixed_rate_payment_stream(
-            &msp_id,
-            bucket_id,
-            &user_id,
-            RateDeltaParam::Increase(delta),
-        )?;
-
-        Ok(())
+            Ok::<_, DispatchError>(())
+        })
     }
 
     fn decrease_bucket_size(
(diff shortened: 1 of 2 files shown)


undercover-cactus:
Nicely done!

undercover-cactus (approved):

test/suites/integration/msp/debt-collection.test.ts
```diff
@@ -145,22 +176,47 @@ describeMspNet("Single MSP collecting debt", ({ before, createMspApi, it, create
...
+    const firstFileSize = (
+      await mspApi.rpc.storagehubclient.getFileMetadata(bucketOption, acceptedFileKey)
+    )
+      .unwrap()
+      .file_size.toNumber();
+    const unitsInGigaUnit = 1024 * 1024 * 1024;
+    let expectedPaymentStreamRate =
+      (valueProps[0].value_prop.price_per_giga_unit_of_data_per_block.toNumber() * firstFileSize) /
+        unitsInGigaUnit +
+      zeroSizeBucketFixedRate;
```
snowmead, line 208: note: it would be nice to have an RPC method for users to be able call and return the cost of adding a file with some size to some bucket. then dApps can expose this information when users are about to add files via some dashboard.

snowmead (approved):

pallets/providers/src/utils.rs
```diff
@@ -1732,27 +1732,29 @@ impl<T: pallet::Config> MutateBucketsInterface for pallet::Pallet<T> {
...
-        let (msp_id, user_id) = Buckets::<T>::try_mutate(&bucket_id, |maybe_bucket| {
+        Buckets::<T>::try_mutate(&bucket_id, |maybe_bucket| {
             let bucket = maybe_bucket.as_mut().ok_or(Error::<T>::BucketNotFound)?;
 
-            bucket.size = bucket.size.saturating_add(delta);
+            // Get the MSP ID and the user owner of the bucket
+            let msp_id = bucket
+                .msp_id
+                .ok_or(Error::<T>::BucketMustHaveMspForOperation)?;
+            let user_id = bucket.user_id.clone();
```
ffarall, line 1742: Why when increasing we error out if there is no MSP, but when decreasing we don't error out, but just don't update the payment stream. It seems like the behaviour of `decrease_bucket_size` is the correct one, since there could be a bucket without an MSP, but if that is the case, we just don't update the payment stream.

ffarall (approved):
Approved with comment

pallets/providers/src/utils.rs
TDemeco: I might be wrong since I'm not the author of that logic, but I believe the idea behind this implementation was:
- A bucket can **_decrease_** in size if its owner submits an extrinsic to delete a specific file of that bucket and, in that extrinsic, includes a forest inclusion proof that proves that the file is part of that bucket. This flow can happen even if the bucket does not currently have an MSP and we shouldn't error out.
- On the other hand, the only way for a bucket to **_increase_** in size is for its MSP to accept a storage request, so increasing a bucket's size without it being currently stored by an MSP is an error, it should never happen.

TDemeco: @ffarall just implemented the change that we talked about in the daily: buckets can now increase in size without having a MSP (or at least it's no longer enforce at the Providers level, the File System should be the one enforcing that if needed). Will merge when CI passes

PR #290, opened 2024-12-14
fix: 🩹 Stop retries if proof is outdated and integration test fix

ffarall:
In this PR:
1. Fix race condition in `multiple-delete.test.ts`.
2. Prevent submit proof retries if proof is outdated.
3. Improve some logs.


node/src/tasks/bsp_submit_proof.rs
```diff
@@ -268,21 +310,23 @@ where
...
-                    "Failed to submit proof after {} attempts",
-                    MAX_PROOF_SUBMISSION_ATTEMPTS
-                )
+                error!(target: LOG_TARGET, "❌ Failed to submit proof due to: {}", e);
+                anyhow!("Failed to submit proof due to: {}", e)
             })?;
 
         trace!(target: LOG_TARGET, "Proof submitted successfully");
 
+        // TODO: Don't do this in this task any more.
```
TDemeco, line 329: Where would you want to do it?

node/src/tasks/bsp_submit_proof.rs
```diff
@@ -246,17 +246,59 @@ where
...
+                        .is_err();
+                let has_forest_root_changed = {
+                    let fs = cloned_sh_handler
+                        .forest_storage_handler
+                        .get(&current_forest_key)
+                        .await;
+
+                    match fs {
+                        Some(fs) => fs.read().await.root() != *cloned_forest_root,
+                        None => true,
```
TDemeco, line 287: Here `fs` might be none because the forest key has changed after a reorg? Or why?

test/suites/integration/bsp/multiple-delete.test.ts
```diff
@@ -81,6 +81,9 @@ describeBspNet("Single BSP Volunteering", ({ before, createBspApi, it, createUse
     await sleep(500);
     await userApi.wait.bspStored(1);
 
+    // Wait for BSP to update its local Forest root before starting to generate the inclusion proofs
+    await sleep(2000);
```
TDemeco, line 85: We could maybe have a `wait` that waits for certain file keys to be in the forest storage of the BSP (which should be enough to guarantee that the forest root has been updated)

TDemeco (approved):
Not much to say, love the added comments and most of all the emojis on the logs 😎 just left some questions for myself and a suggestion

node/src/tasks/bsp_submit_proof.rs
ffarall: It'll be done in every block, in the block import. We will run a `catch_up_to_forest_root_changes` (or some name like that).

PR #291, opened 2024-12-16
feat: stop challenge cycle if BSP doesn't have files anymore

undercover-cactus:
In this PR we handle the case where the BSP doesn't store file anymore and we are required to stop the challenge cycle so no proof are requested to the BSP.

We also fix the test in submit-proof.test.ts that didn't correctly handle this case.


pallets/proofs-dealer/src/utils.rs
```diff
@@ -1046,6 +1046,39 @@ impl<T: pallet::Config> ProofsDealerInterface for Pallet<T> {
         )
     }
 
+    // Remove a provider from challenge cycle.
+    fn stop_challenge_cycle(provider_id: &Self::ProviderId) -> DispatchResult {
```
snowmead, line 1050: I would add unit tests for this.

snowmead, line 1050: This should also be called when the proofs-dealer applies a remove mutation and as a result the forest root would be the default one.

snowmead (changes requested):

pallets/proofs-dealer/src/utils.rs
undercover-cactus: Can you point me where this is happening ?

snowmead: The mutation is applied in the `do_submit_proof` function. At some point in the execution we should check that the forest root is the default one, and if it is, we stop the challenge cycle.

snowmead: this is safe to do, since the `do_bsp_confirm_storing` function in the file system pallet will initialize the challenge cycle if the root is no longer the default.

https://github.com/Moonsong-Labs/storage-hub/blob/e722b3c6ca1fae6ee8e4bca31f4154bea69af391/pallets/file-system/src/utils.rs#L1306-L1310

undercover-cactus: Done!

pallets/file-system/src/utils.rs
```diff
@@ -1913,6 +1923,11 @@ where
         // Decrease data used by the SP.
         <T::Providers as MutateStorageProvidersInterface>::decrease_capacity_used(&sp_id, size)?;
 
+        // If it doesn't store any files we stop the challenge cycle.
+        if new_root == <T::Providers as shp_traits::ReadProvidersInterface>::get_default_root() {
+            <T::ProofDealer as shp_traits::ProofsDealerInterface>::stop_challenge_cycle(&sp_id)?;
+        };
```
TDemeco, line 1929: Shouldn't this only be done for BSPs? Since we don't have a challenge cycle for MSPs

pallets/file-system/src/utils.rs
```diff
@@ -1400,6 +1400,11 @@ where
         // Update root of BSP.
         <T::Providers as shp_traits::MutateProvidersInterface>::update_root(bsp_id, new_root)?;
 
+        if new_root == <T::Providers as shp_traits::ReadProvidersInterface>::get_default_root() {
+            // We should remove the BSP from the dealer proof
+            <T::ProofDealer as shp_traits::ProofsDealerInterface>::stop_challenge_cycle(&bsp_id)?;
+        };
```
TDemeco, line 1406: This is the `confirm_storing` extrinsic, which means the only mutations this can have is trie add mutations (not remove), so the new root after apply delta should never be the default root (if it is, there's something wrong)

TDemeco (changes requested):
Left some comments. There's also the issue of stopping a BSP's challenge cycle if, after responding to a checkpoint challenge that was a trie remove mutation, it's left without files. This could be done directly in the Proofs Dealer since it has that info readily available (the new root, the default root, and the challenge cycle)

test/suites/integration/bsp/submit-proofs.test.ts
```diff
@@ -237,8 +237,10 @@ describeBspNet(
       );
     });
 
-    it("BSP is not challenged any more", { skip: "Not implemented yet." }, async () => {
-      // TODO: Check that BSP-Three no longer has a challenge deadline.
+    it("BSP three is not challenged any more", async () => {
+      const result = await userApi.call.proofsDealerApi.getNextDeadlineTick(ShConsts.BSP_THREE_ID);
+
+      assert(result.isErr, "BSP three doesn't have files so it it shouldnt have deadline");
```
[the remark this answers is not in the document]
TDemeco:
```suggestion
      assert(result.isErr, "BSP three doesn't have files so it shouldn't have deadline");
```
😁

pallets/file-system/src/utils.rs
undercover-cactus: I moved things a bit to do all the task for bsp providers. It should be resolved.

pallets/file-system/src/utils.rs
undercover-cactus: I have removed it entirely. Not sure if I should still leave a check with an unreachable call...

TDemeco (commented):
Left a comment to optimize the way we check in the `sp_stop_storing_for_insolvent_user`. Also to reiterate my previous comment, the call to `stop_challenge_cycle` extrinsic in the `do_submit_proof` logic is missing. We should call it after calling `apply_delta` ([here](https://github.com/Moonsong-Labs/storage-hub/blob/949c014291a74d9578c338239da1e30c710c1e04/pallets/proofs-dealer/src/utils.rs#L272)) to make sure we don't keep challenging BSPs that do not have any files left

pallets/file-system/src/utils.rs
```diff
@@ -1790,6 +1790,11 @@ where
         // Remove the pending stop storing request from storage.
         <PendingStopStoringRequests<T>>::remove(&bsp_id, &file_key);
 
+        if new_root == <T::Providers as shp_traits::ReadProvidersInterface>::get_default_root() {
+            // We should remove the BSP from the dealer proof
+            <T::ProofDealer as shp_traits::ProofsDealerInterface>::stop_challenge_cycle(&bsp_id)?;
+        };
```
ffarall, line 1796: This check do it as you did the one in `do_sp_stop_storing_for_insolvent_user` (also considering my suggestion for that check).

pallets/proofs-dealer/src/utils.rs
```diff
@@ -295,17 +295,22 @@ where
...
 
+                // Update root of Provider after all mutations have been applied to the Forest.
+                <T::ProvidersPallet as MutateChallengeableProvidersInterface>::update_root(
+                    *submitter, new_root,
+                )?;
+
+                if new_root == ProvidersPalletFor::<T>::get_default_root() {
+                    // We should remove the BSP from the dealer proof
+                    Self::stop_challenge_cycle(submitter)?;
+                };
```
ffarall, line 306: We should be stopping the randomness cycle here too, don't we? Which will be a problem given that this pallet doesn't have access to the `cr-randomness` pallet. 

Anyway, we will have to add considering this in the benchmarks. @santikaplan please add a task for this.

ffarall: Also @santikaplan, add a task for creating an extrinsic to manually stop the challenge cycle, checking if the root is the default one. And the same for the randomness cycle.

test/suites/integration/bsp/submit-proofs.test.ts
```diff
@@ -245,8 +245,10 @@ describeBspNet(
       );
     });
 
-    it("BSP is not challenged any more", { skip: "Not implemented yet." }, async () => {
-      // TODO: Check that BSP-Three no longer has a challenge deadline.
+    it("BSP three is not challenged any more", async () => {
+      const result = await userApi.call.proofsDealerApi.getNextDeadlineTick(ShConsts.BSP_THREE_ID);
+
+      assert(result.isErr, "BSP three doesn't have files so it shouldn't have deadline");
```
ffarall, line 251: Awesome!

ffarall (changes requested):
Great work, just some comments that you made me think of.

pallets/file-system/src/utils.rs
```diff
@@ -1944,13 +1949,25 @@ where
...
-        }
+        if <T::Providers as ReadStorageProvidersInterface>::is_bsp(&sp_id) {
+            // If it doesn't store any files we stop the challenge cycle and stop its randomness cycle.
+            if new_root == <T::Providers as shp_traits::ReadProvidersInterface>::get_default_root()
+            {
+                ensure!(
+                    <T::Providers as ReadStorageProvidersInterface>::get_used_capacity(&sp_id)
+                        == Zero::zero(),
+                    Error::<T>::UsedCapacityShouldBeZero,
+                );
```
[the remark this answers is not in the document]
undercover-cactus: I was also on the fence with adding this. I think checking the merkle root will lead to less inconsistencies than checking the used capacity. I will change it to just an error message.

It would help to have a single source of truth for example the provider could have different state (e.g Idle, Hosting, etc...).

pallets/file-system/src/utils.rs
```diff
@@ -1944,13 +1958,24 @@ where
...
         <T::Providers as MutateStorageProvidersInterface>::decrease_capacity_used(&sp_id, size)?;
 
-        // If the new capacity used is 0 and the Provider is a BSP, stop its randomness cycle.
         if <T::Providers as ReadStorageProvidersInterface>::is_bsp(&sp_id)
-            && <T::Providers as ReadStorageProvidersInterface>::get_used_capacity(&sp_id)
-                == Zero::zero()
+            && new_root == <T::Providers as shp_traits::ReadProvidersInterface>::get_default_root()
         {
+            // If it doesn't store any files we stop the challenge cycle and stop its randomness cycle.
+
```
ffarall, line 1965:
```suggestion
            // If it doesn't store any files we stop the challenge cycle and stop its randomness cycle.
```

ffarall: (Removed unnecessary whitespace)

pallets/file-system/src/utils.rs
```diff
@@ -1944,13 +1958,24 @@ where
...
+            let used_capacity =
+                <T::Providers as ReadStorageProvidersInterface>::get_used_capacity(&sp_id);
+            if used_capacity == Zero::zero() {
+                // Emit event if we have inconsistency. We can later monitor for those.
```
ffarall, line 1969: Wouldn't it be the other way around? The inconsistency would be if the capacity IS NOT zero.

```suggestion
            if used_capacity != Zero::zero() {
                // Emit event if we have inconsistency. We can later monitor for those.
```

pallets/file-system/src/utils.rs
```diff
@@ -1790,6 +1790,20 @@ where
         // Remove the pending stop storing request from storage.
         <PendingStopStoringRequests<T>>::remove(&bsp_id, &file_key);
 
+        if new_root == <T::Providers as shp_traits::ReadProvidersInterface>::get_default_root() {
+            let used_capacity =
+                <T::Providers as ReadStorageProvidersInterface>::get_used_capacity(&bsp_id);
+            if used_capacity == Zero::zero() {
```
ffarall, line 1796:
```suggestion
            if used_capacity != Zero::zero() {
```

ffarall (approved):
Please review the changes suggested. Otherwise good to go!

pallets/file-system/src/utils.rs
```diff
@@ -1790,6 +1790,20 @@ where
...
+                <T::Providers as ReadStorageProvidersInterface>::get_used_capacity(&bsp_id);
+            if used_capacity != Zero::zero() {
+                // Emit event if we have inconsistency. We can later monitor for those.
+                Self::deposit_event(Event::UsedCapacityShouldBeZero {
+                    actual_used_capacity: used_capacity,
+                });
+            }
+
+            // We should remove the BSP from the dealer proof
+            <T::ProofDealer as shp_traits::ProofsDealerInterface>::stop_challenge_cycle(&bsp_id)?;
```
TDemeco, line 1804:
```suggestion
            // Stop the BSP's challenge and randomness cycles.
            <T::ProofDealer as shp_traits::ProofsDealerInterface>::stop_challenge_cycle(&bsp_id)?;
            <T::CrRandomness as CommitRevealRandomnessInterface>::stop_randomness_cycle(&bsp_id)?;
```

pallets/proofs-dealer/src/utils.rs
TDemeco: +1

pallets/file-system/src/lib.rs
```diff
@@ -822,6 +826,8 @@ pub mod pallet {
         NoPrivacyChange,
         /// Operations not allowed for insolvent provider
         OperationNotAllowedForInsolventProvider,
+        /// Used Capacity value should be 0.
+        UsedCapacityShouldBeZero,
```
TDemeco, line 830: Delete this since we don't error out anymore for this

TDemeco (approved):
Approving to not block it anymore but requested some minor changes

PR #295, opened 2024-12-19
wip: 🚧 benchmark a few more extrinsics of the File System pallet

TDemeco:
This PR adds the benchmark code for `request_move_bucket`, `msp_respond_move_bucket_request`, `update_bucket_privacy`, `create_and_associate_collection_with_bucket`, `delete_bucket`, `bsp_volunteer` and the partial setup for `bsp_confirm_storing`. To keep going with the rest of the extrinsics, a script to generate varios non-inclusion and inclusion forest proofs and file key proofs is needed, which is currently being worked on.


pallets/file-system/src/utils.rs
```diff
@@ -400,7 +400,6 @@ where
 
         if response == BucketMoveRequestResponse::Rejected {
             <PendingBucketsToMove<T>>::remove(&bucket_id);
```
snowmead, line 402: you could also just call this once since we delete it whether the msp rejects or accepts the request.

snowmead (approved):

pallets/file-system/src/utils.rs
ffarall: Agreed

ffarall (approved):

pallets/file-system/src/utils.rs
TDemeco: I refactored this! Also the MoveBucketAccepted event was being emitted twice so I removed the call from `utils.rs`

PR #299, opened 2024-12-23
test: ✅ Fix race conditions in integration tests

ffarall:
In this PR:
1. Fix race condition in `skipTo` waiting for proofs (eliminated `sleep`s).
2. Fix race condition where there could have been multiple `PaymentStreamCharged` events in `debt-collection.test.ts`.
3. Fix race condition where MSP didn't add a file to it's local File Storage in time in `respond-multi-requests.test.ts`.
4. Remove redundant iterations in `waitForTxInPool`.
5. Change file from `smile.jpg` to `whatsup.jpg` in `bsp-threshold.test.ts` to make it so that `BSP_DOWN` cannot immediately volunteer.
6. Simplify logic in `bsp-threshold.test.ts` and make it deterministic which BSP should be volunteering first (that shouldn't change given that their IDs and the file keys of the files uploaded don't change ever).
7. Fix race condition in `bsp-threshold.test.ts` where BSP was not yet synced with chain tip before waiting for volunteer.
8. Wait for BSPs to catch up to chain tip in `submit-proofs.test.ts`.
9. Add minimum check of number of txs in pool.
10. Fix race condition in `storage-capacity.test.ts` where a transaction was submitted with the BSP key, but the BSP node had already submitted a transaction with the same nonce.

### DIFF
--- a/test/util/bspNet/block.ts
+++ b/test/util/bspNet/block.ts
@@ -16,6 +16,7 @@ import * as ShConsts from "./consts";
 import assert, { strictEqual } from "node:assert";
 import * as Assertions from "../asserts";
 import { waitForLog } from "./docker";
+import { waitForTxInPool } from "./waits";
 
 export interface SealedBlock {
   blockReceipt: CreatedBlock;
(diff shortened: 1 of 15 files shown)


test/util/bspNet/block.ts
```diff
@@ -397,16 +398,23 @@ export const advanceToBlock = async (
...
+      let txsToWaitFor = 0;
       for (const challengeBlockNumber of challengeBlockNumbers) {
         if (currentBlockNumber === challengeBlockNumber.nextChallengeBlock) {
-          // Wait for the BSP to process the proof.
-          await sleep(500);
+          txsToWaitFor++;
 
           // Update next challenge block.
           challengeBlockNumbers[0].nextChallengeBlock += challengeBlockNumber.challengePeriod;
           break;
```
TDemeco, line 408: Should this break be here? What's its purpose?

ffarall: No it shouldn't, you're right. There could be more than one BSP having its next challenge in `currentBlockNumber`, in which case we need to exhaustively check all the items in `challengeBlockNumbers`.

PR #300, opened 2024-12-24
fix: 🚑 add missing `update_payment_stream` calls when deleting files 

TDemeco:
This PR fixes some lingering issues in the runtime regarding payment streams not being correctly updated for Buckets and BSPs when deleting files. Those fixes are:
- In `bsp_confirm_stop_storing`, after the proof has been verified and the used capacity of the BSP has been decreased, added a calculation of the new amount provided from the BSP to the corresponding user and the update (or delete if this new amount is zero) of the payment stream between them. Since the user of the file was not available at this time I added it to the tuple found in `PendingStopStoringRequests`, which is set in the `bsp_request_stop_storing` extrinsic.
- In the `delete_file` extrinsic, if the user supplied an inclusion proof for the file key it wants to delete, the bucket size was being decreased (and the payment stream was updated accordingly) but the new bucket root after the file deletion was not being calculated nor updated. Added that.
- In the `pending_file_deletion_request_submit_proof` (used by MSPs when a user requests a file deletion), if the proof submitted by the MSP includes the file key to delete, the bucket size nor root were being updated (and the payment stream wasn't either). Added the bucket root update and the bucket size update (which also updates the payment stream).
  - To do this we needed the file size which was not readily available at the time the extrinsic was executing, so I added the file size to the tuple stored under `PendingFileDeletionRequests`. This gets set when the user calls the `delete_file` extrinsic and gets verified in the `pending_file_deletion_request_submit_proof` call.
- Changed the `update_provider_after_key_removal` logic to delete the payment stream if the new amount provided after the file key deletion is zero. Previously, it was always calling update instead of delete and the call would fail. This function gets called after the mutations get applied in a `submit_proof` extrinsic call, to update the BSP's used capacity and payment streams.
- Updated the forest verifier's `apply_delta` logic to first get the value (encoded file metadata) under the key to remove with `.get()` and then remove it with `.remove()`. This was needed as calling `.remove()` directly and handling the returned value as the leaf value was not giving correct results, because the `.remove()` call from the `TrieDBMut` structure returns the literal value under the key in the trie, which is almost always a node that contains the hash that points to the actual leaf value, and since the actual leaf value gets removed in the call, calling `.get()` afterwards for the obtained hash returned `None`. I believe this behaviour is quite confusing and should be fixed (or at least clarified) in `trie-db`.
- Finally, after these last two fixes, now payment streams were correctly updating for trie remove mutations done in `submit_proof`, which meant that there was a new worst-case scenario to handle for this extrinsic (updating a payment stream for each mutation). I added a payment stream creation in the setup of the `submit_proof` benchmarks and recalculated the results.


santikaplan:
Fantastic PR! Thank you for spotting all these issues and taking the time to fix them! 💪

pallets/file-system/src/utils.rs
```diff
@@ -2074,7 +2106,24 @@ where
...
+                    &forest_proof,
+                )?;
+
+            // Update root of the Bucket.
+            <T::Providers as shp_traits::MutateBucketsInterface>::change_root_bucket(
+                bucket_id, new_root,
+            )?;
+
+            // Decrease size of the bucket.
+            <T::Providers as MutateBucketsInterface>::decrease_bucket_size(&bucket_id, file_size)?;
```
ffarall, line 2125: This is worth adding a check in at least one of the tests for this extrinsic.

pallets/file-system/src/utils.rs
```diff
@@ -2007,15 +2025,28 @@ where
...
+                        &inclusion_forest_proof,
+                    )?;
+
+                // Update root of the Bucket.
+                <T::Providers as shp_traits::MutateBucketsInterface>::change_root_bucket(
+                    bucket_id, new_root,
+                )?;
+
+                // Decrease size of the bucket.
+                <T::Providers as MutateBucketsInterface>::decrease_bucket_size(&bucket_id, size)?;
```
ffarall, line 2042: This is worth adding a check in at least one of the tests for this extrinsic.

pallets/proofs-dealer/src/utils.rs
```diff
@@ -1032,13 +1032,13 @@ impl<T: pallet::Config> ProofsDealerInterface for Pallet<T> {
...
+        let (_, new_root, _) =
             <T::ForestVerifier as TrieProofDeltaApplier<T::MerkleTrieHashing>>::apply_delta(
                 &root, mutations, proof,
             )
-            .map_err(|_| Error::<T>::FailedToApplyDelta)?
-            .1,
-        )
+            .map_err(|_| Error::<T>::FailedToApplyDelta)?;
+
+        Ok(new_root)
```
ffarall, line 1041: This change is not really necessary isn't it?

pallets/proofs-dealer/src/benchmark_proofs.rs
```diff
@@ -30,6 +32,11 @@ pub fn get_root() -> Vec<u8> {
     hex::decode("912896cfb8a6b305bd4865e5412b31b102b79b5f9652df471a9c93477e54d916").expect("Root should be a decodable hex string")
 }
 
+#[rustfmt::skip]
+pub fn get_user_account() -> AccountId32 {
+	<AccountId32 as Ss58Codec>::from_ss58check("5CombC1j5ZmdNMEpWYpeEWcKPPYcKsC1WgMPgzGLU72SLa4o").expect("User account should be a decodable string")
+}
```
ffarall, line 38: This change should actually go in the `benchmark_proofs_template.rs` file, and be updated by the script.

pallets/proofs-dealer/Cargo.toml
```diff
@@ -16,8 +16,10 @@ targets = ["x86_64-unknown-linux-gnu"]
 
 [dependencies]
 codec = { workspace = true }
+log = { workspace = true, optional = true }
```
ffarall, line 19: Not needed

pallets/proofs-dealer/Cargo.toml
```diff
@@ -16,8 +16,10 @@ targets = ["x86_64-unknown-linux-gnu"]
...
 hex = { workspace = true, optional = true }
 scale-info = { workspace = true }
+sp-core = { workspace = true, optional = true }
```
ffarall, line 22: This should be down in `# Substrate`

ffarall (changes requested):

pallets/proofs-dealer/src/utils.rs
TDemeco: No but I just felt what we were returning wasn't clear enough 😬

test/util/bspNet/waits.ts
```diff
@@ -162,7 +164,8 @@ export const waitForBspStored = async (
         module: "fileSystem",
         method: "bspConfirmStoring",
         checkTxPool: true,
-        timeout: 300
+        timeout: 10000,
```
ffarall, line 167: This was not really necessary given that this function already iterates too. But it's fine anyway.

ffarall (approved):

PR #301, opened 2024-12-26
feat: add test to verify that at max capacity bsp take only what it can store

undercover-cactus:
This PR add a test to verify that the BSP will not take more than it can store. When increasing the capacity we can batch it to satisfy all the storage requests but it should never increase more than the max set in the configuration.


ffarall (approved):
Great job!

PR #303, opened 2024-12-31
feat: ✨ Trigger proof submission catch up every 4 blocks

ffarall:
In this PR:
1. Make Blockchain Service trigger proof submission catch up every 4 blocks (configurable) to further ensure always being up to date with proofs.
2. Solve race condition in integration tests where sometimes the BSP being awaited to catch up to chain tip, would get stuck in a previous block, never syncing until it is gossiped of a new block.
3. Solve race condition in `storage-delete.test.ts` where sometimes it generate a Forest proof before having processed the latest Forest changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -191,9 +191,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.94"
+version = "1.0.95"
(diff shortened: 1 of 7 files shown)


TDemeco (approved):
LGTM 👍🏻 hopefully the fix works

PR #304, opened 2025-01-02
feat : add exclude list

undercover-cactus:
In this PR we add an exclude list into the `FileStorage` it allows us to add and remove filekey of file a BSP doesn't want to store. Especially this will help avoiding a scenario where a BSP delete a file and directly try re-upload it because it received a storage request.

We will have the following features added : 
* In memory exclude list for in memory storage (will diasappeared after stopping the node)
* RocksDB exclude list storage for when using RocksDB storage (persistent storage)
* ~~Automatically exclude file a BSP have deleted~~
* allow to add/remove file to the exclude list from RPC


snowmead:
was there a reason for adding the exclude list to the file storage modules instead of the more _common_ [state](https://github.com/Moonsong-Labs/storage-hub/blob/5a782f55a954335bdd4617f4a4bda84067415309/client/blockchain-service/src/state.rs) where we store other relevant information?

imo, the file and forest storage should only be managing and aware of metadatas, files, file proofs and forests. mainly, the file storage does not utilize this information in any way. we are simply persisting this data to avoid losing it after restarts.

cc @ffarall @links234

ffarall:
> was there a reason for adding the exclude list to the file storage modules instead of the more common [state](https://github.com/Moonsong-Labs/storage-hub/blob/5a782f55a954335bdd4617f4a4bda84067415309/client/blockchain-service/src/state.rs) where we store other relevant information?

@snowmead yes. The reason is that we had to be able to modify the allow list on-the-fly through something like an RPC method. RPC methods cannot have access to the Blockchain Service. Putting it in the File Storage gave us the freedom of having different implementations for the allow list (persistent or not), but most importantly, it is accessible both from the RPCs and the tasks. Also, it didn't feel too out of place to have a _file exclude list_ in the _File Manager_.

snowmead (approved):
beast

ffarall (approved):
I see that you weren't able to fully test this out, given that the file is deleted automatically due to the expiring storage request at the end of the test. Just like you describe in the tests. This is a problem, and we should fix it in another PR.

I'm referencing this comment in the integration tests:
https://github.com/Moonsong-Labs/storage-hub/blob/648c286b3b3190c7cb5abbd5759c7e402906fd65/test/suites/integration/bsp/storage-delete.test.ts#L115-L129

PR #307, opened 2025-01-06
feat: ✨ finish File System pallet benchmarking

TDemeco:
This PR completes the benchmarking of the File System pallet.

- Client changelog:
  - Added two RPC methods to the client (`generateFileKeyProofBspConfirm` and `generateFileKeyProofMspAccept`) which generate the file key proof needed by BSPs or MSPs to confirm or accept a storage request, using the special chunk challenges given by the runtime.
  - Added a check to the BSP submit proof task when reacting to the `ProcessSubmitProofRequest` event, to avoid submitting proofs if there're no challenges to respond to.
    - This should not happen in a real-world scenario since there should always be random challenges, but it was panicking when building with the `runtime-benchmarks` feature flag since that scenario does not have them, so I though might as well add a check there since it doesn't hurt and gives us a bit more flexibility with random challenges in the future.
- Runtime changelog: 
  - Modified the `bsp_confirm_storing` extrinsic of the File System pallet to use the already-existing struct `FileKeyWithProof` instead of a tuple of the file key and proof to make it tidier. Modified all interactions with `bsp_confirm_storing` to accomodate for this change.
  - Added missing calls to `decrease_capacity_used` for MSPs in both the `delete_file` and `pending_file_deletion_request_submit_proof` extrinsics of the File System pallet. Previously, used capacity of the MSP that held the deleted file wasn't being updated when it should after verifying the inclusion proof.
  - Created the `benchmark_proofs_template.rs` template file, which is used by the script that generates the proofs required to benchmark the File System pallet.
  - Incremented the `StorageRequestTtl` when compiling with the `runtime-benchmarks` feature flag to 2000 to avoid storage requests expiring when generating the required file key proofs in the aforementioned script.
  - Benchmarked the File System pallet completely, considering the input variables for each extrinsic and the worst-case scenario when checking for divergent paths in the weight calculation phase would be too costly. Details of each benchmark can be found in the corresponding `benchmarking.rs` file.
- Scripts and integration tests changelog:
  - Developed the script to generate the required inclusion and non-inclusion forest proofs, file keys and file key proofs, bucket IDs and bucket roots for all required benchmark cases of the File System pallet.
  - Updated the `newBucket` method of the integration testing API to be able to receive the MSP ID that should store the new bucket.


undercover-cactus (approved):

pallets/file-system/src/tests.rs
```diff
@@ -5030,13 +5030,11 @@ mod bsp_confirm {
...
-                        (
-                            file_key,
-                            CompactProof {
-                                encoded_nodes: vec![file_key.as_ref().to_vec()],
-                            },
-                        )
+                    .map(|file_key| FileKeyWithProof {
+                        file_key,
+                        proof: CompactProof {
+                            encoded_nodes: vec![H256::default().as_ref().to_vec()],
```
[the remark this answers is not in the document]
TDemeco: It hasn't but the test still behaves correctly since the check for capacity is before the verification of the proof. I'll revert back to the previous generated proof

pallets/file-system/src/utils.rs
```diff
@@ -2063,6 +2068,13 @@ where
                 // Decrease size of the bucket.
                 <T::Providers as MutateBucketsInterface>::decrease_bucket_size(&bucket_id, size)?;
 
+                // If the bucket has a MSP, decrease its used capacity.
+                if let Some(msp_id) = msp_id {
+                    <T::Providers as MutateStorageProvidersInterface>::decrease_capacity_used(
+                        &msp_id, size,
+                    )?;
+                }
```
snowmead, line 2076: can you add the checks in the unit tests for this? as well for the other place you added it in

snowmead (approved):
beast

test/scripts/generateFileSystemBenchmarkProofs.ts
```diff
@@ -0,0 +1,553 @@
...
+  // * For a MSP accepting storage requests, we need to generate a non-inclusion proof to be used
+  // * when accepting from 1 to 10 (MaxBatchMspRespondStorageRequests) file keys and rejecting 1 to 10 (MaxBatchMspRespondStorageRequests)
+  // * file keys. This has to be done for 1 to 10 (MaxBatchMspRespondStorageRequests) buckets.
+  // * We also need to generate the file key proofs for each one of the file keys to be accepted.
+  // To do this, we generate a non-inclusion forest proof for N file keys (which must be the ones not
+  // added to the Forest in the previous step of this script) and add the proof to the array. Then, generate for each file key its proof.
+  // Finally, repeat this for each bucket.
+  // * For a BSP confirming storing, we need the same thing (a non-inclusion forest proof for 1 to 10 file keys that
+  // * the BSP wants to confirm) and the file key proofs for each one of the file keys.
+  // Since the BSP only requires one non-inclusion proof and 1 to 10 (MaxBatchBspConfirmStoring) file key proofs, we reutilise the
```
ffarall, line 288:
```suggestion
  // Since the BSP only requires one non-inclusion proof and 1 to 10 (MaxBatchBspConfirmStoring) file key proofs, we reuse the
```

test/scripts/generateFileSystemBenchmarkProofs.ts
```diff
@@ -0,0 +1,553 @@
...
+  // non-inclusion proof from the first bucket and generate the corresponding file key proofs since the challenges for the BSP are different.
```
ffarall, line 289: I assume this means you'll set the BSP's root to be the root of the bucket, right?

ffarall (approved):

test/scripts/generateFileSystemBenchmarkProofs.ts
TDemeco: Yep!

PR #308, opened 2025-01-06
chore: Unallow buckets without MSPs (only while they are being moved)

snowmead:
Require an MSP for the following extrinsics:

- `create_bucket`
- `issue_storage_request`
- `delete_file`

Run `fullnet` for BspNet to run an MSP node to pass integration tests. All storage requests require an MSP to accept a storage request for it to be fulfilled.


test/suites/integration/bsp/debt-collection.test.ts
```diff
@@ -606,7 +606,7 @@ describeBspNet(
         method: "chargeMultipleUsersPaymentStreams",
         module: "paymentStreams",
         checkTxPool: true,
-        assertLength: 3
+        assertLength: 4
```
ffarall, line 609: Why a fourth? The MSP? The MSP is not guaranteed to send a charge transaction at this point, they do that periodically, not after sending proofs. So maybe using the `exactLength = false` option is a better idea. And keeping it at 3.

ffarall (approved):
Great work!

PR #309, opened 2025-01-07
feat: Undbound MSP respond to storage requests extrinsic

snowmead:
`msp_respond_storage_requests_multiple_buckets` now accepts an unbounded number of buckets and file keys accepted or rejected.

The benchmarking ([PR](https://github.com/Moonsong-Labs/storage-hub/pull/307)) of this function will be the limiting factor for extrinsics to fit into a block. Therefore boundness is not needed.


client/blockchain-service/src/utils.rs
```diff
@@ -673,23 +672,16 @@ where
...
-                for _ in 0..max_batch_respond {
-                    if let Some(request) = state_store_context
-                        .pending_msp_respond_storage_request_deque()
-                        .pop_front()
-                    {
-                        respond_storage_requests.push(request);
-                    } else {
-                        break;
-                    }
+                while let Some(request) = state_store_context
```
ffarall, line 676: We should probably still limit this. We don't want it to try and batch all the confirmations, as it may reach a point where it's just too heavy of a transaction to be included in a block. It could be a local configuration, or later on, from the config file.

ffarall (approved):

client/blockchain-service/src/utils.rs
snowmead: ok. I left a TODO in the task itself to split this into chunks if it was too big. but it seems just better to split at the event emission level. will revert

PR #310, opened 2025-01-07
feat: ✨ enforce file deletion by MSPs

TDemeco:
This PR adds the logic to remove a MSP from the privileged providers list from the moment a user executes the `delete_file` extrinsic for a file key (without providing an inclusion proof) up until the MSP executes the `pending_file_deletion_request_submit_proof` extrinsic to provide an inclusion or non-inclusion proof for that file key.
This way MSPs are incentivized to submit the required proof since otherwise they won't be able to charge any of their users.


ffarall (approved):
Great work!

PR #311, opened 2025-01-07
feat: ✨ remove file deletion requests expiration

TDemeco:
This PR makes it so file deletion requests no longer expire. This is possible since MSPs are now incentivized to submit the required proofs to close the pending file deletion requests because they are removed from the privileged providers in the meantime, so they can't charge any users.
This PR should be merged after #310 

TODO list:
- [x] Keep track of how many pending file deletion requests a MSP has and add it to the privileged providers storage only if it has submitted a proof for the last one remaining.
- [x] Check if incentives are aligned correctly in terms of avoiding MSPs being spammed by malicious users with file deletion requests.


snowmead (approved):
love seeing all the deleted complexity

ffarall (approved):

ffarall:
@TDemeco Regarding the soundness of incentives to prevent spamming with file deletions, I can think of two mitigations:
1. Only allowing file deletions for buckets that we can check that exist. This one we should implement for sure. It's free and gives us some better guarantees, as the MSP could only be spammed by users from whom it accepted storing buckets.
2. Requiring a deposit from the user to delete a file, that would be released if the MSP responds with an inclusion proof (i.e. the file exists). But if the MSP responds with an non-inclusion proof (the file doesn't exist), the deposit goes to the MSP.

PR #312, opened 2025-01-07
feat: ✨ Implement Forest root changes catch up in Blockchain Service

ffarall:
In this PR:
1. Now Forest root changes are applied by the Blockchain Service instead of the the tasks themselves. This unifies the flow and makes sure that at every block, we are in sync with the chain's Forest root (JUST FOR BSPs).
2. Utilises the above to handle re-orgs and Forest root changes that have to be rolled back/applied in them.
3. Adds events to `proofs-dealer` pallet to better listen to Forest root changes.
4. Adds tests to `reorg-proof` test suite with edge case scenarios of reorgs.


test/util/bspNet/test-api.ts
```diff
@@ -456,24 +465,39 @@ export class BspNetTestApi implements AsyncDisposable {
...
        *
-       * @returns A promise that resolves when the chain re-org is complete.
+       * @throws Will throw an error if the head block is already finalised.
+       * @returns A Promise that resolves when the chain reorganization is complete.
        */
       reOrgWithFinality: () => BspNetBlock.reOrgWithFinality(this._api),
       /**
-       * Causes a chain re-org by creating a longer forked chain.
-       * Note: This requires the head block to be unfinalised, otherwise it will throw!
+       * Performs a chain reorganisation by creating a longer forked chain.
```
timbrinded, line 483: these comments are good

client/blockchain-service/src/utils.rs
```diff
@@ -909,13 +1043,236 @@ where
...
+    ) -> Result<()> {
+        let current_forest_key = CURRENT_FOREST_KEY.to_vec();
+        let fs = self
+            .forest_storage_handler
+            .get(&current_forest_key.into())
+            .await
+            .ok_or_else(|| anyhow!("CRITICAL❗️❗️ Failed to get forest storage."))?;
+
+        // Write lock is released when exiting the scope of this `match` statement.
+        match mutation {
```
TDemeco, line 1213: Maybe I'm mistaken, but shouldn't we send a tx to get the write-lock of the forest root? Or does it happen automatically here in the blockchain service instead of having to do it manually like we do on the tasks?

test/suites/integration/bsp/reorg-proof.test.ts
```diff
@@ -1,20 +1,46 @@
...
     before(async () => {
       userApi = await createUserApi();
       bspApi = await createBspApi();
     });
 
+    it("Set tick range to maximum threshold to immediately accept volunteers", async () => {
+      // Set global parameters
+      const { extSuccess } = await userApi.block.seal({
+        calls: [userApi.tx.sudo.sudo(userApi.tx.fileSystem.setGlobalParameters(null, 1))]
+      });
```
TDemeco, line 37: This is going to have to be updated after a merge from main since these are now dynamic params in `pallet-parameters`

test/util/bspNet/block.ts
```diff
@@ -99,6 +99,8 @@ export const sealBlock = async (
     | SubmittableExtrinsic<"promise", ISubmittableResult>
     | SubmittableExtrinsic<"promise", ISubmittableResult>[],
   signer?: KeyringPair,
+  nonce?: number,
+  parentHash?: string,
```
TDemeco, line 103: Add these to the parameters in the docs above

TDemeco (approved):
Overall looks really good! Just left some minor comments

client/blockchain-service/src/utils.rs
```diff
@@ -138,7 +146,85 @@ where
...
+
+                // Update last block added.
+                last_block_added = parent_block_info;
+            }
+
+            // The first element in the route is the last block processed, which will also be the
+            // `pivot`, so it will be ignored when processing the `tree_route`.
+            route.push(last_best_block.clone().into());
+
+            // Revert the route so that it is in ascending order of blocks.
```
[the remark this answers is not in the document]
ffarall: The clearer the better.

client/blockchain-service/src/utils.rs
ffarall: We have two write locks, and I'm not entirely sure which one you're referring to.

There's the write lock we take when we do `fs.write()`, which comes from the `RwLock` type, and it waits until no other threads are reading or writing to the file. Then takes the lock,
and as long as it holds the lock, no other threads can read or write to this local forest. This one we use it here when we apply the mutation.

Then there's the "forest write lock" that we implement with `oneshot` channels. That lock is so that there isn't more than one asynchronous task trying to send forest-root-changing transactions in the same block. This one is not used here, because this code doesn't send any transactions, it merely reacts to transactions that already changed the forest root onchain.

client/blockchain-service/src/utils.rs
```diff
@@ -909,13 +1043,236 @@ where
...
+        if revert {
+            trace!(target: LOG_TARGET, "Reverting Forest root changes for block number {:?} and hash {:?}", block.number, block.hash);
+        } else {
+            trace!(target: LOG_TARGET, "Applying Forest root changes for block number {:?} and hash {:?}", block.number, block.hash);
+        }
+
+        // Process the events in the block, specifically those that are related to the Forest root changes.
+        match get_events_at_block(&self.client, &block.hash) {
+            Ok(events) => {
+                for ev in events {
```
snowmead, line 1082: Shouldn't we reverse the list of events to start from the last event within the block.

I know we are only updating the forest technically once every block, but if ever at some point this changes, this will be an issue since ordering of operations matters. 

Right now it's at a per block basis so there is no issue.

snowmead (approved):
Beast job!

I will approve simply because of the timeline but we can leave some TODOs if ever we need to change some things.

snowmead:
Slight nitpick I forgot to mention is that we could have the `NewBlockNotificationKind` variants have `tree_route` be optional so we don't have to build one manually in the instance when there is `None` tree route in the blockchain import notification.

client/blockchain-service/src/utils.rs
TDemeco: Yep, got confused with the forest write lock we use for txs

client/blockchain-service/src/handler.rs
```diff
@@ -1042,22 +1054,30 @@ where
...
             NewBlockNotificationKind::Reorg {
                 old_best_block: _,
                 new_best_block,
-            } => {
-                // TODO: Handle catch up of reorgs.
-                new_best_block
-            }
+                tree_route,
+            } => (new_best_block, tree_route),
         };
```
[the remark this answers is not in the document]
ffarall: Fair point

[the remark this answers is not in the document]
ffarall: I think for simplicity and ease of use though, I'd rather keep it like this. But I can add assertions to check the cases you just mentioned.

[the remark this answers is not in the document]
ffarall: In fact, now that I check it again, the way the code is done, it is sort of guaranteed that that is the case 🤔

client/blockchain-service/src/utils.rs
ffarall: You mean in case we are reverting the changes from that block? Actually the ordering of operations (adding or removing files) doesn't affect the end result. So it's not a problem.

test/util/bspNet/block.ts
```diff
@@ -128,15 +131,22 @@ export const sealBlock = async (
...
       results.hashes.push(hash);
     }
   }
 
+  let blockReceipt: CreatedBlock;
+  if (parentHash) {
+    blockReceipt = await api.rpc.engine.createBlock(true, finaliseBlock, parentHash);
+  } else {
+    blockReceipt = await api.rpc.engine.createBlock(true, finaliseBlock);
+  }
```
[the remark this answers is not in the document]
ffarall: I tried that first and for whatever reason, when calling `createBlock` with `parentHash == undefined` it interprets it as block hash `0x0000...`.

client/blockchain-service/src/handler.rs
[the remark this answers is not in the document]
ffarall: @santikaplan We agreed that it would be nice to have a `new()` method for the `NewBlockNotificationKind` kinds, which would ensure that this inconsistency is not possible. This is a nice to have.

PR #313, opened 2025-01-07
feat: Add test for checking BSP skipping volunteer if file key is already stored

snowmead:
Add integration test for checking if BSP skips volunteering for file key which is already stored.


ffarall (approved):
Good job! Be careful to merge the other PR and finish the `TODO` here before merging.

snowmead:
complete the TODO before merging. dependency on https://github.com/Moonsong-Labs/storage-hub/pull/308

https://github.com/Moonsong-Labs/storage-hub/blob/aa782f8b7788aadfce89298dd28c9b1621530a21/test/suites/integration/bsp/single-volunteer.test.ts#L6-L7

PR #314, opened 2025-01-08
refactor: ♻️ Replace `TrieRemoveMutation` usage in storage for a struct with a `bool`

ffarall:


pallets/proofs-dealer/src/utils.rs
```diff
@@ -258,11 +262,19 @@ where
...
+                .filter_map(|custom_challenge| {
+                    // The key should be removed if `should_remove_key` is `true` and if when the Provider responds to this challenge with a proof,
+                    // in that proof there is an inclusion proof for that key (i.e. the key is in the Merkle Patricia Forest).
+                    if custom_challenge.should_remove_key
+                        && forest_keys_proven.contains(&custom_challenge.key)
+                    {
+                        Some((
+                            custom_challenge.key,
+                            TrieMutation::Remove(Default::default()),
+                        ))
```
snowmead, line 274: should we unify this as well and have it be a `CustomChallenge` instead of a tuple?

snowmead (approved):

pallets/proofs-dealer/src/utils.rs
ffarall: The internal `apply_delta` from `TrieProofDeltaApplier` I still want it to use `TrieMutations`, because that one can apply both `TrieAddMutation` and `TrieRemoveMutation`s (is abstracted in that way) and also that is the one which would later give me the information of the applied mutations, and I need it to give me a `Vec<TrieMutation>` to publish with the event.

ffarall: I will make `do_submit_proof` use the internal function `apply_delta`, so that all goes in one place. But that will be in the other PR where I also add the events when there is an `apply_delta`

PR #315, opened 2025-01-08
tests: Add integration tests for challenge deadlines changing based on capacity change

snowmead:
Added `challenge-cycle-2.test.ts` integration test in order to test the functionality when a BSP changes their capacity (deposit/stake), their current next challenge tick has not changed only after they have reached it. We then check that the next challenge tick is expected to be calculated based on the new challenge deadline.


ffarall (approved):
Great job!

PR #321, opened 2025-01-10
feat: Implement BSP stop cycles extrinsic

snowmead:


pallets/providers/src/tests.rs
```diff
@@ -6276,6 +6276,68 @@ mod delete_provider {
...
+        #[test]
+        fn stop_all_cycles_works() {
+            ExtBuilder::build().execute_with(|| {
+                // register BSP
+                let bob: AccountId = accounts::BOB.0;
+                let (_bob_deposit, _bob_bsp) = register_account_as_bsp(bob, 100);
+
+                assert_ok!(StorageProviders::stop_all_cycles(RuntimeOrigin::signed(
+                    bob
+                ),));
```
ffarall, line 6335: Maybe worth checking that the storage elements for deadlines was updated. I realise you can't really do that for the CrRandomness since it's mocked, but for the ProofsDealer yes.

pallets/providers/src/lib.rs
```diff
@@ -1450,6 +1459,25 @@ pub mod pallet {
...
+        /// BSP operation to stop all of your automatic cycles.
+        ///
+        /// This includes:
+        ///
+        /// - Commit reveal randomness cycle
+        /// - Proof challenge cycle
+        ///
+        /// If you are an BSP, the only requirement that must be met is that your root is the default one (an empty root).
+        #[pallet::call_index(16)]
+        #[pallet::weight(Weight::from_parts(10_000, 0) + T::DbWeight::get().writes(1))]
```
ffarall, line 1472: We should benchmark this one, however small it is. Maybe leave a task for later @santikaplan ?

ffarall (approved):
Great job!

pallets/providers/src/utils.rs
```diff
@@ -1188,6 +1188,28 @@ where
...
+        }
+
+        <T::ProofDealer as shp_traits::ProofsDealerInterface>::stop_challenge_cycle(&provider_id)?;
+
+        <T::CrRandomness as shp_traits::CommitRevealRandomnessInterface>::stop_randomness_cycle(
+            &provider_id,
+        )?;
+
+        Ok(())
+    }
```
TDemeco, line 1211: Maybe we should add a call to this function in the `bsp_sign_off` extrinsic, just to make sure a BSP cannot delete itself without having its cycles stopped

TDemeco (approved):
Looks good! Just a suggestion

pallets/providers/src/tests.rs
snowmead: agreed. added a check to see if `ProviderToProofSubmissionRecord(provider)` is none

pallets/providers/src/utils.rs
snowmead: that's a good idea thanks for the suggestion

pallets/providers/src/lib.rs
snowmead: agreed. included it in this pr

PR #322, opened 2025-01-10
feat: MSP responding/accepting file key already accepted previously

snowmead:
The goal of this change is to enable MSPs to respond to pre-existing file keys already responded to and stored. This situation arises when users create subsequent storage requests for file key already stored among providers. 

MSPs will have to respond to storage requests by accepting a file key and sending a file key proof along with it just like any other storage request. This is intentional to keep the logic concise and avoiding adding additional complexity with not much benefit.

Some changes were done on the file uploading process to allow for *providers* (MSPs and BSPs) to respond to file upload requests by sending back to the user a boolean flag, signalling that they already are storing all the data for a given file key. This is to avoid entire processes from continuing for potentially large files which is not needed. The MSP will proceed to accept the file.

Replicated the `download_request` and `download_response` pattern for `upload_request`, adding `upload_response` and the necessary tracking ID (`UploadRequestId`) to send a response back from the provider back to the user asynchronously, to prevent slowing down the `FileTransferService` from processing other events.

If a provider attempts to `write_chunk` to file storage and the chunk already exists, we check if the file is already completely stored (all chunks are stored) for a given file key. If it is the case, the provider sends back an upload response with the flag mentioned earlier set to `true` which stops the user from sending anymore chunks.

---
The `DownloadRequestId` and `UploadRequestId` have atomic counters to avoid the IDs not incrementing sequentially since they are updated in a multi-threaded context.

--- 
Merged `unregister_bucket` and `unregister_bucket_with_grace_period` commands into one command called `schedule_unregister_bucket`. Passing `None` as the `grace_period_seconds` unregisters the bucket immediately.


client/file-transfer-service/src/handler.rs
```diff
@@ -524,18 +525,38 @@ impl FileTransferService {
...
+                        Some(success) => success,
+                        None => {
+                            error!(
+                                target: LOG_TARGET,
+                                "Failed to receive upload request result from upper layers"
+                            );
+                            self.handle_bad_request(pending_response);
+                            return;
+                        }
+                    };
```
snowmead, line 552: @links234 wanted your specific input on the added channel field to the `RemoteUploadTask` task event.

snowmead: I didn't see another way to return a custom result type back from a specific event being processed by a task without introducing the channel to the event.

snowmead: this has been reverted

node/src/tasks/bsp_upload_file.rs
```diff
@@ -140,111 +140,23 @@ where
...
-            Ok(proven) => proven,
+        let file_complete = match self.handle_remote_upload_request_event(event.clone()).await {
+            Ok(complete) => complete,
             Err(e) => {
-                warn!(target: LOG_TARGET, "{}", e);
-
-                // Unvolunteer the file.
-                self.unvolunteer_file(event.file_key.into()).await;
+                // Send error response before propagating error
+                let _ = event.file_complete_channel.send(false).await;
```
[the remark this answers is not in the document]
snowmead: this has been reverted

node/src/tasks/bsp_upload_file.rs
```diff
@@ -140,111 +140,23 @@ where
...
-                FileStorageWriteError::FailedToConstructTrieIter => {
-                    // This should never happen for a well constructed trie.
-                    // This means that something is seriously wrong, so we error out the whole task.
-
-                    // Unvolunteer the file.
-                    self.unvolunteer_file(event.file_key.into()).await;
+        // Send completion status to user
+        if let Err(e) = event.file_complete_channel.send(file_complete).await {
+            error!(target: LOG_TARGET, "Failed to send response to channel: {:?}", e);
+        }
```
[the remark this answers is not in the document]
snowmead: this has been reverted in favor of file upload responses. please re-review

client/file-manager/src/rocksdb.rs
```diff
@@ -574,6 +574,41 @@ where
...
+
+        let file_trie =
+            RocksDbFileDataTrie::<T, DB>::from_existing(self.storage.clone(), &mut partial_root);
+
+        let stored_chunks = file_trie.stored_chunks_count()?;
+        Ok(metadata.chunks_count() == stored_chunks)
+    }
+
     /// Stores file metadata and an empty root.
     /// Should only be used if no chunks have beeen written yet.
```
ffarall, line 613:
```suggestion
    /// Should only be used if no chunks have been written yet.
```

client/file-transfer-service/src/events.rs
```diff
@@ -1,13 +1,17 @@
...
+};
 
 #[derive(Clone)]
 pub struct RemoteUploadRequest {
     pub peer: PeerId,
     pub file_key: FileKey,
     pub file_key_proof: FileKeyProof,
     pub bucket_id: Option<BucketId>,
+    /// The request ID used to send back the response through the FileTransferService
+    pub request_id: UploadRequestId,
```
ffarall, line 14: The docs for the new field are appreciated, but let's also add it for the other fields.

pallets/file-system/src/types.rs
```diff
@@ -127,6 +127,8 @@ impl<T: Config> Debug for FileKeyWithProof<T> {
 #[scale_info(skip_type_params(T))]
 pub struct StorageRequestMspAcceptedFileKeys<T: Config> {
     pub file_keys_and_proofs: Vec<FileKeyWithProof<T>>,
+    /// File keys which have already been accepted by the MSP in a previous storage request should be included
+    /// in the proof.
     pub non_inclusion_forest_proof: ForestProof<T>,
```
ffarall, line 132: How about renaming the field to `forest_proof` and explaining in the docs that it can be an inclusion or non-inclusion proof, and how it is interpreted in each case.

ffarall (approved):
Good job!

client/file-transfer-service/src/events.rs
snowmead: agreed

undercover-cactus (approved):

PR #325, opened 2025-01-15
fix: replace file ID from i32 to i64

undercover-cactus:
Replacing file ID from i32 type to i64 type to allow producing IDs for a bigger number of files. Instead of limited the database to containing 2 billions files we can now have a lot more.


links234:
If we are doing this for Files, maybe we should do it for all tables?

undercover-cactus:
> If we are doing this for Files, maybe we should do it for all tables?

I guess yes that would remove any chance of running out of IDs.

links234 (approved):

ffarall (approved):

PR #326, opened 2025-01-15
refactor: :recycle: use ticks instead of blocks in file-system expirations

TDemeco:
This PR changes all items with expirations from the file-system pallet (storage requests, file deletion requests and move bucket requests) to use the ProofsDealer tick instead of the block number.
This makes it so, in case of spamming or ongoing MBMs, items do not expire, which makes sense since for example for storage requests the volunteer threshold doesn't increase so new BSPs cannot volunteer for them if the tick does not advance.


snowmead (commented):
We probablys should do this for provider top expirations in the provider's pallet to avoid provider's from not being able to respond in time and being removed from the network. 

If so, we can do this in another pr

snowmead (approved):

ffarall (approved):
Neat job!

PR #327, opened 2025-01-16
fix: 🗑️ Remove inaccurate message in proviers-pallet lib.rs 

santikaplan:
Delete outdated information in provider-randomness pallet lib.rs


TDemeco (approved):
💯

PR #330, opened 2025-01-17
refactor: :recycle: Changing static sleeps to use polling

timbrinded:
Improve our test cases across the board to use polling instead of static waits.


test/suites/integration/msp/debt-collection.test.ts
```diff
@@ -1,401 +1,424 @@
...
-        file_metadata: { location, fingerprint, file_size }
-      } = await userApi.rpc.storagehubclient.loadFileInStorage(
-        source[i],
-        destination[i],
-        userApi.shConsts.NODE_INFOS.user.AddressId,
-        bucketId
-      );
+describeMspNet(
+  "Single MSP collecting debt",
+  { keepAlive: false },
```
ffarall, line 9: Leftover?

test/suites/integration/msp/debt-collection.test.ts
```diff
@@ -1,401 +1,424 @@
...
+      const sumOfSizeOfFiles = firstFileSize + secondFileSize + thirdFileSize;
+      expectedPaymentStreamRate = Math.round(
+        (valueProps[0].value_prop.price_per_giga_unit_of_data_per_block.toNumber() *
+          sumOfSizeOfFiles) /
+          unitsInGigaUnit +
+          zeroSizeBucketFixedRate
+      );
+      strictEqual(paymentStream.rate.toNumber(), expectedPaymentStreamRate);
+
+      await it("MSP is charging user", async () => {
```
ffarall, line 299: I've had problems when using `only` and nested `it` blocks. That's why I personally avoid them usually. Any thoughts?

test/suites/integration/bsp/single-volunteer.test.ts
```diff
@@ -128,7 +132,13 @@ describeBspNet("Single BSP Volunteering", ({ before, createBspApi, it, createUse
 
     strictEqual(bspConfirmRes_bspId.toHuman(), userApi.shConsts.TEST_ARTEFACTS[source].fingerprint);
 
-    await sleep(1000); // wait for the bsp to process the BspConfirmedStoring event
+    // TODO: Investigate what needs to be added to poll
+    await sleep(1000); // to avoid extFailure- IssueRequest already registered for file
```
ffarall, line 136: Leftover? Since you added the `waitFor`.

ffarall (approved):
Remarkable job! Just a few minor comments

undercover-cactus (approved):

test/suites/integration/bsp/single-volunteer.test.ts
timbrinded: Never got to the bottom of this! If you remove this sleep then ANOTHER test case after this fails. Clearly a task completion needed to be waited for but no indication in logs or events emitted as to which.

test/suites/integration/msp/debt-collection.test.ts
timbrinded: as a rule i dont like nested blocks - however it does allow you to have test dependency. In the original form it's not clear that the test cannot be run in isolation but relied on the previous test case to be run correctly. This explicitly couples the two tests together (i wasted about an hour trying to debug why this test case was failing but passing when running as a suite).

test/suites/integration/msp/debt-collection.test.ts
timbrinded: Done

PR #331, opened 2025-01-17
feat: expand exclude list

undercover-cactus:
In this PR we expand the exclude list to allow excluding specific files, users, buckets or file fingerprints.

This add a check for each category when receiving a StorageRequest event to see if storing this file is allowed.


ffarall (approved):
Loved the way you implemented the `ExcludeType`s, great job! Just a few nitpicks to fix before merging.

PR #332, opened 2025-01-20
feat: New `MspDeleteFileTask`

snowmead:
Implements the `MspDeleteFileTask` task to delete files as an MSP node.

The events handled by this task are the following:

- `FileDeletionRequest`: Event emitted when a file is requested to be deleted by the user
  
  We queue up a file deletion to be processed in batches.

- `ProcessFileDeletionRequest`: Event emitted as a batch of file keys to be deleted

  Generate a proof of (non-)inclusion for each file key, submitting a call to the pending file deletion request extrinsic which will apply a remove delta if it was a proof of inclusion, which in turn will be deleted from the local forest storage of the MSP.

- `FinalisedProofSubmittedForPendingFileDeletionRequest` event emitted when the proof submission was finalized

  If the file is not part of the forest up until this point, we delete the file from the file storage.


node/src/tasks/msp_delete_file.rs
```diff
@@ -0,0 +1,291 @@
...
+            Some(tx) => tx,
+            None => {
+                let err_msg = "CRITICAL❗️❗️ This is a bug! Forest root write tx already taken. This is a critical bug. Please report it to the StorageHub team.";
+                error!(target: LOG_TARGET, err_msg);
+                return Err(anyhow!(err_msg));
+            }
+        };
+
+        // Process each file deletion request sequentially
+        for request in &event.data.file_deletion_requests {
```
ffarall, line 135: I have a problem with this approach. It's not really batching anything, it's just sending one extrinsic per file deletion to respond to. Moreover, it's keeping the forest write lock for (at the very least) one block per file deletion. If we had an extrinsic that would accept multiple file deletions of a single bucket, with one proof of inclusion/non-inclusion for all of them, then maybe it would make sense to batch file deletions from the same bucket, but responding to all of them in just one extrinsic. But that's not the case here. And anyway, currently our file deletion flow is bottlenecked by the priority challenges, so no point in batching a lot right now.

node/src/tasks/msp_delete_file.rs
```diff
@@ -0,0 +1,291 @@
...
+            // Get the forest storage for the bucket
+            let forest_storage = self
+                .storage_hub_handler
+                .forest_storage_handler
+                .get(&request.bucket_id.as_ref().to_vec())
+                .await
+                .ok_or_else(|| anyhow!("Failed to get forest storage."))?;
+
+            // Acquire write lock once for the entire operation
+            let mut forest_storage_write = forest_storage.write().await;
```
ffarall, line 151: Acquiring the write lock here means that you'll have it until the extrinsic is successful (at least a block). So no other task can even read this bucket's forest during that time. This presents a very high risk of a deadlock.

node/src/tasks/msp_delete_file.rs
```diff
@@ -0,0 +1,291 @@
...
+
+            // Generate proof and check existence
```
ffarall, line 153: It doesn't check existence though.

node/src/tasks/msp_delete_file.rs
```diff
@@ -0,0 +1,291 @@
...
+                    user: request.user.clone(),
+                    file_key: request.file_key.into(),
+                    file_size: request.file_size,
+                    bucket_id: request.bucket_id,
+                    forest_proof: forest_proof.proof.clone(),
+                },
+            );
+
+            // Submit extrinsic with retry and wait for it to be included in a block
+            let _events = self
```
ffarall, line 169: Why setting `with_events: true` if you're not using the returned `_events` after?

client/blockchain-service/src/commands.rs
```diff
@@ -27,6 +27,8 @@ use shc_common::types::{
 };
 use storage_hub_runtime::{AccountId, Balance, StorageDataUnit};
 
+use crate::types::FileDeletionRequest;
+
 use super::{
```
ffarall, line 32: Merge these two.

client/blockchain-service/src/events.rs
```diff
@@ -11,7 +11,7 @@ use sp_runtime::AccountId32;
 use std::sync::Arc;
 use tokio::sync::{oneshot, Mutex};
 
-use crate::types::{ConfirmStoringRequest, RespondStorageRequest};
+use crate::types::{self, ConfirmStoringRequest, RespondStorageRequest};
```
ffarall, line 14: Is this necessary?

client/blockchain-service/src/utils.rs
```diff
@@ -526,7 +528,8 @@ where
     /// 2. `ConfirmStoringRequest`.
     ///
     /// If this node is managing a MSP, the priority is given by:
-    /// 1. `RespondStorageRequest`.
+    /// 1. `RespondStorageRequest` over...
+    /// 2. `FileDeletionRequest`.
```
ffarall, line 532: Up for discussion: which should take priority? Because responding to file deletions lets the MSP charge...

On the other hand, we don't want to facilitate spamming an MSP and not having it respond to storage requests because it's busy responding to file deletions.

ffarall: Also, as part of a larger discussion. The MSP doesn't need a "global" forest write lock, only for each bucket. In other words, two forest-root-changing transactions from different buckets, in the same block, are allowed with an MSP.

ffarall (changes requested):

node/src/tasks/msp_delete_file.rs
```diff
@@ -0,0 +1,291 @@
...
+use shc_forest_manager::traits::{ForestStorage, ForestStorageHandler};
+
+use crate::services::{
+    handler::StorageHubHandler,
+    types::{MspForestStorageHandlerT, ShNodeType},
+};
+
+const LOG_TARGET: &str = "msp-delete-file-task";
+const MAX_DELETE_FILE_REQUEST_TRY_COUNT: u32 = 5;
+const MAX_DELETE_FILE_REQUEST_TIP: u128 = 100;
```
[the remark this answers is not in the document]
snowmead: yes they are arbitrary.

client/blockchain-service/src/utils.rs
snowmead: yes I think that with the bucket-wise forest locks this won't be a problem. but even with this I agree with you, we can have the file deletion execute with priority over the respond.

client/blockchain-service/src/events.rs
snowmead: no we don't

undercover-cactus (approved):

ffarall (approved):

PR #334, opened 2025-01-21
fix: 🔒 change volunteer threshold calculation

TDemeco:
This PR:
- Updates the task that volunteers a BSP for a file to send the volunteer extrinsic after it imports the block with `current_tick = volunteer_tick - 1` instead of `current_tick = volunteer_tick`, so the BSP sends the extrinsic to be included in the block when it can actually volunteer instead of waiting one extra block.
  - Since ticks might not advance if the chain is currently being spammed or a MBM is ongoing, if it detects that the first submission of the volunteer extrinsic failed it then waits for `current_tick = volunteer_tick` and sends it again.
- Refactors the calculation of the volunteer tick to hopefully make it a bit more clear, focusing mainly on the difference between the BSP threshold (which is fixed since it depends only on the BSP ID and fingerprint of the file) and the eligibility value (which varies with the BSP's reputation weight, the global reputation weight, the replication factor of the storage request and the ticks since the request was initiated).
- Changes the math so, in a network where all BSPs have the same reputation, the number of BSPs that are eligible to volunteer in the initial tick of a storage request is probabilistically equal to the replication target. This minimizes the probability that a malicious actor controlling a big part of the BSPs in the network can held hostage a file from a user by fully controlling its BSPs, since it makes it so the actor must by chance control all the eligible BSPs of the initial tick.
  - For the lowest security setting we recommend, there's a 1% chance of this happening.
- Updates the `issue_storage_request` extrinsic to accept a variant of the new `ReplicationTarget` enum. This enum can be used by the users to select the security level that they want for the file, with a tradeoff between security and cost of storing (and storage request fulfillment time). There's also the Custom variant of the enum, which allows more advanced users to exactly choose how many BSPs they want as their replication target.
  - All integration tests were updated to use the LowSecurity variant, since in those situations there are no adversaries and it speeds up testing.


pallets/file-system/src/types.rs
```diff
@@ -99,6 +99,65 @@ impl<T: Config> StorageRequestMetadata<T> {
...
 }
 
+/// The enum which holds different options for the replication target of a storage request.
+///
+/// When a user wants to issue a storage request, it can select between any of these options as
+/// the replication target for it. There's a tradeoff between the security level of the data and
+/// both the time it takes for the storage request to be fulfilled and the price paid per byte
+/// during the file's lifetime in StorageHub.
+/// Each option has a different security level, which represents the resiliency that the data will
+/// have against a malicious actor controlling 1/3 of the BSPs of the network.
```
ffarall, line 109:
```suggestion
/// have against a malicious actor controlling 1/3 of the total BSPs weight of the network.
```

pallets/file-system/src/lib.rs
```diff
@@ -300,9 +301,40 @@ pub mod pallet {
         #[pallet::constant]
         type StorageRequestCreationDeposit: Get<BalanceOf<Self>>;
 
-        /// Default replication target
+        /// Low security replication target for a new storage request.
+        ///
+        /// This should be high enough so that it gives users a ~1% chance of their file
+        /// being controlled by a single malicious entity.
```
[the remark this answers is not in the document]
ffarall: I see it explained in more detail in the `ReplicationTarget` enum. Maybe we can reference that here.

pallets/file-system/src/types.rs
```diff
@@ -99,6 +99,65 @@ impl<T: Config> StorageRequestMetadata<T> {
...
+/// actor can hold the file hostage by controlling all its BSPs is ~0.001%.
+/// - UltraHighSecurity: the data will be stored by enough BSPs so the probability that a malicious
+/// actor can hold the file hostage by controlling all its BSPs is ~0.0001%.
+/// - Custom: the user can select the number of BSPs that will store the data. This allows the user to
+/// select the security level of the data manually.
+#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, PartialEq, Eq, Clone)]
+#[scale_info(skip_type_params(T))]
+pub enum ReplicationTarget<T: Config> {
+    LowSecurity,
+    MediumSecurity,
```
ffarall, line 131: Same comment about naming here.

pallets/file-system/src/utils.rs
```diff
@@ -126,57 +127,187 @@ where
...
+
+        // Get the threshold for the BSP to be able to volunteer for the storage request.
+        // The current eligibility value of this storage request for this BSP has to be greater than
+        // this for the BSP to be able to volunteer.
+        let bsp_volunteering_threshold =
+            Self::get_volunteer_threshold_of_bsp(&bsp_id, &fingerprint);
+
+        // Get the current eligibility value of this storage request and the slope with which it
+        // increments every tick (this is weighted considering the BSP reputation so it depends on the BSP).
+        let (bsp_current_eligibility_value, bsp_eligibility_slope) =
```
ffarall, line 153: Like the renaming to `eligibility_value`.

pallets/file-system/src/utils.rs
```diff
@@ -126,57 +127,187 @@ where
...
         T: frame_system::Config,
     {
-        // Get the tick number at which the storage request was created.
-        let (storage_request_tick, fingerprint) = match <StorageRequests<T>>::get(&file_key) {
-            Some(storage_request) => (storage_request.requested_at, storage_request.fingerprint),
-            None => {
-                return Err(QueryFileEarliestVolunteerTickError::StorageRequestNotFound);
-            }
-        };
+        // Get the tick number at which the storage request was created, the associated file fingerprint and
```
ffarall, line 130: Loving the clarity of the changes to this function.

pallets/file-system/src/utils.rs
```diff
@@ -126,57 +127,187 @@ where
...
+            };
+
+        // If the BSP can't volunteer yet, calculate the number of ticks it has to wait for before it can.
+        let min_ticks_to_wait_to_volunteer =
+            match eligibility_diff.checked_div(&bsp_eligibility_slope) {
+                Some(ticks) => max(ticks, T::ThresholdType::one()),
+                None => {
+                    return Err(QueryFileEarliestVolunteerTickError::ThresholdArithmeticError);
+                }
+            };
```
ffarall, line 180: Careful here. `checked_div` will return `None` if there is an underflow. Meaning that if `eligibility_diff < bsp_eligibility_slope`, and `eligibility_diff != 0`, then the result of the division would be a decimal number, so `checked_div` would return `None`. In that case, `min_ticks_to_wait_to_volunteer` should also be 1, not return error.

pallets/file-system/src/utils.rs
```diff
@@ -126,57 +127,187 @@ where
...
-    {
-        // Compute the threshold to succeed and the slope of the bsp.
-        let (to_succeed, slope) =
-            Self::compute_threshold_to_succeed(&bsp_id, storage_request_tick)?;
+    /// Compute the eligibility value threshold for a BSP to be able to volunteer for a storage request
+    /// for a file which has the specified fingerprint.
+    ///
+    /// The threshold is computed by concatenating the encoded BSP ID and file's fingerprint,
+    /// then hashing the result to get the volunteering hash. The volunteering hash is then
+    /// converted to the threshold type.
```
ffarall, line 195: Love it.

pallets/file-system/src/utils.rs
```diff
@@ -126,57 +127,187 @@ where
...
-            }
-        };
+    /// Compute the eligibility value of a storage request issued at `requested_at` with `bsps_required` as
+    /// its replication target, for a specific BSP identified by `bsp_id`, and the slope with which the
+    /// eligibility value increases every tick.
+    ///
+    /// The eligibility initial value depends on the global BSP reputation weight and the storage
+    /// request's replication target, and increases linearly over time to achieve its maximum value after
+    /// `TickRangeToMaximumThreshold` ticks, a constant defined in this pallet's configuration.
+    /// Both these values get weighted with the BSP's reputation weight to give an advantage to BSPs with
```
ffarall, line 216: /// [`crate::Config::TickRangeToMaximumThreshold`] ticks, a constant defined in this pallet's configuration.

pallets/file-system/src/utils.rs
```diff
@@ -126,57 +127,187 @@ where
...
+    /// higher reputation weights, since both the initial weighted eligibility value and slope will be
+    /// higher for them.
+    ///
+    /// A BSP is eligible to volunteer for a storage request when the returned eligibility value
+    /// is greater than that BSP's volunteer threshold value.
+    ///
+    /// The formalized formulas are documented in the [README](https://github.com/Moonsong-Labs/storage-hub/blob/main/pallets/file-system/README.md#volunteering-succeeding-threshold-checks).
```
ffarall, line 223: Love it.

runtime/src/configs/runtime_params.rs
```diff
@@ -137,39 +137,77 @@ pub mod dynamic_params {
...
         #[codec(index = 18)]
         #[allow(non_upper_case_globals)]
-        pub static DefaultReplicationTarget: ReplicationTargetType = 7;
+        pub static LowSecurityReplicationTarget: ReplicationTargetType = 7;
+
+        /// The amount of BSPs that a medium security storage request should use as the replication target.
+        ///
+        /// This must be the lowest amount of BSPs that guarantee that the probability that a malicious
+        /// actor controlling 1/3 of the BSPs can hold the file hostage by controlling all its
+        /// volunteered BSPs is ~0.1%.
```
ffarall, line 153: Worth referencing somewhere how this calculations were done (with the simulations), so that we can recalculate if we adjust some parameters later on.

ffarall (approved):
Remarkable job. A few comments that I trust you can take care of.

pallets/file-system/src/lib.rs
```diff
@@ -300,9 +301,40 @@ pub mod pallet {
...
+        #[pallet::constant]
+        type LowSecurityReplicationTarget: Get<ReplicationTargetType<Self>>;
```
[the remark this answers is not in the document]
snowmead: I would leave the names. Developers are the ones interacting with our extrinsics. All this will most likely be abstracted away from the user. It is clearer to developers what they should use this way.

snowmead (approved):

pallets/file-system/src/lib.rs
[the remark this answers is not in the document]
santikaplan: I've been thinking about this and find myself in the middle. I do like the straightforward approach, but I also feel that `LowSecurity` might not be the best choice. That said, I'm not too convinced about using `Affordable` either.

I suggest the following changes:

- Change `LowSecurityReplicationTarget` to `BasicReplicationTarget`
- Change `MediumSecurityReplicationTarget` to `StandardReplicationTarget`

We can keep the existing names for the higher tiers: `HighSecurityReplicationTarget`, `SuperHighSecurityReplicationTarget`, and `UltraHighSecurityReplicationTarget`.

Let me know what you think!

pallets/file-system/src/types.rs
santikaplan:
```suggestion
/// have against a malicious actor controlling 1/3 of the total BSPs of the network.
```

runtime/src/configs/runtime_params.rs
```diff
@@ -137,39 +137,77 @@ pub mod dynamic_params {
...
 
-        /// The default number of BSPs that a storage request targets when issued by the file system pallet.
+        /// The amount of BSPs that a low security storage request should use as the replication target.
         ///
-        /// This number is a safe default so users that do not have specific requirements for replication
-        /// can still have their files stored in a way that is resilient to failures, while being small enough
-        /// to not be a burden on the network.
+        /// This must be the lowest amount of BSPs that guarantee that the probability that a malicious
+        /// actor controlling 1/3 of the BSPs can hold the file hostage by controlling all its
+        /// volunteered BSPs is ~1%.
```
santikaplan, line 144:
```suggestion
        /// volunteered BSPs is ~1%. This probability is derived using binomial distribution calculations,
        /// where the total number of BSPs is set to 1000, the fraction of malicious BSPs is 1/3, and the target number of BSPs
        /// is incremented until the probability of all selected BSPs being malicious falls below 1%. The formula used is:
        ///
        ///     num_bsps = 1000
        ///     fraction_evil = 1/3
        ///     n_evil = int(num_bsps * fraction_evil)  // = 333
        ///     target = 7
        ///     p_init = target / num_bsps
        ///     prob = binomial_cdf_at_least(n_evil, target, p_init)
        ///
        /// This ensures that the replication target is selected optimally to balance security and storage efficiency.

```

runtime/src/configs/runtime_params.rs
santikaplan: I would clarify this with the comments I added for the previous target, and don't need to comment it every single time (see above).

xcm-simulator/src/storagehub/configs/runtime_params.rs
```diff
@@ -72,50 +72,88 @@ pub mod dynamic_params {
...
 
-        /// The default number of BSPs that a storage request targets when issued by the file system pallet.
+        /// The amount of BSPs that a low security storage request should use as the replication target.
         ///
-        /// This number is a safe default so users that do not have specific requirements for replication
-        /// can still have their files stored in a way that is resilient to failures, while being small enough
-        /// to not be a burden on the network.
+        /// This must be the lowest amount of BSPs that guarantee that the probability that a malicious
+        /// actor controlling 1/3 of the BSPs can hold the file hostage by controlling all its
+        /// volunteered BSPs is ~1%.
```
santikaplan, line 79:
```suggestion
        /// volunteered BSPs is ~1%. This probability is derived using binomial distribution calculations,
        /// where the total number of BSPs is set to 1000, the fraction of malicious BSPs is 1/3, and the target number of BSPs
        /// is incremented until the probability of all selected BSPs being malicious falls below 1%. The formula used is:
        ///
        ///     num_bsps = 1000
        ///     fraction_evil = 1/3
        ///     n_evil = int(num_bsps * fraction_evil)  // = 333
        ///     target = 7
        ///     p_init = target / num_bsps
        ///     prob = binomial_cdf_at_least(n_evil, target, p_init)
        ///
        /// This ensures that the replication target is selected optimally to balance security and storage efficiency.

```

santikaplan (approved):
LGTM! Nice work!

pallets/file-system/src/utils.rs
TDemeco: Actually, for unsigned integers, `checked_div` can only return `None` if it faces a division by zero, since it can't overflow. A division that returns a result smaller than 1 just gets rounded down to 0, that's why I added the `max` with 1

pallets/file-system/src/lib.rs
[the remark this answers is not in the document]
TDemeco: Updated to reflect Santi's comments!

pallets/file-system/src/lib.rs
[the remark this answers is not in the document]
TDemeco: Updated the docs to link to the explanation

PR #335, opened 2025-01-22
feat: ✨ Apply Forest root changes to Buckets in Blockchain Service

ffarall:
This PR depends on #312.

In this PR:
1. Handle `MutationsApplied` event in `apply_forest_root_changes` of the Blockchain Service, to apply forest root changes of MSPs.
2. Remove `insert_files_metadata` call from `msp_upload_file`, now that this is done in the Blockchain Service.
3. Add `query_buckets_for_msp` new runtime API to Providers pallet.
4. Add `event_info` to `generic_apply_delta` function from `ProofsDealerInterface`. The intention is that users of this interface can pass a `Vec<u8>` of encoded info for context on what the delta is applied on. For the purpose of this PR, that is used for the File System pallet to pass down the encoded Bucket ID.


client/blockchain-service/src/utils.rs
```diff
@@ -1182,92 +1188,109 @@ where
...
+                                // In StorageHub, we assume that all `MutationsApplied` events are emitted by bucket
+                                // root changes, and they should contain the encoded `BucketId` of the bucket that was mutated
+                                // in the `event_info` field.
+                                if event_info.is_none() {
+                                    error!(target: LOG_TARGET, "MutationsApplied event with `None` event info, when it is expected to contain the BucketId of the bucket that was mutated. This should never happen. This is a bug. Please report it to the StorageHub team.");
+                                    continue;
+                                }
+                                let event_info =
+                                    event_info.expect("Just checked that this is not None; qed");
+                                let bucket_id = match BucketId::decode(&mut event_info.as_ref()) {
```
snowmead, line 1259: I think we should have a runtime api to decode this based on the current block being processed. It will avoid any breaking changes we apply to `event_info` through runtime upgrades. If from block to block the event info we insert is different then what we expect (i.e. a `BucketId`), the decode will fail.

pallets/file-system/src/utils.rs
```diff
@@ -2674,6 +2677,7 @@ where
                     &bucket_root,
                     &[(file_key, TrieRemoveMutation::default().into())],
                     &forest_proof,
+                    Some(bucket_id.encode()),
```
snowmead, line 2680: we should have a common encode and decode function (decode function would also be exposed as a runtime api).

pallets/proofs-dealer/src/lib.rs
```diff
@@ -484,10 +484,14 @@ pub mod pallet {
         /// This is the generic version of [`MutationsAppliedForProvider`](Event::MutationsAppliedForProvider)
         /// when [`generic_apply_delta`](ProofsDealerInterface::generic_apply_delta) is used
         /// and the root is not necessarily linked to a specific Provider.
+        ///
+        /// Additional information for context on where the mutations were applied can be provided
+        /// by using the `event_info` field.
```
snowmead, line 489: I would point to the common encode and decode functions if you agree to add them

snowmead:
I noticed there are no integration tests added to the [reorg integration test](https://vscode.dev/github/Moonsong-Labs/storage-hub/blob/feat/forest-root-catch-ups-msp/test/suites/integration/bsp/reorg-proof.test.ts). Was this planned for a separate PR?

ffarall:
> I noticed there are no integration tests added to the [reorg](https://vscode.dev/github/Moonsong-Labs/storage-hub/blob/feat/forest-root-catch-ups-msp/test/suites/integration/bsp/reorg-proof.test.ts). Was this planned for a separate PR?

Yes. I gave it less urgency as "handling the reorgs" is essentially already tested with the BSP tests. It is the same functionality. But the right thing to do would be also to have tests for the MSPs. We'll just do it in a different PR, to prioritise other issues and tasks.

client/blockchain-service/src/utils.rs
ffarall: Fair point. I'll add it.

snowmead (approved):
Nice job!

pallets/proofs-dealer/src/lib.rs
snowmead: Yes that makes sense.

PR #336, opened 2025-01-22
feat: Skip forest root write lock with non latest block

snowmead:
Ensure that the forest root write lock is only locked if by a task if the latest processed block matches the current best block. This avoids any possibility for a task to start with an outdated forest which would inevitably fail to operate with.


ffarall (approved):

santikaplan (approved):
LGTM!

PR #337, opened 2025-01-22
feat: msp check previous block

undercover-cactus:
In this PR, we add a check that allow msp to answer to `NewStorageRequests` when catching up to the block chain.

When closing up to the tip of the chain, a function will be called that will check for all the storage requests through a runtime api. If the storage request as the MSP id in it it will triggered a `NewStorageRequest` event. The MSP will then proceed the volunteer to it.


client/blockchain-service/src/handler.rs
```diff
@@ -1156,9 +1156,28 @@ where
...
+                        who: sr.owner,
+                        file_key: file_key.into(),
+                        bucket_id: sr.bucket_id,
+                        location: sr.location,
+                        fingerprint: Fingerprint::from(sr.fingerprint.as_bytes()),
+                        size: sr.size,
+                        user_peer_ids: sr.user_peer_ids,
+                        expires_at: sr.expires_at,
+                    })
+                }
```
snowmead, line 1180: At this point we have started any critical tasks that needed to be completed (that were in progress). 

I think we should run this asynchronously to start importing blocks right away.

undercover-cactus: We are just emitting an event here. The actual task of volunteering for a storage request will happen in an other thread. So we should be ok here ?

pallets/file-system/src/utils.rs
```diff
@@ -2398,6 +2398,22 @@ where
 
         Ok((to_succeed, threshold_slope))
     }
+
+    pub fn storage_requests_by_msp(
```
snowmead, line 2402: Leaving a trace here: We could rename this to `unresponded_storage_requests_by_msp`

client/blockchain-service/src/handler.rs
```diff
@@ -1193,9 +1193,30 @@ where
...
+
+                info!(target: LOG_TARGET, "Checking for storage requests for this MSP");
+
+                let storage_requests: Vec<(H256, StorageRequestMetadata)> = self
+                    .client
+                    .runtime_api()
+                    .unresponded_storage_requests_by_msp(block_hash, msp_id)
+                    .unwrap();
+
+                for (file_key, sr) in storage_requests {
```
links234, line 1208: _maybe ignore since it's personal preference_

Can you rename `sr` to `storage_request`?

node/src/tasks/bsp_charge_fees.rs
```diff
@@ -272,7 +272,7 @@ where
             .map_err(|e| anyhow!("Failed to get metadata from Forest: {:?}", e))?;
 
         if !user_files.is_empty() {
-            let (file_key, metadata) = user_files.first().expect("User files is not empty");
+            let (file_key, metadata) = user_files.first().expect("User files is not empty"); // NOTE: WHY ONLY THE FIRST ONE ?
```
links234, line 275: Now that we know why, you can add a comment with explanation.

pallets/file-system/src/utils.rs
```diff
@@ -2487,6 +2487,22 @@ where
         }
         .file_key::<FileKeyHasher<T>>()
     }
+
+    pub fn unresponded_storage_requests_by_msp(
```
links234, line 2491: `pending_` ? :D

links234 (commented):
Left some preliminary comments.

client/blockchain-service/src/handler.rs
undercover-cactus: It is hard to differentiate `storage_requests` and `storage_request` when browsing the code. I realize that `sr` is not a good name but we can quickly understand what it is by looking at the loop. It is a trade off and I guess linked to programming habits.

Maybe to improve readability I could add a comment ?

node/src/tasks/bsp_charge_fees.rs
undercover-cactus: Oh, right I forgot about that.

pallets/file-system/src/utils.rs
undercover-cactus: `unresponded_` was @snowmead proposition but I think I like `pending_` and will switch to that.

client/blockchain-service/src/handler.rs
ffarall: Adding a comment is fine by me, btw.

pallets/file-system/runtime-api/src/lib.rs
```diff
@@ -45,19 +46,27 @@ pub enum QueryConfirmChunksToProveForFileError {
     ChallengedChunkToChunkIdError,
 }
 
+/// Error type for the `pending_storage_requests_by_msp`.
+#[derive(Eq, PartialEq, Encode, Decode, RuntimeDebug, TypeInfo)]
+pub enum StorageRequestsByMSPError {
```
ffarall, line 51:
```suggestion
pub enum StorageRequestsByMspError {
```

pallets/file-system/src/utils.rs
```diff
@@ -2487,6 +2487,22 @@ where
         }
         .file_key::<FileKeyHasher<T>>()
     }
+
+    pub fn pending_storage_requests_by_msp(
+        msp_id: ProviderIdFor<T>,
+    ) -> Vec<(MerkleHash<T>, StorageRequestMetadata<T>)> {
+        // Get the storeage requests for a specific MSP
```
ffarall, line 2494:
```suggestion
        // Get the storage requests for a specific MSP
```

pallets/file-system/runtime-api/src/lib.rs
```diff
@@ -45,19 +46,27 @@ pub enum QueryConfirmChunksToProveForFileError {
...
         FileKey: Codec,
         TickNumber: Codec,
         ChunkId: Codec,
+        StorageRequestMetadata: Codec,
     {
         fn is_storage_request_open_to_volunteers(file_key: FileKey) -> Result<bool, IsStorageRequestOpenToVolunteersError>;
         fn query_earliest_file_volunteer_tick(bsp_id: BackupStorageProviderId, file_key: FileKey) -> Result<TickNumber, QueryFileEarliestVolunteerTickError>;
         fn query_bsp_confirm_chunks_to_prove_for_file(bsp_id: BackupStorageProviderId, file_key: FileKey) -> Result<Vec<ChunkId>, QueryBspConfirmChunksToProveForFileError>;
         fn query_msp_confirm_chunks_to_prove_for_file(msp_id: MainStorageProviderId, file_key: FileKey) -> Result<Vec<ChunkId>, QueryMspConfirmChunksToProveForFileError>;
+        fn pending_storage_requests_by_msp(msp_id: MainStorageProviderId) -> Vec<(H256, StorageRequestMetadata)>;
```
ffarall, line 70: Change this so that it returns a `sp_std::collections::btree_map::BTreeMap`.

node/src/tasks/bsp_charge_fees.rs
```diff
@@ -272,6 +272,8 @@ where
             .map_err(|e| anyhow!("Failed to get metadata from Forest: {:?}", e))?;
 
         if !user_files.is_empty() {
+            // We only take the first file of the list in order to generate a proof submit it with an extrinsic and then release the lock to process the next file and generate the next proof.
+            // It is not ideal because it means one extrinsic per file but batch deletion is not yet implemented.
```
ffarall, line 276: Good comment. I'd add a `TODO` to it, so that it is not forgotten and easier to find.

test/suites/integration/msp/catch-up-storage.test.ts
```diff
@@ -0,0 +1,99 @@
+import assert, { strictEqual } from "node:assert";
+import { describeMspNet, shUser, type EnrichedBspApi, sleep } from "../../../util";
+
+describeMspNet(
+  "MSP catching up with chain and volunteering for storage request",
+  { initialised: false },
```
ffarall, line 6: No need, non-initialised is the default value as far as I remember.

test/suites/integration/msp/catch-up-storage.test.ts
```diff
@@ -0,0 +1,99 @@
...
+      mspApi = maybeMspApi;
+    });
+
+    it("Network launches and can be queried", async () => {
+      const userNodePeerId = await userApi.rpc.system.localPeerId();
+      strictEqual(userNodePeerId.toString(), userApi.shConsts.NODE_INFOS.user.expectedPeerId);
+
+      const mspNodePeerId = await mspApi.rpc.system.localPeerId();
+      strictEqual(mspNodePeerId.toString(), userApi.shConsts.NODE_INFOS.msp1.expectedPeerId);
+    });
```
ffarall, line 25: We can remove this one, it's everywhere 😂

test/suites/integration/msp/catch-up-storage.test.ts
```diff
@@ -0,0 +1,99 @@
...
+        source,
+        destination,
+        userApi.shConsts.NODE_INFOS.user.AddressId,
+        newBucketEventDataBlob.bucketId
+      );
+
+      await userApi.docker.pauseBspContainer("docker-sh-msp-1");
+
+      // We need to wait so it won't try to answer the request storage
+      await sleep(5000);
```
ffarall, line 48: Why do you need to wait here? It won't respond if it is paused.

test/suites/integration/msp/catch-up-storage.test.ts
```diff
@@ -0,0 +1,99 @@
...
+      // Required to trigger out of sync mode
+      // await userApi.rpc.engine.createBlock(true, true);
+
+      // await waitFor({
+      //   lambda: async () =>
+      //     (await newMspApi.rpc.storagehubclient.isFileInFileStorage(event.data.fileKey)).isFileFound
+      // });
+
+      // await userApi.assert.eventPresent("fileSystem", "MspAcceptedStorageRequest");
+    });
```
ffarall, line 97: What happened here? This is not really testing the right outcome 😂.

client/blockchain-service/src/handler.rs
```diff
@@ -1156,9 +1156,28 @@ where
                 // TODO: Send events to check that this node has a Forest Storage for the BSP that it manages.
                 // TODO: Catch up to Forest root writes in the BSP Forest.
             }
-            Some(StorageProviderId::MainStorageProvider(_msp_id)) => {
+            Some(StorageProviderId::MainStorageProvider(msp_id)) => {
                 // TODO: Send events to check that this node has a Forest Storage for each Bucket this MSP manages.
                 // TODO: Catch up to Forest root writes in the Bucket's Forests.
+
+                let storage_requests: Vec<(H256, StorageRequestMetadata)> = self
```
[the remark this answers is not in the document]
santikaplan: Hey @undercover-cactus! Just a quick tip for PR comments in general: please don’t resolve comments if no action has been taken. If there’s nothing to do, it’s best to leave a reply explaining why instead of resolving it.
Ideally, the reviewer should be the one to resolve their own comments, unless it’s something simple and directly addressed. This helps a lot when re-reviewing to track what’s actually been handled. Thanks!

test/suites/integration/msp/catch-up-storage.test.ts
undercover-cactus: I would rather leave it because otherwise there is no way to know explicitly.

test/suites/integration/msp/catch-up-storage.test.ts
undercover-cactus: Why can we remove it ?

test/suites/integration/msp/catch-up-storage.test.ts
undercover-cactus: I had the behavior that if even if the container was paused I had a NewStorageRequest event from the MSP. So I added a delay to be sure it is paused before creating the new storage request.

test/suites/integration/msp/catch-up-storage.test.ts
undercover-cactus: The MSP being offline the user cannot send it the file and the MSP can never finish the task because it doesn't have the file.

#361 will solve this by downloading the file from a BSP in this particular case. The commented part in this test should then pass. It is commented so the CI would pass.

test/suites/integration/msp/catch-up-storage.test.ts
ffarall: It's testing the connections to the nodes, and we are testing this in pretty much every test. It doesn't hurt that it's there, but at this point it's not necessary really.

test/suites/integration/msp/catch-up-storage.test.ts
ffarall: You could pause the container even before creating the bucket, or even add a check to see that it is not responding. But waiting 5s will unnecessarily slow down the integration test suite.

client/blockchain-service/src/handler.rs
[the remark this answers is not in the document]
ffarall: You don't need to use `HashMap`, you can use `BTreeMap` instead, which is available in the runtime, under the `sp_std` crate.

test/suites/integration/msp/catch-up-storage.test.ts
```diff
@@ -0,0 +1,97 @@
+import assert, { strictEqual } from "node:assert";
+import { describeMspNet, shUser, type EnrichedBspApi, sleep } from "../../../util";
```
ffarall, line 2: The CI is complaining about this unused `sleep` import.

runtime/src/apis.rs
```diff
@@ -24,6 +25,7 @@ use sp_runtime::{
     transaction_validity::{TransactionSource, TransactionValidity},
     ApplyExtrinsicResult, ExtrinsicInclusionMode,
 };
+use sp_std::collections::btree_map::BTreeMap;
 use sp_std::prelude::Vec;
```
ffarall, line 29: Ideally, merge these two `use` statements into one.

ffarall (commented):
Commenting and not approving due to the missing test. But good job.

node/src/tasks/user_sends_file.rs
```diff
@@ -411,7 +429,25 @@ where
                             // Wait for a short time before retrying
                             tokio::time::sleep(std::time::Duration::from_secs(1)).await;
```
undercover-cactus, line 430: In order to standardize the retrying logic, should we also wait for one block here instead of one second ? Because `Refused` happened if the request hasn't been yet processed.

And because of latency and block propagation time, one second could be too short in a real network.

snowmead: Isn't this flow independent from the blockchain. The user here is trying to send data to a peer (provider).

pallets/file-system/runtime-api/src/lib.rs
```diff
@@ -53,21 +55,29 @@ pub enum GenericApplyDeltaEventInfoError {
...
         ChunkId: Codec,
         GenericApplyDeltaEventInfo: Codec,
+        StorageRequestMetadata: Codec,
     {
         fn is_storage_request_open_to_volunteers(file_key: FileKey) -> Result<bool, IsStorageRequestOpenToVolunteersError>;
         fn query_earliest_file_volunteer_tick(bsp_id: BackupStorageProviderId, file_key: FileKey) -> Result<TickNumber, QueryFileEarliestVolunteerTickError>;
         fn query_bsp_confirm_chunks_to_prove_for_file(bsp_id: BackupStorageProviderId, file_key: FileKey) -> Result<Vec<ChunkId>, QueryBspConfirmChunksToProveForFileError>;
         fn query_msp_confirm_chunks_to_prove_for_file(msp_id: MainStorageProviderId, file_key: FileKey) -> Result<Vec<ChunkId>, QueryMspConfirmChunksToProveForFileError>;
         fn decode_generic_apply_delta_event_info(encoded_event_info: Vec<u8>) -> Result<GenericApplyDeltaEventInfo, GenericApplyDeltaEventInfoError>;
+        fn pending_storage_requests_by_msp(msp_id: MainStorageProviderId) -> BTreeMap<H256, StorageRequestMetadata>;
```
ffarall, line 81: We need to add this runtime API to the `types-bundle` package so that we can access from TS if needed.

client/blockchain-service/src/handler.rs
```diff
@@ -1264,11 +1286,48 @@ where
...
+                        file_key: file_key.into(),
+                        bucket_id: sr.bucket_id,
+                        location: sr.location,
+                        fingerprint: Fingerprint::from(sr.fingerprint.as_bytes()),
+                        size: sr.size,
+                        user_peer_ids: sr.user_peer_ids,
+                        expires_at: sr.expires_at,
+                    })
+                }
+
```
ffarall, line 1330: With the new MSP/BSP separation this should go inside `msp_initial_sync()`

ffarall (commented):
Approving with some comments. I trust you'll take care of it.

ffarall (approved):

pallets/file-system/runtime-api/src/lib.rs
undercover-cactus: I will skip this for now. The type `StorageRequestMetadata` is not defined in typescript as far as I can tell and many of the types needed in this one are also missing. Once we will need it in a test I will implement it.

PR #338, opened 2025-01-23
fix: :bug: update spam handling to avoid inconsistencies after runtime upgrades

TDemeco:
This PR changes the way we detect if the chain is currently being spammed or not to solve [issue #26](https://github.com/Moonsong-Labs/internal-storage-hub-design-audit/issues/26):
- Currently, we have a storage value that contains the count of not-full blocks in the last `BlockFullnessPeriod` blocks. We add one to it if the previous block (`current_block - 1`) was not full, and substract one from it if the block `current_block - BlockFullnessPeriod - 1` was not full.
- Now we have instead a storage value `PastBlocksStatus` that contains a bounded vector of boolean elements and size `BlockFullnessPeriod`. Each index of the vector represents a block (the first element represents the oldest block of the set `current_block - BlockFullnessPeriod - 1`, the last element the newest block of the set `current_block - 1`) and each value represents whether that block was considered full or not.
- Each new block where the chain is not current undergoing a multiblock migration (so the `on_poll` hook gets executed):
  - The oldest block of the vector gets removed (only if the vector is full, to avoid errors when starting from genesis).
  - The previous block gets checked to see if it was considered spammed. If it was, pushes a `true` value to the bounded vector, otherwise pushes a `false` value.
  - After the changes to the bounded vector, the amount of non-full blocks get counted and, depending if it's under the minimum stipulated by the runtime or not, the chain is considered spammed or not, and the `ChallengesTickerPaused` storage gets set or cleared accordingly.
  - Finally, update the `PastBlocksStatus` storage with the new vector.

_Note: ideally, we would use a `BitVec` to be more efficient storage-wise, but there's no current implementation of a bounded `BitVec`, and since we are modifying the vector in every block, it gets included in the PoV so it cannot be unbounded_


snowmead (approved):

api-augment/dist/types/interfaces/augment-api-query.d.ts
```diff
@@ -1372,11 +1372,13 @@ declare module "@polkadot/api-base/types/storage" {
       lastDeletedTick: AugmentedQuery<ApiType, () => Observable<u32>, []> &
         QueryableStorageEntry<ApiType, []>;
       /**
-       * The number of blocks that have been considered _not_ full in the last [`Config::BlockFullnessPeriod`].
+       * The vector holding whether the last [`Config::BlockFullnessPeriod`] blocks were full or not.
        *
-       * This is used to check if the network is presumably under a spam attack.
+       * Each element in the vector represents a block, and is `true` if the block was full, and `false` if it was not.
```
santikaplan, line 1377:
```suggestion
       * Each element in the vector represents a block, and is `true` if the block was full, and `false` otherwise.
```

pallets/proofs-dealer/src/benchmarking.rs
```diff
@@ -361,14 +362,33 @@ mod benchmarks {
...
+            - One::one();
         PastBlocksWeight::<T>::insert(old_block, max_weight_for_class);
 
-        // Setting `NotFullBlocksCount` to be just exactly the minimum, so that with this increment the
-        // chain is considered to be not spammed.
-        let min_not_full_blocks_to_spam = Pallet::<T>::calculate_min_non_full_blocks_to_spam();
-        NotFullBlocksCount::<T>::set(min_not_full_blocks_to_spam);
+        // Setting the `PastBlocksStatus` bounded vector to contain as the first element a block considered full, then
+        // exactly the minimum non full blocks, and then all full blocks, so that when adding the new non full block
+        // the chain goes from being considered spammed to not spammed.
```
santikaplan, line 372:
```suggestion
        // Setting the `PastBlocksStatus` bounded vector to contain, as the first element, a block considered full, followed by
        // exactly the minimum required non-full blocks, and then all full blocks, so that when adding the new non-full block,
        // the chain transitions from being considered spammed to not spammed.
```

api-augment/src/interfaces/augment-api-query.ts
```diff
@@ -1314,11 +1314,13 @@ declare module "@polkadot/api-base/types/storage" {
       lastDeletedTick: AugmentedQuery<ApiType, () => Observable<u32>, []> &
         QueryableStorageEntry<ApiType, []>;
       /**
-       * The number of blocks that have been considered _not_ full in the last [`Config::BlockFullnessPeriod`].
+       * The vector holding whether the last [`Config::BlockFullnessPeriod`] blocks were full or not.
        *
-       * This is used to check if the network is presumably under a spam attack.
+       * Each element in the vector represents a block, and is `true` if the block was full, and `false` if it was not.
```
santikaplan, line 1319:
```suggestion
       * Each element in the vector represents a block, and is `true` if the block was full, and `false` otherwise.
```

pallets/proofs-dealer/src/lib.rs
```diff
@@ -377,11 +377,14 @@ pub mod pallet {
     pub type PastBlocksWeight<T: Config> =
         StorageMap<_, Blake2_128Concat, BlockNumberFor<T>, Weight>;
 
-    /// The number of blocks that have been considered _not_ full in the last [`Config::BlockFullnessPeriod`].
+    /// The vector holding whether the last [`Config::BlockFullnessPeriod`] blocks were full or not.
     ///
-    /// This is used to check if the network is presumably under a spam attack.
+    /// Each element in the vector represents a block, and is `true` if the block was full, and `false` if it was not.
```
santikaplan, line 382:
```suggestion
    /// Each element in the vector represents a block, and is `true` if the block was full, and `false` otherwise.
```

santikaplan (approved):
All good! Some minor comments, good to merge!

PR #27, opened 2024-03-29
feat: upgrade ^ polkadot-v1.7.0, enable async backing

snowmead:
This PR:

- Moves all dependencies to root of the workspace.
- Upgrades polkadot crates from `polkadot-v1.5.0` > `polkadot-v1.7.0`.
- Updates storage-hub in various places (client + runtime) to enable async backing by following Web3 Foundation's docs _(updated last Feb 27, 2024)_.

PR #339, opened 2025-01-23
fix: 🚑 remove `Pays::No` from extrinsics that shouldn't be free

TDemeco:
This PR changes some extrinsics that were previously free to not be free anymore if there was no solid argument for them to be free.
It also adds extra docs for the ones that remain free to explain in more detail the reasoning for those extrinsics to not charge the callers.

Fixes [issue #27](https://github.com/Moonsong-Labs/internal-storage-hub-design-audit/issues/27).


snowmead (approved):

santikaplan (approved):
All good here!

PR #341, opened 2025-01-24
fix: Skip BSP confirm file keys that have failed to update or create a payment stream

snowmead:
Moved the payment stream create and update block in `bsp_confirm_storing` to skip file keys if those operations have failed.

Moving these operations at the top avoids any possibility for any storage elements from being updated when they shouldn't if we skipped the file key.

See the audit issue raised for more details.

Resolves [issue #30](https://github.com/Moonsong-Labs/internal-storage-hub-design-audit/issues/30)

### DIFF
--- a/pallets/file-system/src/utils.rs
+++ b/pallets/file-system/src/utils.rs
@@ -1382,6 +1382,10 @@ where
 
             // Check if a payment stream between the user and provider already exists.
             // If it does not, create it. If it does, update it.
+            //
+            // We ignore all errors from the payment stream operations (create/update) and add the file key to the skipped_file_keys set.
+            // This operation must be the first one to be executed before we being updating storage elements as to avoid any potential case
+            // where a storage element is updated but should not be.
             match <T::PaymentStreams as PaymentStreamsInterface>::get_dynamic_rate_payment_stream_amount_provided(&bsp_id, &storage_request_metadata.owner) {
 				Some(previous_amount_provided) => {
 					// Update the payment stream.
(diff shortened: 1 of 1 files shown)


pallets/file-system/src/utils.rs
```diff
@@ -1242,6 +1242,50 @@ where
                 Error::<T>::InsufficientAvailableCapacity
             );
 
+            // Check if a payment stream between the user and provider already exists.
+            // If it does not, create it. If it does, update it.
+            //
+            // We ignore all errors from the payment stream operations (create/update) and add the file key to the skipped_file_keys set.
+            // This operation must be the first one to be executed before we being updating storage elements as to avoid any potential case
```
santikaplan, line 1249:
```suggestion
            // All errors from the payment stream operations (create/update) are ignored, and the file key is added to the `skipped_file_keys` set.
            // This operation must be executed first, before updating any storage elements, to prevent potential cases
```

santikaplan (approved):
LGTM!

pallets/file-system/src/utils.rs
TDemeco:
```suggestion
            // All errors from the payment stream operations (create/update) are ignored, and the file key is added to the `skipped_file_keys` set instead of erroring out.
            // This is done to avoid a malicious user, owner of one of the files from the batch of confirmations, being able to prevent the BSP from confirming any files by making itself insolvent so payment stream operations fail.
            // This operation must be executed before updating any storage elements related to the file storage confirmation since those updates should also be skipped.
```

TDemeco (approved):
Looks good! 💯

PR #342, opened 2025-01-27
feat: Add new chunk count column for tracking

snowmead:
Adds new `CHUNK_COUNT_COLUMN` to track number of chunks for a given file key. This avoids having to iterate through the entire file trie to count all the chunks and conclude if we have uploaded all the chunks yet for a file.

These changes also include a refactor of our column definitions to have consistent index retrievals and column length constant. In addition, I added some documentation to these columns to have a clear definition for their purpose in the context of RocksDB file storage.

Resolves [issue #20](https://github.com/Moonsong-Labs/internal-storage-hub-design-audit/issues/20)


client/file-manager/src/in_memory.rs
```diff
@@ -352,11 +346,13 @@ where
             .as_str(),
         );
 
-        // Check if we have all the chunks for the file.
-        let stored_chunks = file_data
-            .stored_chunks_count()
-            .map_err(|_| FileStorageWriteError::FailedToConstructTrieIter)?;
-        if metadata.chunks_count() != stored_chunks {
+        // Increment chunk count
+        let current_count = self.chunk_counts.get(file_key).copied().unwrap_or(0);
```
TDemeco, line 350: Shouldn't we fail if we can't get the stored chunk count? I'd do a function similar to the one that we previously used here `stored_chunks_count` that returns a result

client/file-manager/src/in_memory.rs
```diff
@@ -352,11 +346,13 @@ where
...
+        let new_count = current_count + 1;
+        self.chunk_counts.insert(*file_key, new_count);
```
TDemeco, line 352: This probably can't fail (we should check that the file is not big enough in a previous step) but just in case I'd use safe math here.

client/file-manager/src/rocksdb.rs
```diff
@@ -356,29 +383,35 @@ where
...
+                break;
+            }
+
+            trie.remove(&chunk_id_struct.as_trie_key()).map_err(|e| {
+                error!(target: LOG_TARGET, "Failed to delete chunk from RocksDb: {}", e);
+                FileStorageWriteError::FailedToDeleteChunk
+            })?;
+
+            chunk_id += 1;
         }
```
TDemeco, line 409: To avoid looping (which could introduce issues and is not as clear), in this scenario considering we already built the trie I wouldn't mind counting its elements as we did before

client/file-manager/src/rocksdb.rs
```diff
@@ -20,23 +20,50 @@ use crate::{
     LOG_TARGET,
 };
 use codec::{Decode, Encode};
+use strum::{EnumCount, VariantArray};
+
+#[derive(Debug, Clone, Copy, EnumCount, VariantArray)]
```
TDemeco, line 25: Love the refactor 💯

client/file-manager/src/rocksdb.rs
```diff
@@ -20,23 +20,50 @@ use crate::{
...
+pub enum Column {
+    Metadata,
```
TDemeco, line 27: We should have docs for the `Metadata` member as well, and if possible for all the columns

client/file-manager/src/rocksdb.rs
```diff
@@ -58,7 +85,8 @@ fn open_or_creating_rocksdb(db_path: String) -> io::Result<kvdb_rocksdb::Databas
     Ok(db)
 }
 
-/// Storage backend for RocksDB.
+/// Storage backend implementation for RocksDB.
```
TDemeco, line 88: Love the improved docs on all functions!

TDemeco (approved):
Nice job! Left some comments but overall great 💯

undercover-cactus (approved):

client/file-manager/src/rocksdb.rs
snowmead: I am not sure I understand, because in the end we will have to loop over every chunkId to delete it from storage.

client/file-manager/src/in_memory.rs
snowmead: you mean for the overflow (current_count + 1)?

snowmead: now that I am looking at it why isn't `file_size` a `u128`?

https://github.com/Moonsong-Labs/storage-hub/blob/d598f04e20f8d914220b081375b0c66aa0682259/primitives/file-metadata/src/lib.rs#L26

client/file-manager/src/in_memory.rs
snowmead: the file key in `chunk_counts` only exist if there is at least one chunk stored. so I just `unwrap_or` here to start at 0

TDemeco: You have to call the `insert_file` method above before starting to write chunks, otherwise the check above this line
```
let file_data = self
            .file_data
            .get_mut(file_key)
            .ok_or(FileStorageWriteError::FileDoesNotExist)?;
```
would fail, wont it? And in the `insert_file` method you initialize `chunk_counts` to 0, so I believe not having `chunk_counts` initialized at the moment you start writing chunks should be an error.

client/file-manager/src/in_memory.rs
TDemeco: Yep, to avoid not handling the overflow. Regarding file size, u64 allows us to have files of up to 18.44 Exabytes, which we believed to be way more than enough

client/file-manager/src/rocksdb.rs
TDemeco: I was refering to use the `loop` keyword instead of a for loop

client/file-manager/src/in_memory.rs
snowmead: good point!

client/file-manager/src/rocksdb.rs
snowmead: I am using a `loop`. not a for loop.

client/file-manager/src/rocksdb.rs
```diff
@@ -571,65 +627,70 @@ where
...
+            error!(target: LOG_TARGET, "{:?}", e);
+            FileStorageWriteError::FailedToGetStoredChunksCount
+        })?;
+
+        // Increment chunk count.
+        // This should never overflow unless there is a bug or we support file sizes as large as 16 exabytes.
+        let new_count = current_count
+            .checked_add(1)
+            .ok_or(FileStorageWriteError::ChunkCountOverflow)?;
+        transaction.put(
```
links234, line 647: What happens when there are 2 threads writing different chunks to the same file? Do we have this case somewhere or is it protected by mutex a layer above?

If we don't have a lock above this, than we can end up with the wrong chunk count number.

links234 (changes requested):
Looks good but seems to me we need another locker for the `write_chunk` so that we don't end up with wrong `chunk_count` value.w

client/file-manager/src/rocksdb.rs
snowmead: added a note explaining why it is safe since this is executed in the context of a write lock

links234 (approved):

client/file-manager/src/traits.rs
```diff
@@ -15,10 +15,14 @@ pub enum FileStorageWriteError {
     FailedToGetFileChunk,
     /// File metadata fingerprint does not match the stored file fingerprint.
     FingerprintAndStoredFileMismatch,
+    /// Failed to construct file trie.
+    FailedToContructFileTrie,
```
ffarall, line 19:
```suggestion
    FailedToConstructFileTrie,
```

client/file-manager/src/rocksdb.rs
```diff
@@ -20,23 +20,54 @@ use crate::{
...
+    ///
+    /// Used for deleting all files in a bucket efficiently.
+    BucketPrefix,
+    /// Exclude* columns stores keys of 32 bytes representing the `file_key` with empty values.
+    ///
+    /// These columns are used primarily to mark file keys as being excluded from certain operations.
+    ExcludeFile,
+    ExcludeUser,
+    ExcludeBucket,
+    ExcludeFingerprint,
```
ffarall, line 52: I'd copy and paste the doc comment for the other items, or make one for each. Otherwise this doc comment will only be linked to `ExcludeFile`. So for example, if you hover over `ExcludeBucket`, you won't see any doc comment.

ffarall (approved):
Amazing job here! Loved the added docs. Much needed. Just left a minor comment.

client/file-manager/src/rocksdb.rs
snowmead: good point

snowmead: I'll add this in another PR to avoid restarting CI for such a small change

PR #31, opened 2024-04-03
feat: :label: Added typegen feature

timbrinded:

PR #343, opened 2025-01-28
fix: 🚑 change how the storage request creation deposit is calculated

TDemeco:
This PR changes the storage request creation deposit from a constant to the result of the equation `storage_request_deposit = num_of_bsps * bsp_volunteer_fee + base_storage_request_creation_deposit`, and adds the logic to give every volunteering BSP enough funds to pay for the fee it paid to send the `bsp_volunteer()` extrinsic.

This is done to avoid a possible scenario where a malicious user could DoS the network for very cheap by issuing storage requests but never sending the files to the volunteering BSPs. For more details, check out [issue #31](https://github.com/Moonsong-Labs/internal-storage-hub-design-audit/issues/31) (which this PR solves).

The base deposit was added to avoid having issues when starting the network (since BSPs could register after the storage request was issued and volunteer, making the deposit not enough to pay them) and also to account for changes in the fee that `bsp_volunteer()` costs (since it can change depending on the fullness of the previous blocks).

### UPDATE: 
This PR was expanded with new changes:
- Users that want to issue a storage request now have to pay an upfront amount to the treasury. This is added as a deterrent to avoid users spamming the network with storage requests for huge files that they don't actually intend to store (and are only requesting their storage so BSPs inccur the cost of retrieving, storing and deleting those files).
  - The upfront amount that a user has to pay is defined by the equation: `current_price_per_giga_unit_per_tick * file_size_in_gigabytes * amount_of_ticks_to_pay_upfront * replication_target_chosen`.
  - In a typical scenario with the current runtime setup for a 1GB file:
    - `current_price_per_giga_unit_per_tick = 50 NANOUNITs`
    - `file_size_in_gigabytes = 1`
    - `amount_of_ticks_to_pay_upfront = 72000` (5 days at 6 seconds per tick)
    - `replication_target_chosen = 12` (Defined as the `StandardReplicationTarget`)
    - Then the upfront cost is 0.0432 UNITs

This PR then also solves [issue #36](https://github.com/Moonsong-Labs/internal-storage-hub-design-audit/issues/36)


pallets/file-system/src/tests.rs
```diff
@@ -2026,6 +2025,16 @@ mod request_storage {
...
+				// Calculate the deposit that the user is going to have to pay to issue this storage request.
+				let number_of_bsps = <<Test as crate::Config>::Providers as ReadStorageProvidersInterface>::get_number_of_bsps();
+        		let number_of_bsps_balance_typed =
+            		<Test as crate::Config>::ReplicationTargetToBalance::convert(number_of_bsps);
+       			let storage_request_deposit = <<Test as crate::Config>::WeightToFee as sp_weights::WeightToFee>::weight_to_fee(
+			&<Test as crate::Config>::WeightInfo::bsp_volunteer(),
+        		)
+        		.saturating_mul(number_of_bsps_balance_typed)
+        		.saturating_add(<Test as crate::Config>::BaseStorageRequestCreationDeposit::get());
+
```
snowmead, line 2037: Can we replace this with a helper function.

snowmead (approved):
nice job!

pallets/file-system/src/tests.rs
TDemeco: Done!

pallets/bucket-nfts/src/mock.rs
```diff
@@ -242,6 +243,8 @@ impl ProofsDealerInterface for MockProofsDealer {
     }
 }
 
+pub(crate) type ReplicationTargetType = u32;
```
ffarall, line 246: Shouldn't you be using this down in the pallet `Config`?

ffarall (approved):
Exemplary work as usual

PR #344, opened 2025-01-30
feat: 📈 finish benchmarking the `storage-providers` pallet's extrinsics

TDemeco:
This PR benchmarks the two remaining unbenchmarked extrinsics from the `storage-providers` pallet (`top_up_deposit` and `delete_provider`). To do so, it:
- Adds two variables to the `MainStorageProvider` type that holds the metadata of a MSP:
  - `amount_of_value_props`: the total amount of value propositions that this MSP has added to its repertoire. This is used to know how much weight is going to be needed to clean those value propositions from storage.
    - Logic to add or substract one to this value for tracking was added where needed.
  - `amount_of_buckets`: the current amount of buckets that the MSP is storing. This is also used to know how much weight is going to be needed to delete or move those buckets.
    - This variable uses a type defined in the pallet's configuration (`BucketCount`) to allow for flexibility when defining the maximum amount of buckets the user of the pallet deems a MSP can store.
    - Logic to add or substract one to this value for tracking was added where needed.
    - **_Note: currently, the buckets are not being moved nor deleted when cleaning up storage. This should be designed and implemented in another PR._**
- Adds comments to all fields of the MSP's metadata, for clarity.
- Fixes the `msp_sign_off` extrinsic. Previously it was not clearing up the storage occupied by the MSP's value propositions, now it correctly clears everything.
  - Since this changes the weight requirement of the extrinsic the benchmark was refactored to account for this change, and now the required weight is a function of the amount of value propositions the MSP has.
- Fixes the `delete_provider` extrinsic. Like `msp_sign_off`, it was not clearing up the storage used by the value propositions if the provider to delete was a MSP, but this time because `drain_prefix` was used but the iterator was never iterated so no elements were deleted.
  - Also added a cleanup of the `MainStorageProviderIdsToBuckets` storage, since with this extrinsic a MSP can be deleted while still having buckets. _Like it was mentioned, specific logic to delete or move these buckets should be implemented in another PR._
  - Also added a call to stop all BSP cycles before deleting it, to avoid a non-existent BSP having to submit proofs and randomness seeds.
  - Updates the extrinsic test to check for all this scenarios.
- Adds the benchmarks for both `top_up_deposit` and `delete_provider`.

This is part of a fix for [issue #24](https://github.com/Moonsong-Labs/internal-storage-hub-design-audit/issues/24)


pallets/providers/src/utils.rs
```diff
@@ -1123,11 +1151,34 @@ where
...
         if let Some(msp) = MainStorageProviders::<T>::get(provider_id) {
+            // Remove the Provider from the InsolventProviders storage
             InsolventProviders::<T>::remove(StorageProviderId::<T>::MainStorageProvider(
                 *provider_id,
             ));
+
+            // Delete the MSP's data
             MainStorageProviders::<T>::remove(&provider_id);
             AccountIdToMainStorageProviderId::<T>::remove(msp.owner_account);
+            // TODO: The buckets of this MSP should be moved to another one somehow, we can't leave the user without its data because
```
snowmead, line 1162: We could emit an event with all buckets that were being stored by the deleted MSP so users can recover them with move bucket requests.

runtime/src/configs/mod.rs
```diff
@@ -498,6 +498,7 @@ impl pallet_storage_providers::Config for Runtime {
     type StorageDataUnit = StorageDataUnit;
     type StorageDataUnitAndBalanceConvert = StorageDataUnitAndBalanceConverter;
     type SpCount = u32;
+    type BucketCount = u32;
```
snowmead, line 501: Can we set this to be `u128` to avoid any chances of overflowing?

santikaplan (approved):
Beast!!

pallets/providers/src/utils.rs
TDemeco: Good call, I added that!

runtime/src/configs/mod.rs
TDemeco: Done!

PR #345, opened 2025-01-30
feat: :chart_with_upwards_trend: benchmark `on_idle` hook of `pallet-storage-providers`

TDemeco:
This PR refactors the `on_idle` hook to make it benchmarkeable and to match the structure of the `on_idle` hook of `pallet-file-system` and adds the benchmarks for both when the provider to mark as insolvent is a BSP and for when it's a MSP.

This is a draft because it has pending ToDos:
- [ ] Wait for #344 to be merged.
- [ ] Benchmark the whole `pallet-storage-providers`, update weights and uncomment lines which use the calculated weights.

### DIFF
--- a/pallets/providers/src/utils.rs
+++ b/pallets/providers/src/utils.rs
@@ -2521,11 +2521,9 @@ where
 
 mod hooks {
     use crate::{
-        pallet,
-        types::{ShTickGetter, StorageHubTickNumber},
-        utils::StorageProviderId,
-        AwaitingTopUpFromProviders, BackupStorageProviders, Event, HoldReason, InsolventProviders,
-        MainStorageProviders, NextStartingShTickToCleanUp, Pallet, ProviderTopUpExpirations,
+        pallet, types::StorageHubTickNumber, utils::StorageProviderId, AwaitingTopUpFromProviders,
+        Event, HoldReason, InsolventProviders, NextStartingShTickToCleanUp, Pallet,
+        ProviderTopUpExpirations,
     };
 
     use frame_support::{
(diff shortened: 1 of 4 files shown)


pallets/providers/src/utils.rs
```diff
@@ -2534,149 +2532,202 @@ mod hooks {
...
+                let mut expired_provider_top_ups =
+                    ProviderTopUpExpirations::<T>::take(tick_to_process);
+                meter.consume(db_weight.reads_writes(1, 1));
+
+                // Get the required weight to process an expired provider top up in its worst case scenario.
+                let maximum_required_weight_expired_provider_top_up = db_weight.reads_writes(2, 1);
+                /* T::WeightInfo::process_expired_provider_top_up_bsp()
+                .max(
+                    T::WeightInfo::process_expired_provider_top_up_msp(),
+                ); */
```
snowmead, line 2599: can we remove this?

pallets/providers/src/utils.rs
```diff
@@ -2534,149 +2532,202 @@ mod hooks {
...
+                        ran_out_of_weight = true;
+                        break;
+                    }
+                }
 
-            // TODO: After benchmarking, we should check before this loop that there is enough remaining weight to
-            // TODO: process all the expired move bucket requests. If not, we should return early.
-            while let Some(typed_provider_id) = provider_top_up_expirations.pop() {
-                Self::process_expired_provider_top_up_period(typed_provider_id, meter);
+                // If the expired storage requests were not fully processed, re-insert them into storage.
```
snowmead, line 2618: can you change the comment to reflect what is there.

pallets/providers/src/utils.rs
```diff
@@ -2534,149 +2532,202 @@ mod hooks {
...
-                meter.consume(db_weight.writes(1));
-            }
+            ran_out_of_weight
         }
 
-        fn process_expired_provider_top_up_period(
+        pub(crate) fn process_expired_provider_top_up(
             typed_provider_id: StorageProviderId<T>,
-            meter: &mut WeightMeter,
+            _meter: &mut WeightMeter,
```
snowmead, line 2633: Why did you stop consuming weight for this function?

pallets/providers/src/utils.rs
TDemeco: Nope, that's the actual logic that will get uncommented in the future. I left a placeholder for now since the idea is to merge #344 then benchmark the whole `pallet-storage-providers` again in this PR and update the weights and code

pallets/providers/src/utils.rs
TDemeco: If you check at the bottom of this function you'll see that the parts where the weight is consumed are currently commented. This is for the same reason as stated in my comment above, this PR is waiting for #344 to get merged to be finalized but the logic that it introduces can be reviewed in its entirety and won't change

pallets/providers/src/utils.rs
snowmead: ah ok yes.

pallets/providers/src/utils.rs
snowmead: ok perfect

snowmead (approved):

santikaplan (approved):
Well done! 🚀

PR #346, opened 2025-01-30
fix: 🐛 avoid update of top up metadata if it already existed

TDemeco:
This PR fixes a bug that happened when a provider had previously been slashed and did not have the funds to cover its capacity deficit (and as such, it had a top up metadata element in the `AwaitingTopUpFromProviders` storage with a `started_at` and a `end_tick_grace_period` tick) and, if it's slashed again and again is found without funds, the top up metadata was overwritten with a new `started_at` and `end_tick_grace_period` ticks, which is not the desired behaviour (since the provider was awaiting the top up from before).

- [x] Define if we want the behaviour of checking whether the current tick being processed in `on_idle` matches the one found for the provider in `AwaitingTopUpFromProviders`.
- [x] If we do, define the behaviour that the logic should have when it fails to append the provider to the actual tick of expiration in `ProviderTopUpExpirations`. I left a note on how I'd do it.


pallets/providers/src/utils.rs
```diff
@@ -2888,103 +2899,151 @@ mod hooks {
                     );
                     meter.consume(db_weight.writes(1));
                 }
+            } else {
+                ran_out_of_weight = true;
```
ffarall, line 2903: Nice catch

ffarall (approved):
Great job!

PR #347, opened 2025-01-31
feat: index user insolvent event

undercover-cactus:
In this PR, we handle `SpStopStoringInsolventUser` event and update the indexer to remove the file we are stopping to store because the user is insolvent.

see https://github.com/Moonsong-Labs/internal-storage-hub-design-audit/issues/32

### DIFF
--- a/client/indexer-service/src/handler.rs
+++ b/client/indexer-service/src/handler.rs
@@ -305,7 +305,15 @@ impl IndexerService {
             pallet_file_system::Event::StorageRequestRejected { .. } => {}
             pallet_file_system::Event::BspRequestedToStopStoring { .. } => {}
             pallet_file_system::Event::PriorityChallengeForFileDeletionQueued { .. } => {}
-            pallet_file_system::Event::SpStopStoringInsolventUser { .. } => {}
+            pallet_file_system::Event::SpStopStoringInsolventUser {
+                sp_id,
+                file_key,
+                owner,
+                location,
+                new_root,
+            } => {
+                File::delete(conn, file_key).await?;
+            }
             pallet_file_system::Event::FailedToQueuePriorityChallenge { .. } => {}
             pallet_file_system::Event::FileDeletionRequest { .. } => {}
             pallet_file_system::Event::ProofSubmittedForPendingFileDeletionRequest { .. } => {}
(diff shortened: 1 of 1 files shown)


snowmead (approved):

TDemeco:
Hi! I have two comments regarding this since I'm not super familiarised with how the indexer works, so take what I say with a grain of salt:
- Regarding the event processing, shouldn't we check that it was a MSP that called the `stop_storing_for_insolvent_user` extrinsic before deleting the file from the indexer? Right now we are deleting the file from the indexer if any provider calls that extrinsic, but if it wasn't the MSP that called the extrinsic and a user calls `move_bucket`, the new MSP still needs to know the file exists (even though a BSP deleted it) because on-chain it still "exists".
- Following this, I feel like it might make sense for the file info to only be deleted from the indexer after all the BSPs + the MSP stops storing it, so instead of deleting it when the MSP calls the extrinsic, we could set a flag in the indexer to let other MSPs know that this file should not be retrieved alongside the bucket. That way, before the file is completely deleted, other providers have access to its metadata in case they need it.

links234 (commented):
We don't really handle the stop storing for insolvent user case in the MSP code (shouldn't this be sent automatically?). For this case we need to talk about the whole flow and maybe adjust/add new runtime events to keep track of.

As discussed on slack, I would keep this PR to only delete the `BspFile` entry if the provider is a BSP and update the MSP indexer side when we actually handle it there too.

client/indexer-service/src/handler.rs
```diff
@@ -305,7 +305,17 @@ impl IndexerService {
...
+            pallet_file_system::Event::SpStopStoringInsolventUser {
+                sp_id,
+                file_key,
+                owner: _,
+                location: _,
+                new_root: _,
+            } => {
+                // We are now only deleting for BSP as BSP are associating with files
+                // MSP will handle insolvent user at the level of buckets (an MSP will delete the full bucket for an insolvent user and it will produce a new kind of event)
+                BspFile::delete(conn, file_key, sp_id.to_string()).await?;
```
links234, line 317: I would leave a note for future that:
1. this does not take into account MSPs
2. after we take into account MSPs we need to manually delete the `File` too if there is no MSP/BSP pointing to it

links234 (approved):
Nice. Approved but left a small note.

PR #350, opened 2025-02-03
feat: Hash BSP ID with file key to obtain threshold

snowmead:
Replacing `fingerprint` with `file_key` for BSP threshold hash.

This change brought by discussion raised in the [audit issue](https://github.com/Moonsong-Labs/internal-storage-hub-design-audit/issues/33#issuecomment-2618859255).

### DIFF
--- a/pallets/file-system/src/utils.rs
+++ b/pallets/file-system/src/utils.rs
@@ -126,26 +126,21 @@ where
     where
         T: frame_system::Config,
     {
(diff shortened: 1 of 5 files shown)


TDemeco (approved):
Looks good, can't think of any issues that might arise from this and it helps with malicious actors predicting thresholds 💯

santikaplan (approved):
LGTM!

PR #351, opened 2025-02-03
fix: Register peer ID for BSP volunteer after file storage insert file

snowmead:
Register peer ids for BSP volunteer after inserting the file in file storage.

Resolves [#issue-39](https://github.com/Moonsong-Labs/internal-storage-hub-design-audit/issues/39)


links234 (approved):
This looks good!

Mostly as a side discussion, should we ditch requests that send corrupted data (i.e. invalid `PeerId`) ?

snowmead:
> Mostly as a side discussion, should we ditch requests that send corrupted data (i.e. invalid PeerId) ?

I don't think we should reject if there is at least a single valid `PeerId`

santikaplan (approved):
Beast! Looks good.

PR #353, opened 2025-02-04
fix: Create file deletion request upon revoke storage request without priority challenge

snowmead:
Create a file deletion request when a storage request has been revoked while an MSP has already confirmed storing the file.

Without this, there was no way to force MSPs from deleting the file from their forest and this allowed them to continue charging for the file.

Added an additional flag to the `do_delete_file` to specify whether or not to queue a priority challenge to avoid challenging the BSPs twice since we already issue a priority challenge in the `cleanup_storage_request`.


ffarall (approved):
Great job!

PR #354, opened 2025-02-04
feat: :white_check_mark: add helper to check for storage request fulfillment

TDemeco:
This PR adds a helper that can be used in integration tests to await for a storage request to be fulfilled, and uses it in the move bucket tests to avoid a racing condition where the test tries to request to move a bucket that still has a pending storage request because the BSP wasn't able to confirm storing the file in time.

### DIFF
--- a/test/util/bspNet/test-api.ts
+++ b/test/util/bspNet/test-api.ts
@@ -312,7 +312,7 @@ export class BspNetTestApi implements AsyncDisposable {
 
       /**
        * Waits for a storage request to be fulfilled on-chain.
-       * @param fileKey - File key of the storage request to wait for..
+       * @param fileKey - File key of the storage request to wait for.
        * @returns A promise that resolves when the storage request is fulfilled.
        */
       storageRequestFulfilled: (fileKey: H256 | string) =>
--- a/test/util/bspNet/test-api.ts
+++ b/test/util/bspNet/test-api.ts
@@ -311,9 +311,20 @@ export class BspNetTestApi implements AsyncDisposable {
         Waits.waitForAvailabilityToSendTx(this._api, address),
 
       /**
-       * Waits for a storage request to be fulfilled on-chain.
+       * Waits for a storage request to be removed from on-chain storage.
+       * This could happen because the storage request was fulfilled, it has expired or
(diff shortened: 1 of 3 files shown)


test/util/bspNet/test-api.ts
```diff
@@ -308,7 +308,15 @@ export class BspNetTestApi implements AsyncDisposable {
        * @returns A promise that resolves when the address has no pending extrinsics.
        */
       waitForAvailabilityToSendTx: (address: string) =>
-        Waits.waitForAvailabilityToSendTx(this._api, address)
+        Waits.waitForAvailabilityToSendTx(this._api, address),
+
+      /**
+       * Waits for a storage request to be fulfilled on-chain.
+       * @param fileKey - File key of the storage request to wait for..
```
santikaplan, line 315:
```suggestion
       * @param fileKey - File key of the storage request to wait for.
```

test/util/bspNet/waits.ts
```diff
@@ -502,3 +502,29 @@ export const waitForMspFileStorageComplete = async (api: ApiPromise, fileKey: H2
...
+  const iterations = 10;
+  const delay = 1000;
+  for (let i = 0; i < iterations + 1; i++) {
+    try {
+      await sleep(delay);
+      // Try to get the storage request from the chain
+      const result = await api.query.fileSystem.storageRequests(fileKey);
+
+      // If the storage request wasn't found, it has been fulfilled.
+      if (result.isNone) {
```
snowmead, line 516: Can we add a check for the `StorageRequestFulfilled` event instead of relying on the empty storage which can happen for a number of reasons (like errors or expires)

snowmead (changes requested):

test/util/bspNet/waits.ts
TDemeco: The issue with that approach is that we might not detect the event because the `wait` helper could get called after the storage request was already fulfilled. The idea I had with this `wait` is for it to be called to make sure the storage request was not in storage to be able to execute other operations that rely on that (such as requesting to move the bucket). I'll change the name of the wait to reflect this

snowmead (approved):

timbrinded:
LGTM!

PR #356, opened 2025-02-05
fix: Benchmark slash extrinsic

snowmead:
Notable changes:
- queueing provider top up expirations when slashed (was not done before)
- clearing queued provider top up expirations when top up is done before the end tick as been reached (before we would offload this to the `on_idle` hook to clean it up)
- benchmarking slash extrinsic with 2 paths and getting the `max` weight from both for the extrinsic -- `slash_with_awaiting_top_up` and `slash_without_awaiting_top_up`


pallets/providers/src/benchmarking.rs
```diff
@@ -1917,6 +1915,164 @@ mod benchmarks {
...
+        let (_account_id, _capacity, used_capacity) = Pallet::<T>::get_provider_details(bsp_id)?;
+
+        // Capacity needed for the provider to remain active
+        let needed_capacity = used_capacity.max(T::SpMinCapacity::get());
+
+        // Held deposit needed for required capacity
+        let required_held_amt = Pallet::<T>::compute_deposit_needed_for_capacity(needed_capacity)?;
+
+        // Needed balance to be held to increase capacity back to `needed_capacity`
+        use sp_runtime::Saturating;
```
TDemeco, line 1927: I would move this `use` statement to the top of the file

pallets/providers/src/utils.rs
```diff
@@ -1011,24 +1011,32 @@ where
...
 
+            // Remove provider top up expiration item from the queue
+            if let Some(top_up_metadata) = top_up_metadata {
+                ProviderTopUpExpirations::<T>::remove(top_up_metadata.end_block_grace_period);
+            } else {
+                log::warn!(
+                    "AwaitingTopUpFromProviders storage does not contain a top up metadata for provider {:?} while their held deposit covers the needed capacity",
+                    provider_id
+                );
+            }
```
TDemeco, line 1024: Should this be a warning? I think it's perfectly acceptable that a Provider that's getting slashed might never have been awaiting a top up because it has more than enough funds

pallets/providers/src/utils.rs
```diff
@@ -1140,7 +1148,17 @@ where
...
+
+        // Remove provider top up expiration item from the queue
+        if let Some(top_up_metadata) = top_up_metadata {
+            ProviderTopUpExpirations::<T>::remove(top_up_metadata.end_block_grace_period);
+        } else {
+            log::warn!(
+                "AwaitingTopUpFromProviders storage does not contain a top up metadata for provider {:?} while their held deposit does not cover the needed capacity",
+                provider_id
+            );
+        }
```
TDemeco, line 1161: Here I believe the warning makes more sense but, even though calling this extrinsic without having an `AwaitingTopUpFromProviders` doesn't negatively affect the system, I think this should be an error because:
- If the Provider wants to increase its maximum capacity it has the `change_capacity` extrinsic, so it doesn't make sense to call this one.
- This one emits the `TopUpFulfilled` event, which wouldn't make sense in that scenario.

pallets/providers/src/utils.rs
```diff
@@ -1,4 +1,4 @@
-use crate::*;
+use crate::{types::ShTickGetter, *};
```
TDemeco, line 1: There's a `use types::{...` statement a few lines under this one, I would add this one there instead of here.

TDemeco (changes requested):
Overall great work! Left some comments but mostly on the changes that are unrelated to the benchmarks.

pallets/providers/src/utils.rs
snowmead: So if there is some inconsistent state with slashed providers and not having them be in `AwaitedTopUpFromProviders`, they could never top up in this case. Is that what you are saying?

ffarall: I think the warning here is fine. This would be the case where a Provider has a capacity deficit, but was (for whatever reason) not included in `AwaitingTopUpFromProviders`. This would be an inconsistent state, but the result is that the Provider clears its deficit. It's not like a Provider can use this extrinsic to increase its maximum capacity as `change_capacity` does. Because this extrinsic doesn't take the new capacity as an argument, it calculates the deficit. And if there is no deficit, it just returns `Ok(())`.

ffarall (approved):

pallets/providers/src/utils.rs
ffarall: In fact, if it happened that the Provider has a deficit, but it's not in `AwaitingTopUpFromProviders`, if it's forced to use `change_capacity` it would have to wait to be able to use it. This extrinsic would allow him to always, immediately solve the deficit.

PR #358, opened 2025-02-07
Fix: properly check fingerprint

links234:
This PR fixes issue [#22](https://github.com/Moonsong-Labs/internal-storage-hub-design-audit/issues/22).


snowmead (approved):

node/src/tasks/bsp_upload_file.rs
```diff
@@ -839,6 +839,23 @@ where
...
+
+        // Verify that the fingerprint in the proof matches the expected file fingerprint
+        let expected_fingerprint = file_metadata.fingerprint;
+        if event.file_key_proof.file_metadata.fingerprint != expected_fingerprint {
+            error!(
+                target: LOG_TARGET,
+                "Fingerprint mismatch for file {:?}. Expected: {:?}, got: {:?}",
+                file_key, expected_fingerprint, event.file_key_proof.file_metadata.fingerprint
+            );
+            return Err(anyhow!("Fingerprint mismatch"));
```
ffarall, line 856: From what I see, in cases like this where we return error from `handle_remote_upload_request_event`, we don't unvolunteer from the file. Should we? Honestly, right now we're lowering the user's reputation, eventually blocking it as a peer. Which is not that bad imo.

ffarall (approved):
Great job!

PR #359, opened 2025-02-09
refactor: ♻️ move bucket requests

TDemeco:
This PR refactors move bucket requests so, instead of having two storage elements:
- `PendingMoveBucketRequests` which was a double map from MSP ID to bucket ID to move bucket request metadata, where:
  - The MSP ID was the one belonging to the new MSP that the user selected to store its bucket.
  - The move bucket request metadata had info about the user that initiated the request (owner of the bucket) and the value proposition ID to use after moving (which should correspond to a value proposition that belongs to the new MSP and is available).
- `PendingBucketsToMove` which was a map from bucket ID to the unit type `()` which was set when a bucket had a pending move bucket request open and unset otherwise.

Now we have a single storage element:
- `PendingMoveBucketRequests` which is a map from bucket ID to move bucket request metadata where the move bucket request metadata now has:
  - The user that initiated the request (owner of the bucket).
  - The MSP ID of the new MSP that the user chose to store its bucket.
  - The value proposition ID of the value proposition that the user chose for the bucket after the move. This value proposition must belong to the new MSP and be currently available.
 
This makes it so move bucket request logic is easier to follow and develop (no longer needing to keep track of two different storage elements) and more efficient storage-wise (since there's no need to store redundant information: the move bucket request existing in `PendingMoveBucketRequests` is considered equivalent to the bucket ID being set in `PendingBucketsToMove`).


snowmead (approved):

snowmead:
Nice refactor!

PR #360, opened 2025-02-09
fix: 🐛 add insolvent checks and other minor fixes

TDemeco:
This PR:
- Adds checks in each function of the `file-system` pallet that interacts with users to make sure they are not currently marked as insolvent, because if they are they should not be able to interact with StorageHub at all until they have solved their issues. This includes issuing storage requests, creating buckets, etc.
- Upgrades the `file-system`'s mock of the `UserSolvency` trait to allow for more granular control during testing, needed after the aforementioned changes.
- Makes it so when a BSP calls `bsp_confirm_storing` with a list of file keys to confirm (that have been batched), a single insolvent user can't make the transaction fail. Instead, that file key gets skipped.
- Removes a redundant check from the `bsp_confirm_stop_storing_extrinsic` since we were checking first if the BSP's used capacity after stop storing was zero to stop its randomness cycle, and we were then checking if its root was the default root to stop both its randomness and challenge cycles.
- Adds a missing remove of the file key from `PendingStopStoringRequests` when a BSP calls `do_sp_stop_storing_for_insolvent_user` for that file key, since there was a possibility that the BSP had initiated a stop storing request before the user became insolvent.
- Updates the `get_provider_id` methods of both `ReadProvidersInterface` and `ReadChallengeableProvidersInterface` to receive a reference to the account ID instead of moving it into the method. This makes it so we can remove the numerous annoying `.clone()` calls that we had when calling these methods and also makes it more efficient since there's no cloning required.
- Changes the check to see if an extrinsic was included in a manually-sealed block. Now instead of checking that the pending extrinsics of the block increased by one for each call (which could fail if BSPs or MSPs had tasks ongoing that sent extrinsics), it checks that the TX hash associated with the call is in the pending extrinsics.


pallets/file-system/src/tests.rs
```diff
@@ -2856,7 +2856,7 @@ mod revoke_storage_request {
                     ReplicationTarget::Standard
                 ));
 
-                let bsp_id = Providers::get_provider_id(bsp_account_id.clone()).unwrap();
+                let bsp_id = Providers::get_provider_id(&bsp_account_id.clone()).unwrap();
```
undercover-cactus, line 2859: If passed as a ref we can remove the `.clone()` ?

test/suites/integration/bsp/debt-collection.test.ts
```diff
@@ -672,6 +672,7 @@ describeBspNet(
           method: "chargeMultipleUsersPaymentStreams",
           module: "paymentStreams",
           checkTxPool: true,
+          exactLength: false,
```
undercover-cactus, line 675: Can you explain just for my own knowledge why we are adding this parameter ?

undercover-cactus (changes requested):
If I am not mistaken there is still some `.clone()` leftovers? Otherwise it looks good to me :+1:

pallets/file-system/src/utils.rs
```diff
@@ -720,6 +744,7 @@ where
         }
 
         // Return the collection ID associated with the bucket, if any.
+        // TODO: Check if the MSP is deleting the bucket from its local storage as well.
```
snowmead, line 747: what do you mean by this TODO?

pallets/file-system/src/tests.rs
TDemeco: Yep, good catch! I forgot to check for `.clone()` after the merge from main

snowmead (approved):
amazing job! great cleanups

undercover-cactus (approved):

pallets/file-system/src/utils.rs
TDemeco: I saw that we have no task on the MSP that reacts to the `BucketDeleted` event and I was under the impression that we had different forest keys for each bucket in the case of MSPs, so the corresponding forest key should be deleted when receiving a `BucketDeleted` event. Is that not the case?

pallets/file-system/src/utils.rs
```diff
@@ -2197,6 +2265,14 @@ where
         size: StorageData<T>,
         maybe_inclusion_forest_proof: Option<ForestProof<T>>,
     ) -> Result<(bool, ProviderIdFor<T>), DispatchError> {
+        // Check that the user that's sending the deletion request is not currently insolvent.
+        // Insolvent users can't interact with the system and should wait for all MSPs and BSPs
+        // to delete their files and buckets using the available extrinsics.
```
ffarall, line 2270:
```suggestion
        // Check that the user that's sending the deletion request is not currently insolvent.
        // Insolvent users can't interact with the system and should wait for all MSPs and BSPs
        // to delete their files and buckets using the available extrinsics.
        // Or resolve the insolvency.
```

test/util/bspNet/block.ts
```diff
@@ -140,22 +138,33 @@ export const sealBlock = async (
         hash = await call.signAndSend(signer || alice, { nonce: nonceToUse + i });
       }
 
-      for (let i = 0; i < 100; i++) {
-        const pendingTxs = (await api.rpc.author.pendingExtrinsics()).length;
-        if (pendingTxs === originalPendingTxs + 1) {
-          results.hashes.push(hash);
+      // Wait for 2 seconds until the transaction is included in the pending extrinsics
```
ffarall, line 141:
```suggestion
      // Poll for the transaction to be included in the pending extrinsics, or error out in 2 seconds
```

ffarall (approved):
Approving with comments. Great job and I trust you'll take care of them!

pallets/file-system/src/utils.rs
```diff
@@ -981,6 +1010,17 @@ where
         bucket_id: BucketIdFor<T>,
         accepted_file_keys: StorageRequestMspAcceptedFileKeys<T>,
     ) -> Result<MerkleHash<T>, DispatchError> {
+        // TODO: Since we are batching, we should maybe have a way to skip files instead of failing the whole batch.
```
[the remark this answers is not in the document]
TDemeco: Removed this TODO and added a comment about why it makes sense to error out if the user is insolvent

pallets/file-system/src/utils.rs
```diff
@@ -1382,6 +1428,19 @@ where
...
+                &storage_request_metadata.owner,
+            ) {
+                // Skip file key if the owner of the file related to the storage request is currently insolvent.
+                expect_or_err!(
+                    skipped_file_keys.try_insert(file_key),
+                    "Failed to push file key to skipped_file_keys",
+                    Error::<T>::TooManyStorageRequestResponses,
+                    result
+                );
+            }
```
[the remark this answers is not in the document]
TDemeco: Also added a comment here about why it makes sense to error out in this extrinsic if the user is insolvent

pallets/file-system/src/utils.rs
```diff
@@ -1911,24 +1976,31 @@ where
...
-        } = <PendingStopStoringRequests<T>>::get(&bsp_id, &file_key)
+        } = <PendingStopStoringRequests<T>>::take(&bsp_id, &file_key)
             .ok_or(Error::<T>::PendingStopStoringRequestNotFound)?;
 
+        // Check that the user that owns the file is not currently insolvent. The BSP should
+        // call `sp_stop_storing_for_insolvent_user` instead if the user is insolvent.
+        ensure!(
+            !<T::UserSolvency as ReadUserSolvencyInterface>::is_user_insolvent(&file_owner),
+            Error::<T>::OperationNotAllowedWithInsolventUser
+        );
```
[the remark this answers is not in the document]
TDemeco: The extrinsic would eventually fail when trying to update the payment stream so we error out early here. Added a comment with that info

PR #362, opened 2025-02-10
fix: fail when user fail to send file to MSP after several retries

undercover-cactus:
This PR is a small refactor of the user sending file task so it improves the logs. It also add 2 integration tests that test two scenarios : 
* The provider is offline and the user cannot connect to it. It will show the proper error message and exit with an error (the previous behavior showed a success log even when the file was not send).
* The provider has 2 libp2p multiaddresses register to its name. The first one is not valid and the user can connect. The second is correct and the user should successfully send the file.

### DIFF
--- a/node/src/tasks/user_sends_file.rs
+++ b/node/src/tasks/user_sends_file.rs
@@ -1,4 +1,4 @@
-use log::{debug, error, info, warn};
+use log::{debug, info, warn};
 use sc_network::{PeerId, RequestFailure};
 use shc_actors_framework::event_bus::EventHandler;
 use shc_blockchain_service::{
@@ -12,6 +12,7 @@ use shc_file_manager::traits::FileStorage;
 use shc_file_transfer_service::commands::{FileTransferServiceInterface, RequestError};
 use shp_constants::FILE_CHUNK_SIZE;
 use shp_file_metadata::ChunkId;
+use sp_core::H256;
 use sp_runtime::AccountId32;
 
 use crate::services::{handler::StorageHubHandler, types::ShNodeType};
@@ -136,7 +137,8 @@ where
             info!(target: LOG_TARGET, "No peers were found to receive file key {:?}", file_key);
         }
 
-        self.send_chunks_to_provider(peer_ids, &file_metadata).await
+        self.send_chunks_to_providers(peer_ids, &file_metadata)
+            .await
     }
 }
 
@@ -182,7 +184,8 @@ where
             info!(target: LOG_TARGET, "No peers were found to receive file key {:?}", file_key);
         }
 
-        self.send_chunks_to_provider(peer_ids, &file_metadata).await
+        self.send_chunks_to_providers(peer_ids, &file_metadata)
+            .await
     }
 }
 
@@ -190,7 +193,7 @@ impl<NT> UserSendsFileTask<NT>
 where
     NT: ShNodeType,
 {
-    async fn send_chunks_to_provider(
+    async fn send_chunks_to_providers(
         &mut self,
         peer_ids: Vec<PeerId>,
         file_metadata: &FileMetadata,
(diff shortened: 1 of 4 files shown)


node/src/tasks/user_sends_file.rs
```diff
@@ -201,211 +204,233 @@ where
...
+                .send_chunks(peer_id, file_metadata, file_key, chunk_count)
+                .await
+            {
+                Err(err) => {
+                    // if sending chunk failed with one peer id, we try with the next one.
+                    warn!(target: LOG_TARGET, "{:?}", err);
+                    continue;
+                }
+                Ok(()) => {
+                    // if successfull our job is done. No need to try with the next peer id.
```
ffarall, line 217:
```suggestion
                    // If successful our job is done. No need to try with the next peer id.
```

node/src/tasks/user_sends_file.rs
```diff
@@ -201,211 +204,233 @@ where
         // Iterates and tries to send file to peer.
         // Breaks loop after first successful attempt since all peer ids belong to the same provider.
         for peer_id in peer_ids {
-            debug!(target: LOG_TARGET, "Attempting to send chunks of file key {:?} to peer {:?}", file_key, peer_id);
+            match self
+                .send_chunks(peer_id, file_metadata, file_key, chunk_count)
+                .await
+            {
+                Err(err) => {
+                    // if sending chunk failed with one peer id, we try with the next one.
```
ffarall, line 212:
```suggestion
                    // If sending chunk failed with one peer id, we try with the next one.
```

node/src/tasks/user_sends_file.rs
```diff
@@ -201,211 +204,233 @@ where
...
+                                peer_id,
+                                retry_attempts + 1
+                            );
+                            retry_attempts += 1;
+
+                            // Wait for a short time before retrying
+                            tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+                        }
+                        Err(RequestError::RequestFailure(RequestFailure::Refused)) => {
+                            // If MSP doesn't receive file, the burden of downloading the file will be on the MSP.
```
ffarall, line 324: This comment is outdated with what we last talked.

The MSP will not have the burden of downloading the file (at least for now), we'll just retry later.

node/src/tasks/user_sends_file.rs
```diff
@@ -182,15 +184,16 @@ where
...
+            .await
     }
 }
 
 impl<NT> UserSendsFileTask<NT>
 where
     NT: ShNodeType,
 {
-    async fn send_chunks_to_provider(
+    async fn send_chunks_to_providers(
```
ffarall, line 196:
```suggestion
    async fn send_chunks_to_provider(
```

Actually it is still sending it to just ONE provider, which can have many peers.

node/src/tasks/user_sends_file.rs
```diff
@@ -201,211 +204,233 @@ where
...
+                                peer_id,
+                                retry_attempts + 1
+                            );
+                            retry_attempts += 1;
+
+                            // Wait for a short time before retrying
+                            tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+                        }
+                        Err(RequestError::RequestFailure(RequestFailure::Refused)) => {
+                            // If MSP doesn't receive file, the burden of downloading the file will be on the MSP.
```
ffarall, line 418: Same as the comment above. Outdated.

node/src/tasks/user_sends_file.rs
```diff
@@ -201,211 +204,233 @@ where
...
-                        .generate_proof(&file_key, &current_batch)
-                    {
-                        Ok(proof) => proof,
+                            return Err(anyhow::anyhow!("Failed to send file {:?}", file_key));
+                        }
                         Err(e) => {
                             return Err(anyhow::anyhow!(
-                                "Failed to generate proof for final batch of file {:?}\n Error: {:?}",
-                                file_key,
+                                "Failed to upload final batch to peer {:?} (Error: {:?})",
```
ffarall, line 423:
```suggestion
                                "Unexpected error while trying to upload final batch to peer {:?} \n    Error: {:?}",
```

I think this error message is clearer, given that it an error happened for which we didn't have a way to handle it. Also, the line break to print the `Error` helps readability.

node/src/tasks/user_sends_file.rs
```diff
@@ -201,211 +204,233 @@ where
...
+                        }
+                        Err(RequestError::RequestFailure(RequestFailure::Refused)) => {
+                            // If MSP doesn't receive file, the burden of downloading the file will be on the MSP.
+                            return Err(anyhow::anyhow!("Failed to send file {:?}", file_key));
+                        }
                         Err(e) => {
                             return Err(anyhow::anyhow!(
-                                "Failed to generate proof for batch of file {:?}\n Error: {:?}",
-                                file_key,
+                                "Failed to upload batch to peer {:?} (Error: {:?})",
```
ffarall, line 329:
```suggestion
                                "Unexpected error while trying to upload batch to peer {:?} \n    Error: {:?}",
```

I think this error message is clearer, given that it an error happened for which we didn't have a way to handle it. Also, the line break to print the `Error` helps readability.

test/suites/integration/user/send-file-to-provider.test.ts
```diff
@@ -0,0 +1,165 @@
...
+  shUser,
+  type EnrichedBspApi,
+  getContainerIp,
+  addMspContainer,
+  mspThreeKey,
+  createBucket
+} from "../../../util";
+import { CAPACITY, MAX_STORAGE_CAPACITY } from "../../../util/bspNet/consts.ts";
+
+describeMspNet("User: Send file to provider", ({ before, createUserApi, it }) => {
```
ffarall, line 12: We could have a more descriptive name for this test file and the `describe` block. Something like "User: Send file to MSP unhappy paths".

test/suites/integration/user/send-file-to-provider.test.ts
```diff
@@ -0,0 +1,165 @@
...
+
+    await userApi.assert.eventPresent("fileSystem", "NewStorageRequest");
+
+    await userApi.docker.waitForLog({
+      searchString: "Failed to send file",
+      containerName: userApi.shConsts.NODE_INFOS.user.containerName
+    });
+  });
+
+  it("MSP first multihash is wrong and second should be correct. User will be able to connect and send file.", async () => {
```
ffarall, line 66:
```suggestion
  it("MSP first libp2p multiaddress is wrong and second should be correct. User will be able to connect and send file on the second attempt.", async () => {
```

test/util/bspNet/docker.ts
```diff
@@ -140,6 +140,104 @@ export const addBspContainer = async (options?: {
...
+  assert(chainName === "storage-hub-runtime", `Error connecting to MSP via api ${containerName}`);
+
+  await api.disconnect();
+
+  console.log(
+    `▶️ MSP container started with name: ${containerName}, rpc port: ${rpcPort}, p2p port: ${p2pPort}, peerId: ${peerId}`
+  );
+
+  return { containerName, rpcPort, p2pPort, peerId };
+};
```
ffarall, line 239: This function is nearly identical to `addBspContainer`. Please make a common function that is called by either of them, with something like passing the type of container. This is to avoid duplication and forgetting to do changes in one or the other.

FYI, any LLM will do it for you.

test/util/pjsKeyring.ts
```diff
@@ -36,6 +36,8 @@ export const mspDownSeed = "//Sh-MSP-Down";
 export const mspDownKey = keyring.addFromUri(mspDownSeed, { name: "Sh-MSP-Down" });
 export const mspTwoSeed = "//Sh-MSP-Two";
 export const mspTwoKey = keyring.addFromUri(mspTwoSeed, { name: "Sh-MSP-Two" });
+export const mspThreeSeed = "//Sh-MSP-Three";
+export const mspThreeKey = keyring.addFromUri(mspThreeSeed, { name: "Sh-MSP-Three" });
```
ffarall, line 40: You didn't add this keys to the keystore. Your test is somehow still working. Remind me to talk to you about this.

ffarall (approved):
Approved to allow you to merge. I trust you will do the changes requested.

Update the description of the PR, as it is outdated.

node/src/tasks/user_sends_file.rs
undercover-cactus: Same than before. It just print the error code and doesn't have descriptive error message so I removed the break line so it not floating around in the log.

test/suites/integration/user/send-file-to-provider.test.ts
undercover-cactus: It is a general name because it generally test the use task which send file to a provider (for now only MSP). It has happy scenario and less happy one.

The first one is actually the user failing to connect to the MSP and the second it has an incorrect libp2p address registered but the second one is working.

test/util/bspNet/docker.ts
undercover-cactus: It is out of the scope of this PR. It will bring modifications in many places. Also to avoid conflicts with currently open PRs I will create another dedicated PR regarding this and the docker related functionality that have BSP in the name but can be used by any container.

PR #363, opened 2025-02-10
feat: Benchmark bucket nfts pallet

snowmead:


ffarall (approved):
Approving to speed up merge, but remember to add the newly benchmarked weights in the extrinsics in `lib.rs`.

PR #364, opened 2025-02-10
test: :white_check_mark: fix storage capacity test and improve its clarity

TDemeco:
This PR changes the `bsp_upload_file` task (more precisely, the part where the BSP tries to increase its capacity if it's not enough) to make it less likely to run into race conditions, also gives an overall makeover to the testing of that task to minimize its chances to fail when it shouldn't

### DIFF
--- a/test/suites/integration/bsp/storage-capacity.test.ts
+++ b/test/suites/integration/bsp/storage-capacity.test.ts
@@ -12,31 +12,37 @@ import {
   ShConsts
 } from "../../../util";
 
-describeBspNet("BSPNet: Validating max storage", ({ before, it, createUserApi }) => {
+describeBspNet("BSPNet: Change capacity tests.", ({ before, it, createUserApi }) => {
   let userApi: EnrichedBspApi;
 
   before(async () => {
     userApi = await createUserApi();
   });
 
-  it("Unregistered accounts fail when changing capacities", async () => {
+  it("An unregistered account (not BSP nor MSP) can't change its capacity.", async () => {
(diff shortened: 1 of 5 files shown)


test/suites/integration/bsp/storage-capacity.test.ts
```diff
@@ -184,64 +246,93 @@ describeBspNet("BSPNet: Validating max storage", ({ before, it, createUserApi })
...
+    // Assert that the BSP was accepted as a volunteer.
+    const acceptedBspVolunteerEvents = await userApi.assert.eventMany(
+      "fileSystem",
+      "AcceptedBspVolunteer"
+    );
+    assert(acceptedBspVolunteerEvents.length === 2, "BSP wasn't accepted as a volunteer twice.");
+
+    // Wait until the BSP stores both files
+    const bspAddress = userApi.createType("Address", bspKey.address);
+    await userApi.wait.bspStored(1, bspAddress);
```
undercover-cactus, line 328: If we are waiting to store both files, shouldn't we have `await userApi.wait.bspStored(2, bspAddress);` ?

undercover-cactus (approved):

test/suites/integration/bsp/storage-capacity.test.ts
TDemeco: It's two files but one extrinsic, since the BSP batches them. One could argue that maybe sometimes the BSP doesn't get to batch them so it would need two blocks to confirm storing both, but since it's not really that relevant to the test that the file gets actually stored I went for the option that can't fail

links234 (approved):
Thank you!!

ffarall (approved):

PR #366, opened 2025-02-12
feat: ✨ MSP stop storing buckets for insolvent user task

TDemeco:
This PR adds the `MspStopStoringInsolventUserTask`, which is the task in charge of deleting the currently stored buckets that belong to a user that has become insolvent. This task:
- Reacts initially to the `UserWithoutFunds` event from the runtime. When it receives it, checks if any of the MSP's stored buckets belongs to the now-insolvent user, and if it has at least one it queues up a `StopStoringForInsolventUserRequest`.
- Reacts to the `MspStopStoringBucketInsolventUser` event from the runtime. When it receives it it executes the same logic as when reacting to the `UserWithoutFunds`. This is so after successfully deleting a bucket of the insolvent user, the MSP continues deleting if it has more buckets left.
- Reacts to the `ProcessStopStoringForInsolventUserRequest` event from the blockchain service. This event is emitted when the task can acquire a forest write tx lock to process the previously emitted `StopStoringForInsolventUserRequest`.
  - It checks if the MSP is crrently storing any bucket from the insolvent user and, if it is, gets the bucket ID of one of them. To do this and avoid processing the same bucket multiple times, the process is as follows:
    - Gets the indexer database, and from it all the buckets that are marked as stored by this MSP and that belong to the insolvent user.
    - Then, from that list of buckets, gets the first bucket ID that both is currently still stored in the MSP's forest and that it's not marked as already processed.
      - This task keeps a HashSet of buckets that have been procesed (are not currently stored by the MSP according to the runtime) but have not yet been deleted, since bucket deletion waits for finality in case the `msp_stop_storing_bucket_for_insolvent_user` call gets reverted.
    - Finally, builds the `msp_stop_storing_bucket_for_insolvent_user` extrinsic and sends it.
    - Once the extrinsic is successful and gets included in a block, add this bucket to the HashSet of processed buckets that have not yet been deleted.
- Reacts to the `FinalisedMspStopStoringBucketInsolventUser` (which is emitted when a block that contained a `MspStopStoringBucketInsolventUser` event gets finalised). When it receives it, it deletes the bucket from the forest storage and all that bucket's files from the file storage.

**Pending TODOs:**
- [ ] Develop an integration test that comprehensively tests this flow.


node/src/tasks/msp_stop_storing_insolvent_user.rs
```diff
@@ -0,0 +1,455 @@
...
+/// This task will:
+/// - Stop storing the bucket for the insolvent user.
+/// - Delete the bucket from the forest storage.
+impl<NT> EventHandler<ProcessStopStoringForInsolventUserRequest>
+    for MspStopStoringInsolventUserTask<NT>
+where
+    NT: ShNodeType + 'static,
+    NT::FSH: MspForestStorageHandlerT,
+{
+    async fn handle_event(
```
ffarall, line 248: I would argue that this task doesn't need to use the Forest write lock. It's not presenting any proof, just removing the whole bucket.

Sure, if you send this transaction together with confirmation of a storage request for this bucket, either the storage request will fail if it comes in second, or it will succeed if it comes in first. Either way the result is the same: the MSP will no longer be storing the bucket.

ffarall (changes requested):

node/src/tasks/msp_stop_storing_insolvent_user.rs
```diff
@@ -0,0 +1,455 @@
...
+                .blockchain
+                .queue_stop_storing_for_insolvent_user_request(
+                    StopStoringForInsolventUserRequest::new(insolvent_user),
+                )
+                .await?;
+        }
+
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
snowmead: added runtime api in favor of indexer here https://github.com/Moonsong-Labs/storage-hub/pull/366/commits/f9ff4c76bbe94dc8f2858087a8a634380b6540e0

node/src/tasks/msp_stop_storing_insolvent_user.rs
snowmead: I would do this in another PR since the `ProcessStopStoringForInsolventUserRequestData` event is being processed by both BSP and MSP nodes.

We would have to decouple this event so we can emit this under separate circumstances (i.e. emit the event without the need for a forest root write lock).

pallets/providers/runtime-api/src/lib.rs
```diff
@@ -32,6 +32,7 @@ sp_api::decl_runtime_apis! {
         fn query_value_propositions_for_msp(who: &ProviderId) -> sp_runtime::Vec<ValuePropositionWithId>;
         fn get_bsp_stake(bsp_id: &BspId) -> Result<Balance, GetStakeError>;
         fn can_delete_provider(provider_id: &ProviderId) -> bool;
+        fn query_buckets_for_insolvent_user(msp_id: &ProviderId, user: &AccountId) -> Result<sp_runtime::Vec<BucketId>, QueryBucketsForInsolventUserError>;
```
ffarall, line 35: I'd prefer calling it `query_buckets_for_msp_and_user`. It's useful not just for insolvent users.

node/src/tasks/msp_stop_storing_insolvent_user.rs
```diff
@@ -0,0 +1,247 @@
...
+                        target: LOG_TARGET,
+                        "Failed to stop storing bucket {:?}. Continuing with next bucket. Error: {:?}",
+                        bucket_id,
+                        e
+                    );
+                    continue;
+                }
+
+                trace!(target: LOG_TARGET, "Stop storing bucket {:?} for insolvent user submitted successfully", bucket_id);
+            }
```
ffarall, line 242: Instead of doing this in a for loop (and a very slow one, since it will wait each time for the transaction to be included in the block), you can send them all at once. I recommend launching a bunch of threads to do this, and then wait for all of them to finish. You can even use the result of the thread to log the failed transactions that you left as a TODO here. Checkout the code in `msp_move_bucket` for an example of this.

node/src/tasks/msp_stop_storing_insolvent_user.rs
```diff
@@ -0,0 +1,303 @@
...
+            let stop_storing_bucket_tasks: Vec<_> = stored_buckets
+                .into_iter()
+                .map(|bucket_id| {
+                    // Clone the semaphore and task for each bucket.
+                    let semaphore = Arc::clone(&stop_storing_bucket_semaphore);
+                    let task = self.clone();
+
+                    // Spawn a task to stop storing the bucket.
+                    tokio::spawn(async move {
+                        // Try to acquire the semaphore. This is done to avoid having more concurrent tasks that the set limit.
```
ffarall, line 141:
```suggestion
                        // Try to acquire the semaphore. This is done to avoid having more concurrent tasks than the set limit.
```

ffarall (approved):

PR #370, opened 2025-02-14
fix: Optimize file transfer request memory and size configuration

snowmead:
Improves the file transfer service configuration by:

1. Making request/response packet sizes dynamic based on the batch chunk size rather than fixed 1GB limits
2. Adding a 100% overhead margin to accommodate metadata in transfer requests and proof size
3. Calculating max queued requests based on memory constraints (8GB) rather than using a fixed value of 500

resolves [issue #19](https://github.com/Moonsong-Labs/internal-storage-hub-design-audit/issues/19)

### DIFF
--- a/client/file-transfer-service/src/lib.rs
+++ b/client/file-transfer-service/src/lib.rs
@@ -8,7 +8,9 @@ use sc_network::request_responses::IncomingRequest;
 use sc_network::{config::FullNetworkConfiguration, request_responses::ProtocolConfig};
 use sc_service::Configuration;
 use shc_actors_framework::actor::{ActorHandle, ActorSpawner, TaskSpawner};
-use shc_common::types::{BlockHash, OpaqueBlock, ParachainClient};
+use shc_common::types::{
+    BlockHash, OpaqueBlock, ParachainClient, BATCH_CHUNK_FILE_TRANSFER_MAX_SIZE,
+};
 
 pub use self::handler::FileTransferService;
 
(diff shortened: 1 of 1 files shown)


links234 (approved):
Nice! LGTM!

client/file-transfer-service/src/lib.rs
```diff
@@ -21,16 +23,24 @@ pub mod handler;
...
+/// Maximum memory usage target for queued requests (8GB)
+const MAX_QUEUED_REQUESTS_MEMORY_BYTES: u64 = 8 * 1024 * 1024 * 1024;
 
-// TODO determine ideal max request/response sizes (we could technically specify here usize::MAX)
-/// Max size of response packet. (1GB)
-const MAX_RESPONSE_PACKET_SIZE_BYTES: u64 = 1 * 1024 * 1024 * 1024;
+/// Max size of request packet. Calculated based on batch chunk size plus overhead
+const MAX_REQUEST_PACKET_SIZE_BYTES: u64 = {
+    // Add 1KB for message overhead/metadata
+    let overhead = 1024;
```
links234, line 32: Are you sure about this? I think it might be more. We definitely have the (logN) overhead of the proofs and the ChunkIds themselves are some extra bytes. Keep in mind we are sending FileChunksWithProof over the messages.

snowmead: good point, will verify

links234 (approved):

PR #371, opened 2025-02-14
fix: Trigger `MoveBucketRequestedForMsp` task on finality notification

snowmead:
Moves `MoveBucketRequestedForMsp` event emission from a non-finality block import to a finality block import notification.

This is necessary due to the fact that the task that handles the move bucket request relies on the indexer to fetch the list of files that belong to the bucket being moved to download it from the BSPs holding this files. The indexer is based on finality notifications and therefore the following scenario could happen:

storage request completed [not finalized] -> move bucket [not finalized] -> fetch files from indexer (storage request not present since not finalized yet) -> inconsistent files downloaded for bucket (missing file)

We now wait for finality to ensure that there is no inconsistency between the actual state of the bucket and the files indexed since any completed storage requests would be indexed at that point of the task being processed.

resolves [issue #34](https://github.com/Moonsong-Labs/internal-storage-hub-design-audit/issues/34)


links234 (approved):
LGTM!

ffarall (approved):
Good job!

PR #376, opened 2025-02-20
refactor: change name for docker utils function in integration tests

undercover-cactus:
The docker helper functions don't differentiate between BSP and MSP. In this PR we rename it to a more general name as it can be used for any kind of container.

e.g. 
Changing name `stopBspContainer` to `stopContainer`


ffarall (approved):
LGTM

PR #377, opened 2025-02-20
fix: 🐛 Avoid retrying in same tasks

ffarall:
In this PR:
1. Add functionalities to `RetrySrategy`:
    1. Now `should_retry` takes `WatchTransactionError` as an argument, to provide more information for deciding to retry or not.
    2. There is a `retry_only_with_timout` that sets `should_retry` to return `true` only if the extrinsic timed out.
2. `bsp_submit_proof` task only retries sending the same transaction if the extrinsic timed out. It also checks if the proof is still up to date. If it fails beyond the retries (or it is outdated) it finishes the task with an error, trusting that the BlockchainService will check for proof submissions frequently and spawn new tasks to take care of it.
3. `bsp_upload_file` task only retries confirmation extrinsic if it times out.
4. `msp_delete_file` task only retries confirmation of deletion if it times out.

### DIFF
--- a/client/blockchain-service/src/commands.rs
+++ b/client/blockchain-service/src/commands.rs
@@ -29,11 +29,11 @@ use storage_hub_runtime::{AccountId, Balance, StorageDataUnit};
 
 use crate::{
     handler::BlockchainService,
-    transaction::{SubmittedTransaction, WatchTransactionError},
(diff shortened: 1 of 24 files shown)


snowmead (approved):
Nice!

PR #378, opened 2025-02-21
feat: Storage provider capacity manager

snowmead:
Refactors our method of managing storage capacity for storage providers across asynchronous tasks and grouping requests to increase capacity.

BSP and MSP tasks now can simply call the `increase_capacity` blockchain service command to request to increase their capacity by some amount needed to meet the demand. 

All capacity related logic has been unified and deduped from the tasks into a central module `capacity_manager` in the blockchain service. 

The `CapacityRequestQueue` is the global struct holding all the necessary state to manage increasing capacity efficiently. A global queue tracks all capacity requests issued by the tasks and tally a global `total_required` capacity and reject all subsequent requests if we reached the maximum capacity configured by the node operator.

On every block import notification, the capacity manager will send a change capacity extrinsic call, registering it to its global state and will check if the extrinsic made it into any subsequent block. It will send the result back (either success or failure) to all the task callbacks that were part of that batch of requests so they may gracefully fail or continue execution.

While there is a transaction being awaited for its result, the queue is reset and continues to fill up with more requests.


client/blockchain-service/src/capacity_manager.rs
```diff
@@ -0,0 +1,407 @@
...
+    /// Process any pending capacity requests.
+    ///
+    /// Since the `pending_requests` queue is kept in a valid state by pushing capacity requests that would still amount to a valid
+    /// `total_required` value not exceeding the `max_capacity_allowed`, we add them all to the `requests_waiting_for_inclusion` list
+    /// and send the `total_required` value in a single `change_capacity` extrinsic.
+    pub(crate) async fn process_capacity_requests(
+        &mut self,
+        block_number: BlockNumber,
+    ) -> Result<(), anyhow::Error> {
+        info!(target: LOG_TARGET, "[process_capacity_requests] Processing capacity requests");
```
ffarall, line 279:
```suggestion
        trace!(target: LOG_TARGET, "[process_capacity_requests] Processing capacity requests");
```

client/blockchain-service/src/capacity_manager.rs
```diff
@@ -0,0 +1,407 @@
...
+            .expect("Capacity manager should be initialized; qed");
+
+        // Skip the process if there are no pending requests.
+        if !capacity_manager_ref.has_pending_requests() {
+            info!(target: LOG_TARGET, "[process_capacity_requests] No pending requests, skipping");
+            return Ok(());
+        }
+
+        // Query earliest block to change capacity
+        info!(target: LOG_TARGET, "[process_capacity_requests] Querying earliest block to change capacity");
```
ffarall, line 303:
```suggestion
        trace!(target: LOG_TARGET, "[process_capacity_requests] Querying earliest block to change capacity");
```

client/blockchain-service/src/capacity_manager.rs
```diff
@@ -0,0 +1,407 @@
...
+            .runtime_api()
+            .query_earliest_change_capacity_block(current_block_hash, &inner_provider_id)
+            .unwrap_or_else(|_| {
+                error!(target: LOG_TARGET, "Failed to query earliest block to change capacity");
+                Err(QueryEarliestChangeCapacityBlockError::InternalError)
+            })
+            .map_err(|e| anyhow!("Failed to query earliest block to change capacity: {:?}", e))?;
+
+        if block_number < earliest_block - 1 {
+            info!(target: LOG_TARGET, "[process_capacity_requests] Earliest block to change capacity: {:?}", earliest_block);
```
ffarall, line 315:
```suggestion
            trace!(target: LOG_TARGET, "[process_capacity_requests] Earliest block to change capacity: {:?}", earliest_block);
```

client/blockchain-service/src/handler.rs
```diff
@@ -1233,6 +1253,9 @@ where
         // It is not guaranteed that the tick number will increase at every block import.
         self.notify_tick_number(&block_hash);
 
+        // Notify the capacity manager that a new block has been imported.
+        self.notify_capacity_manager(&block_number).await;
```
ffarall, line 1257: I wouldn't call it `notify_...`. Notify we use it for _notifying_ tasks that are waiting for something. This feels more like `run_capacity_manager` or something like that.

client/blockchain-service/src/utils.rs
```diff
@@ -120,6 +122,105 @@ where
         }
     }
 
+    /// Sends back the result of the submitted transaction for all capacity requests waiting for inclusion if there is one.
```
ffarall, line 125: It doesn't always do that though, right? It checks if it should.

client/blockchain-service/src/utils.rs
```diff
@@ -120,6 +122,105 @@ where
...
+            .unwrap_or_else(|_| {
+                error!(target: LOG_TARGET, "[notify_capacity_manager] Failed to query earliest block to change capacity");
+                Err(QueryEarliestChangeCapacityBlockError::InternalError)
+            })
+        else {
+            return;
+        };
+
+        // We can send the transaction 1 block before the earliest block to change capacity since it will be included in the next block.
+        if *block_number >= earliest_block - 1 {
```
ffarall, line 217: May be good to use `saturating_sub` here. I don't think it should happen, but if for whatever error `earliest_block` is 0, this will make the node crash.

node/src/tasks/bsp_upload_file.rs
```diff
@@ -498,108 +488,34 @@ where
...
-                    );
-                    anyhow::anyhow!("Failed to query storage provider capacity: {:?}", e)
-                })?;
-
-            // Wait for the earliest block where the capacity can be changed.
             self.storage_hub_handler
                 .blockchain
-                .wait_for_block(earliest_change_capacity_block)
+                .increase_capacity(CapacityRequestData::new(event.size))
                 .await?;
```
ffarall, line 505: This `.await` resolves when the transaction makes it into the block, right?

test/util/bspNet/block.ts
```diff
@@ -300,7 +306,7 @@ export const skipBlocksUntilBspCanChangeCapacity: (
         `\tSkipping to block #${blockToAdvanceTo} to go beyond MinBlocksBetweenCapacityChanges`
       );
     await advanceToBlock(api, {
-      blockNumber: blockToAdvanceTo,
+      blockNumber: blockToAdvanceTo - 1,
```
ffarall, line 309: I think it'd be clearer to modify the definition of `blockToAdvanceTo` instead of subtracting 1 here.

test/util/bspNet/consts.ts
```diff
@@ -110,4 +110,5 @@ export const TRANSFER_WEIGHT_REF_TIME = 297_297_000;
 export const TRANSFER_WEIGHT_PROOF_SIZE = 308;
 
 export const JUMP_CAPACITY_BSP = 1073741824;
+export const CAPACITY_BUFFER_PERCENTILE = 0.2;
```
ffarall, line 113: Doesn't seem to be used

ffarall (approved):
Great job in this mighty endeavor!

client/blockchain-service/src/capacity_manager.rs
```diff
@@ -0,0 +1,407 @@
...
+            error!(target: LOG_TARGET, "Failed to send capacity request result: {:?}", e);
+        }
+    }
+}
+
+/// Data needed to process a capacity request.
+pub struct CapacityRequestData {
+    /// Capacity requested to be increased.
+    required: StorageData,
+}
```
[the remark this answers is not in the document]
snowmead: The benefit is hard to see from this perspective.

The point of this was to externalize the data/struct as input to the command `increase_capacity` for tasks to call easily without needing to create a callback themselves beforehand to construct the `CapacityRequest`. 

The `CapacityRequestData` is passed from tasks and the blockchain service command handler for `IncreaseCapacity` handles creating a callback for that specific request so tasks are unaware of these details.

Now we can modify the `CapacityRequestData` or `CapacityRequest` which would necessitate a change in the tasks or the command handler independently. Which to me is a good benefit in terms of maintainability.

client/blockchain-service/src/capacity_manager.rs
```diff
@@ -0,0 +1,407 @@
...
+    capacity_config: CapacityConfig,
+    /// Pending capacity requests which have yet to be part of a transaction.
+    pending_requests: VecDeque<CapacityRequest>,
+    /// Capacity requests bundled in a single transaction waiting to be included in a block.
+    ///
+    /// All requesters will be notified via the callback when the `CapacityChanged` event is processed
+    /// in the block important notification pipeline. This list will be cleared subsequently.
+    requests_waiting_for_inclusion: Vec<CapacityRequest>,
+    /// Total required capacity.
+    total_required: StorageData,
```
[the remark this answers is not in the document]
snowmead: Yes, I pointed this out while reviewing the PR with @links234.

This is the total size required of queued requests in `pending_requests`.

client/blockchain-service/src/capacity_manager.rs
```diff
@@ -0,0 +1,407 @@
...
+
+    /// Calculate the new capacity needed based on the total required capacity
+    pub fn calculate_new_capacity(
+        &self,
+        current_capacity: StorageData,
+        total_required: StorageData,
+    ) -> StorageData {
+        // Calculate how many jumps we need to cover the required capacity
+        let jumps_needed = total_required.div_ceil(self.capacity_config.jump_capacity);
+        let total_jump_capacity = jumps_needed * self.capacity_config.jump_capacity;
```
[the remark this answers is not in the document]
snowmead: Nice catch!

client/blockchain-service/src/handler.rs
snowmead: I originally thought this, but the capacity manager also notifies tasks about their request being awaited for.

client/blockchain-service/src/utils.rs
snowmead: At the end of the phrase: 

> ... if there is one

client/blockchain-service/src/utils.rs
snowmead: Agreed.

node/src/tasks/bsp_upload_file.rs
snowmead: Yes exactly

test/util/bspNet/block.ts
snowmead: I don't think the `advanceToBlock` api needs to be aware of the context of advancing to some block. The caller should be the one to decide which block and that would keep the api simple.

test/util/bspNet/consts.ts
snowmead: Yes thanks for spotting. Saw this with @links234. He just didn't submit his review yet.

test/util/bspNet/block.ts
snowmead: I also added a mention in the docs for this function that we do this. 

```typescript
/**
 * Skips blocks until the minimum time for capacity changes is reached.
 * It will stop at the block before the minimum change time is reached since the capacity
 * change extrinsic will be sent and included in the next block.
 *
 * @param bspId - The ID of the BSP that the capacity change is for.
 * @returns A promise that resolves when the minimum change time is reached.
 */
 skipUntilBspCanChangeCapacity: (bspId?: `0x${string}` | H256 | Uint8Array) => 
    BspNetBlock.skipBlocksUntilBspCanChangeCapacity(this._api, bspId),
```

client/blockchain-service/src/capacity_manager.rs
```diff
@@ -0,0 +1,407 @@
...
+    }
+
+    /// Add all pending requests to the list of requests waiting for inclusion of the [`SubmittedTransaction`].
+    pub fn add_pending_requests_to_waiting_for_inclusion(
+        &mut self,
+        submitted_transaction: SubmittedTransaction,
+    ) {
+        self.requests_waiting_for_inclusion
+            .extend(self.pending_requests.drain(..));
+        self.last_submitted_transaction = Some(submitted_transaction);
```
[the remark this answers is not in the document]
snowmead: We handle this outside of the context of this function here.

https://github.com/Moonsong-Labs/storage-hub/blob/00300ac6ebfbe0db70e896c3dadad308d8ebe3e1/client/blockchain-service/src/utils.rs#L193-L201

client/blockchain-service/src/capacity_manager.rs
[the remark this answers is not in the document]
snowmead: I take the `min` at the end between the new capacity and max.

https://github.com/Moonsong-Labs/storage-hub/blob/00300ac6ebfbe0db70e896c3dadad308d8ebe3e1/client/blockchain-service/src/capacity_manager.rs#L103-L104

client/blockchain-service/src/capacity_manager.rs
snowmead: Made this `debug` and the rest of your log relevant suggestions. I find it to hard to debug issues in storage hub if we have to enable trace logs. There is just too much.

links234 (changes requested):
Generally looks good (functionality wise). I am not particularly happy with the way you went around burrow-checker by using `Option`s.

pallets/file-system/src/utils.rs
```diff
@@ -2813,9 +2813,9 @@ where
     }
 
     fn do_decode_generic_apply_delta_event_info(
-        encoded_event_info: &[u8],
+        mut encoded_event_info: &[u8],
```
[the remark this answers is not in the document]
snowmead: yes we do.

cannot borrow `encoded_event_info` as mutable, as it is not declared as mutable
cannot borrow as mutable

node/src/services/builder.rs
```diff
@@ -389,8 +372,7 @@ where
             <(UserRole, NoStorageLayer) as ShNodeType>::FSH::new(),
             // Not used by the user role
             ProviderConfig {
-                max_storage_capacity: 0,
-                jump_capacity: 0,
+                capacity_config: CapacityConfig::new(0, 0),
```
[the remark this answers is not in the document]
snowmead: The reason for this being that we should have different configs for different actors. We should technically have a `UserConfig` instead. So the `capacity_config` should not be optional in `ProviderConfig`. I am keeping this here to remind us.

client/blockchain-service/src/capacity_manager.rs
```diff
@@ -0,0 +1,407 @@
...
+        Self { required }
+    }
+}
+
+impl<FSH> BlockchainService<FSH>
+where
+    FSH: ForestStorageHandler + Clone + Send + Sync + 'static,
+{
+    /// Helper method to temporarily take ownership of the capacity manager
+    pub(crate) fn execute_with_capacity_manager<F, R>(&mut self, f: F) -> Result<R, anyhow::Error>
```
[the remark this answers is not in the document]
snowmead: agreed

links234 (approved):

PR #380, opened 2025-02-24
fix: 🩹 Minor integration test and comment fixes (and testing Blacksmith runners)

ffarall:
This PR:
- Solves minor race condition in integration tests.
- Changes items in doc comments to use `-`.
- Was used to test Blacksmith test runners.


.github/workflows/rust-tests.yml
```diff
@@ -9,95 +9,95 @@
...
+        with:
+          # By default actions/checkout checks out a merge commit. Check out the PR head instead.
+          # https://github.com/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit
+          ref: ${{ github.event.pull_request.head.sha }}
+      - uses: actions-rust-lang/setup-rust-toolchain@v1.8
+        with:
+          cache: false
+      - name: Run sccache-cache
+        uses: mozilla-actions/sccache-action@v0.0.4
+      - uses: rui314/setup-mold@v1
```
undercover-cactus, line 64: To avoid security issue I would avoid using github actions written by individuals or unknown organizations.

timbrinded: mozilla is pretty well known entity, as is the mold action. Industry standard some would say

snowmead (approved):

PR #381, opened 2025-02-24
chore: Refactor `FileMetadata` struct

snowmead:
Remove `pub` from `FileMetadata` fields and validate fields in `new` function to ensure fields are not "empty", like the file size not being 0.


primitives/forest-verifier/src/tests.rs
```diff
@@ -23,12 +23,12 @@ const FILES_BASE_PATH: &str = "./tmp/";
 
 #[derive(Serialize, PartialEq, Debug)]
 struct FileMetadata<'a> {
```
links234, line 25: You should use a `pub type` alias instead of duplicating.

links234 (approved):
LGTM!

ffarall (commented):
Awesome work!

ffarall (approved):

PR #382, opened 2025-02-25
feat: ✨ add RPC methods to use in maintenance mode

TDemeco:
This PR adds the following RPC methods:
- `remove_file_from_file_storage`: Allows a Provider to manually remove a file that it's currently storing in its file storage. Useful when an issue made it so a file that's no longer in its forest was incorrectly kept in its file storage.
- `add_files_to_forest_storage`: Allows a Provider to manually add files to its forest storage.
  - It requires a forest key to know in which trie of the forest the files should be added, which means BSPs and MSPs can batch add multiple files in one call as long as they belong to the same forest key.
  - Useful when an issue made it so files that are currently in the file storage and should be part of a trie were incorrectly deleted from it or not added at all.
- `remove_files_from_forest_storage`: Allows a Provider to manually remove files from a trie of its forest storage.
  - It requires a forest key to know from which trie of the forest the files should be removed, which means BSPs and MSPs can batch remove multiple files in one call as long as they belong to the same forest key.
  - Useful when an issue made it so files that the runtime thinks were deleted are still in the forest and as such there's an inconsistency between the local forest root and the on-chain recorded one.


client/rpc/src/lib.rs
```diff
@@ -121,13 +135,48 @@ pub trait StorageHubClientApi {
         bucket_id: H256,
     ) -> RpcResult<LoadFileInStorageResult>;
 
+    /// Remove a file from the file storage.
+    ///
+    /// This is useful to allow BSPs and MSPs to manually adjust their file storage to match
+    /// the state of the network if any inconsistencies are found.
+    #[method(name = "removeFileFromFileStorage")]
+    async fn remove_file_from_file_storage(&self, file_key: H256) -> RpcResult<()>;
```
snowmead, line 143: just like `add_files_to_forest_storage` we could remove this and just have one RPC to remove a list of files.

client/rpc/src/lib.rs
```diff
@@ -401,6 +462,62 @@ where
         Ok(SaveFileToDisk::Success(file_metadata))
     }
 
+    async fn add_files_to_forest_storage(
```
snowmead, line 465: I believe rocksDB already handles interior mutability preventing race conditions, but I still think we should not rely on this alone. 

At least in another PR when we add the maintenance mode, I believe that these RPC methods can only be called if the node is in maintenance mode. That would ensure that there are no async tasks writing to the forest as well while the operator calls these RPCs.

snowmead (changes requested):

client/rpc/src/lib.rs
snowmead: Let's add a TODO if you agree with this.

client/rpc/src/lib.rs
TDemeco: Yeah I started with the maintenance mode and I was thinking of exactly this, making this method receive a vector of file keys to iterate after acquiring the write lock, instead of having to iterate calling the RPC method.
I was also planning on adding another RPC method to delete all the file keys belonging to a bucket at once with the `delete_files_with_prefix` method.
I'll do both in this PR

client/rpc/src/lib.rs
TDemeco: Yes! That's the idea. This methods should be exclusive to maintenance mode where nothing else is modifying neither the file storage nor the forest storage. I still added write locks to prevent multiple RPC calls conflicting between eachother

client/rpc/src/lib.rs
snowmead: Agreed!

ffarall (approved):
Great job!

snowmead (approved):

PR #384, opened 2025-02-27
fix: limit the number of tasks being spawned

undercover-cactus:
In this PR, we introduce a semaphore to limit the number of tasks being spawned in the event bus. We have a limit of 2000 tasks per event bus. If we have reached the limit we just wait that one task finish.

Blocking the number of tasks being spawned could lead to the queue filling but we can handle that. If the queue handle p2p message we dropped the oldest and in the case of the queue handling events from the blockchain we expect the blockchain to be slower and to never fill the queue. However if for some really unlucky reason it happens we stop the node by panicking.


client/actors-framework/src/event_bus.rs
```diff
@@ -103,13 +106,18 @@ impl<T: EventBusMessage, E: EventHandler<T> + Send + 'static> EventBusListener<T
             match self.receiver.recv().await {
                 Ok(event) => {
                     let mut cloned_event_handler = self.event_handler.clone();
+                    let permit = Arc::clone(&self.semaphore)
```
links234, line 109: IMO this should be only for a subset of queues/tasks. i.e. we don't want to limit event processing and risk losing events - we want to have this happen for file transfer tasks mainly.

ffarall: You raise a good point @links234. I take it that, in part, your reasoning comes from the fact that blockchain-triggered tasks wouldn't even have this problem.

Although we could use the "criticness" of the event to choose whether to use the semaphore or not, the same argument could be used to justify using it. If blockchain triggered tasks wouldn't even reach this number of spawned tasks, there's not much of a problem in having this limit, is there?

client/actors-framework/src/constants.rs
```diff
@@ -1,3 +1,4 @@
 pub const MAX_PENDING_EVENTS: usize = 2000;
+pub const MAX_TASKS_EVENTS: usize = 2000;
```
ffarall, line 2:
```suggestion
pub const MAX_TASKS_SPAWNED_PER_QUEUE: usize = 2000;
```

ffarall (approved):

PR #385, opened 2025-02-27
feat: Unified configurations

snowmead:
Extracted all constants that should be exposed as cli parameters. This mostly were constants used by BSP and MSP tasks. They are now in the form of task specific configuration structs which are optionally specified by the user, otherwise they use their reasonable defaults we had already set to those previous constants.

The blockchain service now has its own configuration struct which only has at the moment the `extrinsic_retry_timeout`. 

Another initiative outside of the scope of this change is to separate the `StorageHubHandler` into specific actor handlers which would allow us to better handle required and non-required parameters based on the actor (`provider_type`) chosen by the user - https://github.com/Moonsong-Labs/storage-hub/pull/386. As of right now this change applies the same pattern as we do for the rest of the existing builder parameters, where everything is optional. 

I suggest we separate the builder even further into a actor specific builder structs to only expose `with_*` methods that apply to those actors specifically. This would be to prevent an MSP actor adding BSP related configurations.

There are 2 new configuration TOML examples in the root of the project - `msp_config.toml` and `bsp_config.toml` which demonstrate the structure and all possible fields you can specify as input. 

Task specific cli parameters are grouped.


client/actors-framework/src/constants.rs
```diff
@@ -1,4 +1,6 @@
+// TODO CONSTANTS
 pub const MAX_PENDING_EVENTS: usize = 2000;
 pub const MAX_TASKS_SPAWNED_PER_QUEUE: usize = 2000;
 
+// TODO CONSTANTS
```
ffarall, line 5: What's up here?

client/blockchain-service/src/capacity_manager.rs
```diff
@@ -304,13 +307,23 @@ where
             pallet_storage_providers::Call::change_capacity { new_capacity },
         );
 
+        let extrinsic_retry_timeout = Duration::from_secs(self.config.extrinsic_retry_timeout);
+
         // Send extrinsic to increase capacity
-        match self.send_extrinsic(call, Default::default()).await {
+        match self
+            .send_extrinsic(call, &SendExtrinsicOptions::new(extrinsic_retry_timeout))
```
ffarall, line 314: This doesn't seem to be useful at all if we're not calling `watch_transaction` (which we shouldn't)

client/blockchain-service/src/types.rs
```diff
@@ -239,11 +239,17 @@ pub struct SendExtrinsicOptions {
...
     /// Optionally override the nonce to use when sending the transaction.
     nonce: Option<u32>,
+    /// Maximum time to wait for a response before assuming the extrinsic submission has failed.
+    timeout: Duration,
 }
 
 impl SendExtrinsicOptions {
-    pub fn new() -> Self {
-        Self::default()
+    pub fn new(timeout: Duration) -> Self {
```
ffarall, line 247: Why is this a "required" parameter now?

client/file-transfer-service/Cargo.toml
```diff
@@ -26,7 +26,6 @@ array-bytes = { workspace = true }
 futures = { workspace = true }
 codec = { workspace = true }
 chrono = { workspace = true }
-
 # Substrate
```
ffarall, line 29: Why deleting?

node/src/service.rs
```diff
@@ -244,19 +251,50 @@ where
...
+
             // Setup specific configuration for the MSP node.
             if *provider_type == ProviderType::Msp {
                 storage_hub_builder
                     .with_notify_period(*msp_charging_period)
                     .with_indexer_db_pool(maybe_db_pool);
             }
 
+            if let Some(c) = blockchain_service {
+                storage_hub_builder.with_blockchain_service_config(c.clone());
```
ffarall, line 295: I think this would look better if all these `with_...` methods accepted an `Option<>`, so that way we can call them regardless of whether the options was passed or not. It would just look cleaner.

ffarall (commented):
Left some comments that are mostly questions. It's looking good. Not yet approving just to remind myself of talking about these things

client/actors-framework/src/constants.rs
snowmead: Thank you for spotting, marked them as constants to remove to make them cli parameters

client/blockchain-service/src/capacity_manager.rs
snowmead: I wanted to remove the duplicated parameter to avoid confusion when they both serve the exact same purpose. We mostly watch for the transaction when we `send_extrinsic` in various places.

client/blockchain-service/src/types.rs
snowmead: reasons above ^

node/src/service.rs
snowmead: good point

PR #388, opened 2025-02-28
fix: 🔒 Increase events queue size and make `NewStorageRequests` non-critical

ffarall:
This PR:
1. Increases the size of each event queue to 10,000.
2. Increases the max number of spawned tasks per event queue to 20,000.
3. Makes `NewStorageRequest` event not-critical.

1 and 2 are for making it so that if a user manages to spam with blockchain events (like `NewStorageRequests` or `FileDeletionRequest`, the limit is higher.
3 is so that even if a user manages to spam enough with `NewStorageRequests` to reach the limit, the policy will be to just ignore and drop those spamming `NewStorageRequests`.


snowmead (approved):
Nice!

PR #389, opened 2025-03-05
feat: :sparkles: add maintenance mode to node

TDemeco:
Maintenance mode allows node operators to run their nodes without them interacting with the rest of the network. This allows them to do manual changes to the file or forest storage in case of inconsistencies between what's on-chain vs the local state.

TODO:
- [x] Check if we should completely ignore block imports instead of just ignoring events in blocks.


client/blockchain-service/src/handler.rs
```diff
@@ -216,6 +218,12 @@ where
         message: Self::Message,
     ) -> impl std::future::Future<Output = ()> + Send {
         async {
+            // If the node is running in maintenance mode, we don't process any messages.
+            if self.maintenance_mode {
+                info!(target: LOG_TARGET, "🔒 Maintenance mode is enabled. Skipping message processing.");
+                return;
+            }
```
ffarall, line 225: I think we could allow commands being sent to the BlockchainService. Just in case we need that functionality while in maintenance.

client/blockchain-service/src/handler.rs
```diff
@@ -1285,6 +1295,12 @@ where
     ) where
         Block: cumulus_primitives_core::BlockT<Hash = H256>,
     {
+        // If the node is running in maintenance mode, we don't process block imports.
+        if self.maintenance_mode {
+            info!(target: LOG_TARGET, "🔒 Maintenance mode is enabled. Skipping processing ofblock import #{}: {}", block_number, block_hash);
+            return;
+        }
```
ffarall, line 1302: I would apply this boolean further up, to basically not include the block import and finality events in the queue. Cause this way it is still possible that the node will attempt some action in the initial sync.

ffarall (changes requested):

ffarall (approved):
Great job!

PR #392, opened 2025-03-11
refactor: ♻️ Add a client crate that includes the tasks

undercover-cactus:


client/Cargo.toml
```diff
@@ -0,0 +1,147 @@
+[package]
+name = "shc-client"
+version = "0.1.0"
+description = "Storage Hub Client."
```
ffarall, line 4:
```suggestion
description = "StorageHub Client."
```

client/Cargo.toml
```diff
@@ -0,0 +1,147 @@
+[package]
+name = "shc-client"
```
ffarall, line 2: `shc` stands for "storage hub client", so this is kinda confusing. I don't see any other way to name it though, so I'm not really suggesting you change it, just pointing it out.

client/Cargo.toml
ffarall, line 1: I believe you copied and pasted every dependency from `node/Cargo.toml`, which is not necessarily bad. But likely MANY of those dependencies are not needed. I would dare say none of the Cumulus dependencies are used by this crate, for instance. Please clean up to leave only the ones used by this crate.

client/src/tasks/msp_upload_file.rs
```diff
@@ -27,8 +27,8 @@ use shc_file_transfer_service::{
 };
 use shc_forest_manager::traits::{ForestStorage, ForestStorageHandler};
 
-use crate::services::types::ShNodeType;
-use crate::services::{handler::StorageHubHandler, types::MspForestStorageHandlerT};
+use crate::types::ShNodeType;
+use crate::{handler::StorageHubHandler, types::MspForestStorageHandlerT};
```
ffarall, line 31: Merge these two

ffarall (approved):
Great job! Just some minor nitpicks that I trust you'll take care of!

PR #391, opened 2025-03-11
refactor: reduce event bus provider boilerplate code

links234:
This is a pure refactor PR (no functional changes expected) to reduce boilerplate code for Storage Hub actors framework (event bus provider).

For more details please read `README.md` inside `actors-derive` crate.

PR #393, opened 2025-03-11
Refactor: actors derive subscribe helper

links234:
This builds on top of https://github.com/Moonsong-Labs/storage-hub/pull/391.

Add `subscribe_actor_event!` helper macro to help with code readability (and allow for more boilerplate reduction in future PRs).


client/actors-derive/Cargo.toml
```diff
@@ -6,7 +6,7 @@ homepage = { workspace = true }
 license = { workspace = true }
 authors = { workspace = true }
 repository = { workspace = true }
-edition = { workspace = true }
+edition = "2021"
```
undercover-cactus, line 9: Should we change this at the root instead ?

node/src/services/handler.rs
```diff
@@ -164,20 +165,25 @@ where
...
-            user_sends_file_task
-                .clone()
-                .subscribe_to(&self.task_spawner, &self.blockchain, true);
-        accepted_bsp_volunteer_event_bus_listener.start();
+        subscribe_actor_event!(
+            event: NewStorageRequest,
+            task: UserSendsFileTask,
+            service: &self.blockchain,
+            spawner: &self.task_spawner,
+            context: self.clone(),
```
undercover-cactus, line 174: Would it be possible to avoid to clone here ? Can we hold a reference of the context ?

undercover-cactus (approved):
The macros looks nice. If possible I would avoid the clone.

client/actors-derive/Cargo.toml
ffarall: I subscribe to this.

client/actors-derive/Cargo.toml
```diff
@@ -15,10 +15,11 @@ workspace = true
...
 syn = { version = "2.0.52", features = ["full", "extra-traits"] }
+quote = "1.0.35"
+proc-macro2 = "1.0.79"
 once_cell = "1.18.0"
-
-# Local dependencies
 shc-actors-framework = { workspace = true }
+
+[dev-dependencies]
+trybuild = "1.0"
```
ffarall, line 25: Why not making these dependencies workspace-dependent as all the other crates?

ffarall (commented):
Approved with comments. I trust you can take care of them!

ffarall (approved):

PR #394, opened 2025-03-13
fix: rename restricted 'size' field to 'file_size'

undercover-cactus:
This is a small fix for the types generation for the `StorageRequestMetadata` type. The field name `size` is restricted in typescript so we need to diverge from Rust and change the field name to `file_size`.


TDemeco (commented):
`current_size` makes it seem like it could change, but the file size is fixed in the metadata. I would rename it to `file_size`

links234 (approved):

PR #395, opened 2025-03-14
Refactor: actors derive subscribe mapping

links234:
This PR builds on top of https://github.com/Moonsong-Labs/storage-hub/pull/393.

Adds `subscribe_actor_event_map` which uses `subscribe_actor_event` to make it easier to subscribe to a single services.


client/actors-derive/src/lib.rs
```diff
@@ -567,6 +600,220 @@ pub fn subscribe_actor_event(input: TokenStream) -> TokenStream {
     result.into()
 }
 
+/// A macro to simplify mapping multiple events to tasks with shared parameters.
```
ffarall, line 603: Worth adding these docs to the README too.

ffarall (approved):
Looks awesome!

PR #328, opened 2025-01-16
test: MSP move bucket reject on max capacity exceeded

links234:
In order to test proper rejection of move bucket request on low capacity, a refactor and addition of onboardMsp (similar to onboardBsp) was needed.

PR #396, opened 2025-03-19
test: MSP move bucket reject on max capacity exceeded

undercover-cactus:
This PR is a replacement for https://github.com/Moonsong-Labs/storage-hub/pull/328

> In order to test proper rejection of move bucket request on low capacity, a refactor and addition of onboardMsp (similar to onboardBsp) was needed.

### DIFF
--- a/node/src/tasks/msp_charge_fees.rs
+++ b/node/src/tasks/msp_charge_fees.rs
@@ -79,9 +79,8 @@ where
                 }
             },
             None => {
-                let err_msg = "Failed to get own MSP ID.";
-                error!(target: LOG_TARGET, err_msg);
-                return Err(anyhow!(err_msg));
+                warn!(target: LOG_TARGET, "Provider not registred yet. We can't charge users.");
+                return Ok(());
             }
         };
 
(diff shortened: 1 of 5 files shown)


node/src/tasks/msp_charge_fees.rs
```diff
@@ -79,9 +79,8 @@ where
                 }
             },
             None => {
-                let err_msg = "Failed to get own MSP ID.";
-                error!(target: LOG_TARGET, err_msg);
-                return Err(anyhow!(err_msg));
+                warn!(target: LOG_TARGET, "Provider not registred yet. We can't charge users.");
+                return Ok(());
```
ffarall, line 83: Why did you change it to a warning?

ffarall (approved):
Good job! Approved with one question.

node/src/tasks/msp_charge_fees.rs
undercover-cactus: Because I consider the case when the provider is not registered as a MSP yet not to be a big issue. Otherwise I noticed at start before we have registered the MSP we have an error message and that can be confusing.

PR #400, opened 2025-03-28
Fix: remove leftover sleep

links234:
This should fix flaky tests for MSPs (i.e. capacity test).


undercover-cactus (approved):

undercover-cactus:
The test is still failing half of the time but if we don't need that sleep it is good to remove it.

ffarall (approved):

PR #401, opened 2025-04-03
feat: use stable2409 for sp-trie; make filemetada account type agnostic;

undercover-cactus:
In this PR, we pin `sp-trie` to use the `stable2409` temporarly to avoid dependencies conflict. We also make `FileMetadataInterface` AccountID agnostic to be able to use `AccountId20` or `AccountId32` without compilation errors.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -80,7 +80,7 @@ strum = { version = "0.26.3", features = ["derive"] }
 thiserror = "1.0.48"
 tokio = "1.36.0"
 toml = "0.8.19"
-trie-db = { version = "0.29.1", default-features = false }
+trie-db = { version = "0.30", default-features = false }
 trybuild = "1.0"
 proc-macro2 = "1.0.79"
 quote = "1.0.35"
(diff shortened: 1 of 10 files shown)


Cargo.toml
```diff
@@ -94,7 +94,7 @@ sp-runtime = { git = "https://github.com/paritytech/polkadot-sdk.git", branch =
 sp-runtime-interface = { git = "https://github.com/paritytech/polkadot-sdk.git", branch = "stable2409", default-features = false }
 sp-std = { git = "https://github.com/paritytech/polkadot-sdk.git", branch = "stable2409", default-features = false }
 # TODO: Change sp-trie to the next stable version when it's released with the applied fix of this [PR](https://github.com/paritytech/polkadot-sdk/pull/6486)
-sp-trie = { git = "https://github.com/paritytech/polkadot-sdk.git", branch = "master", default-features = false }
+sp-trie = { git = "https://github.com/paritytech/polkadot-sdk.git", branch = "stable2409", default-features = false }
```
ffarall: Actually this is old, the fix was already merged afaik. So changing to stable here is the right move.

ffarall (approved):
Neat!

Cargo.toml
undercover-cactus: The fix has been merged but is not present in `stable2409` unfortunately.

PR #402, opened 2025-04-09
feat: make randomness pallets agnostic from relay chain

undercover-cactus:
In this PR we remove mention of the relay chain from randomness pallet so we can use this pallet without a relay chain but some other source of block number.


pallets/randomness/src/lib.rs
```diff
@@ -56,11 +55,15 @@ pub mod pallet {
...
-        /// Interface to get the relay chain block number which was used as an anchor for the last block in the parachain.
-        type RelayBlockGetter: BlockNumberProvider<BlockNumber = RelayChainBlockNumber>;
+        /// Interface to get the relay (or other) chain block number which was used as an anchor for the last block in the parachain.
+        type BabeBlockGetter: BlockNumberProvider<BlockNumber = Self::BabeDataGetterBlockNumber>;
+        // CurrentBlockNumber
 
         /// Weight info
         type WeightInfo: crate::weights::WeightInfo;
+
+        ///
```
ffarall, line 65: Forgot the comment here?

ffarall (commented):
Good job!

ffarall (approved):

pallets/randomness/src/lib.rs
undercover-cactus: comment added!

PR #403, opened 2025-04-14
ci: remove unnecessary action for rustfmt

undercover-cactus:
In this PR we remove `actions-rust-lang/setup-rust-toolchain@v1.8` and `actions-rust-lang/rustfmt@v1` as rustfmt is already installed on ubuntu runner.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,11 +18,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions-rust-lang/setup-rust-toolchain@v1.8
-        with:
-          components: rustfmt
       - name: Rustfmt Check
-        uses: actions-rust-lang/rustfmt@v1
+        uses: cargo fmt --all -- --check
 
   check-rust-lint:
     name: "Check lint with clippy"


test/suites/integration/msp/move-bucket-low-capacity.test.ts
```diff
@@ -316,12 +309,10 @@ describeMspNet(
...
       }
 
-      // Seal 5 more blocks to pass maxthreshold and ensure completed upload requests
-      for (let i = 0; i < 5; i++) {
-        await sleep(500);
-        const block = await userApi.block.seal();
-        await userApi.rpc.engine.finalizeBlock(block.blockReceipt.blockHash);
-      }
+      // Give time for the BSPs to update the local forest root and accept storage.
+      await sleep(3000);
```
ffarall, line 313: Please use polling utility functions like `wait.bspStored`, `wait.bspVolunteer` or `waitFor()` if there is no available one.

undercover-cactus: This sleep should not be needed anymore at all. Looks like our fix on the lock removed the issue we were having here. This `sleep` was useless because the extrinsic was never sent in the first place.

PR #404, opened 2025-04-22
deps: upgrade polkadot 2412

undercover-cactus:
This PR upgrade polkadot SDK from `stable24019` to `stable2412`.

Runtime changes : 
- replace 'sp_runtime::RuntimeString' with 'alloc::string::String' (see https://github.com/paritytech/polkadot-sdk/pull/5693)

### DIFF
--- a/runtime/src/apis.rs
+++ b/runtime/src/apis.rs
@@ -285,7 +285,7 @@ impl_runtime_apis! {
 
         fn dispatch_benchmark(
             config: frame_benchmarking::BenchmarkConfig
-        ) -> Result<Vec<frame_benchmarking::BenchmarkBatch>, sp_runtime::RuntimeString> {
+        ) -> Result<Vec<frame_benchmarking::BenchmarkBatch>, alloc::string::String> {
             use frame_benchmarking::{BenchmarkError, Benchmarking, BenchmarkBatch};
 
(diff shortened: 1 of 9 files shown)


runtime/src/apis.rs
```diff
@@ -220,8 +221,8 @@ impl_runtime_apis! {
     }
 
     impl xcm_runtime_apis::dry_run::DryRunApi<Block, RuntimeCall, RuntimeEvent, OriginCaller> for Runtime {
-        fn dry_run_call(origin: OriginCaller, call: RuntimeCall) -> Result<CallDryRunEffects<RuntimeEvent>, XcmDryRunApiError> {
-            PolkadotXcm::dry_run_call::<Runtime, configs::xcm_config::XcmRouter, OriginCaller, RuntimeCall>(origin, call)
+        fn dry_run_call(origin: OriginCaller, call: RuntimeCall, result_xcms_version: Version) -> Result<CallDryRunEffects<RuntimeEvent>, XcmDryRunApiError> {
```
undercover-cactus, line 224: @TDemeco We talked about the XCM versioning and you recommended I fix it to v4. Should I enforce it here in the implementation ?
It doesn't look like we used it anywhere in the code base.

ffarall (approved):

PR #406, opened 2025-05-29
refactor: ♻️ Runtime API type abstraction in SH Client

undercover-cactus:
This PR is the 1st step in decoupling the storage hub runtime from the storage hub client to allow integration with other client.


ffarall (approved):
Very neat job!

PR #409, opened 2025-07-01
Ignore failing doctests in actors-derive and common crates

HermanObst:
These crates contain procedural macros and common utilities that don't require documentation tests, and their doctests were failing. Disabling them prevents CI failures and developers running `cargo test` to encounter failures.


TDemeco:
Maybe instead of fully disabling doctests in these files (since in the future we might want to have some) we can use the `ignore` flag on the code examples that shouldn't be treated as tests

PR #412, opened 2025-07-07
add sudo extrinsic `force_priority_challenge`. Fix submit-proofs test suite accordingly

HermanObst:
## Context
As we are removing delete_file extrinsic as part of our file deletion refactor, we wanted to add a way to run a priority challenge as a fallback mechanism - essentially as a last resort option. 

## Changes Made
The PR introduces a new sudo extrinsic force_priority_challenge that allows triggering priority challenges without relying on file deletion mechanisms. This change also includes modifications to the priority challenge test suite, specifically removing the file deletion trigger dependency.


snowmead (commented):
Left a few comments

pallets/proofs-dealer/src/lib.rs
```diff
@@ -752,6 +774,30 @@ pub mod pallet {
...
+            // Check that the extrinsic was executed by the custom origin.
+            T::PriorityChallengeOrigin::ensure_origin(origin.clone())?;
+
+            let raw_origin: RawOrigin<T::AccountId> =
+                origin.into().map_err(|_| DispatchError::BadOrigin)?;
+
+            let who = match &raw_origin {
+                RawOrigin::Signed(account) => Some(account.clone()),
+                RawOrigin::Root | RawOrigin::None => None,
+            };
```
snowmead, line 794: We can rely on the `into_signer` function to return an optional account id from an origin.

```suggestion
use frame_support::traits::OriginTrait;

// Check that the extrinsic was executed by the custom origin.
T::ChallengeOrigin::ensure_origin(origin.clone())?;

let who = origin.into_signer();
```

pallets/proofs-dealer/src/lib.rs
```diff
@@ -840,6 +887,30 @@ pub mod pallet {
...
+            // This prevents the misconfiguration where a fee is charged but the origin is root (which cannot be charged).
+            if !T::PriorityChallengesFee::get().is_zero() {
+                // Test that root origin is rejected by PriorityChallengeOrigin
+                let root_origin = frame_system::RawOrigin::Root.into();
+                assert!(
+                    T::PriorityChallengeOrigin::try_origin(root_origin).is_err(),
+                    "PriorityChallengeOrigin cannot be root when PriorityChallengesFee ({:?}) is greater than 0, as root cannot be charged fees",
+                    T::PriorityChallengesFee::get()
+                );
+            }
```
snowmead, line 913: If it's possible, we can replace these checks with if the fees > 0, then check the `origin.into_signer` returns a `Some(_account_id)`.

snowmead (changes requested):

pallets/proofs-dealer/src/lib.rs
```diff
@@ -432,7 +446,7 @@ pub mod pallet {
     pub enum Event<T: Config> {
         /// A manual challenge was submitted.
         NewChallenge {
-            who: AccountIdFor<T>,
+            who: RawOrigin<T::AccountId>,
```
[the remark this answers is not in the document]
HermanObst: Wont be better to use RawOrigin, as now (and specifically in our use case) root can be the one executing the challenge extrinsic?

HermanObst:
@snowmead all comments addressed. Only thing TBD is `RawOrigin` vs `AccountIdFor` in the events.

pallets/proofs-dealer/src/lib.rs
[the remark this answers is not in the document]
snowmead: I don't think there is any benefit to having this which incurs additional code in the extrinsic to convert it to `RawOrigin`. We already have the `into_signer` returning `Option<AccountId>`.

client/indexer-service/src/handler.rs
```diff
@@ -437,6 +437,7 @@ where
             pallet_proofs_dealer::Event::MutationsAppliedForProvider { .. } => {}
             pallet_proofs_dealer::Event::MutationsApplied { .. } => {}
             pallet_proofs_dealer::Event::NewChallenge { .. } => {}
+            pallet_proofs_dealer::Event::PriorityChallenge { .. } => {}
```
TDemeco, line 440: Just a nitpick, I would make the event `NewPriorityChallenge` to keep the format we used for `NewChallenge`

pallets/proofs-dealer/src/lib.rs
```diff
@@ -171,6 +173,12 @@ pub mod pallet {
         #[pallet::constant]
         type ChallengesFee: Get<BalanceFor<Self>>;
 
+        /// The fee charged for submitting a priority challenge.
+        /// This fee goes to the Treasury, and is used to prevent spam. Registered Providers are
```
TDemeco, line 177: Not sure if "Registered providers are exempt from this fee" applies here, I believe they shouldn't be

pallets/proofs-dealer/src/lib.rs
```diff
@@ -233,6 +241,12 @@ pub mod pallet {
         /// the execution of the `on_poll` hook bounded.
         #[pallet::constant]
         type MaxSlashableProvidersPerTick: Get<u32>;
+
+        /// Custom origin that can execute privileged operations.
```
TDemeco, line 245: "Privileged operations" doesn't sit right with me, I'd go for "Custom origin that can dispatch new priority challenges." or something along those lines

pallets/proofs-dealer/src/lib.rs
```diff
@@ -233,6 +241,12 @@ pub mod pallet {
...
+        type PriorityChallengeOrigin: EnsureOrigin<Self::RuntimeOrigin>;
+
+        /// Custom origin that can execute challenge operations.
```
TDemeco, line 248: Similar here, something like "Custom origin that can dispatch regular challenges."

pallets/proofs-dealer/src/lib.rs
```diff
@@ -432,10 +446,17 @@ pub mod pallet {
     pub enum Event<T: Config> {
         /// A manual challenge was submitted.
         NewChallenge {
-            who: AccountIdFor<T>,
+            who: RawOrigin<T::AccountId>,
             key_challenged: KeyFor<T>,
         },
 
+        /// A priority challenge was submitted.
+        PriorityChallenge {
```
TDemeco, line 454: Same as above regarding the event name^

pallets/proofs-dealer/src/lib.rs
```diff
@@ -840,6 +896,46 @@ pub mod pallet {
                 T::BlockFullnessPeriod::get(),
                 T::ChallengeTicksTolerance::get()
             );
+
+            // Check that if `ChallengesFee` is greater than 0, then `ChallengeOrigin` cannot be root or none.
+            // This prevents the misconfiguration where a fee is charged but the origin is root or none (which cannot be charged).
+            if !T::ChallengesFee::get().is_zero() {
```
TDemeco, line 902: This is a major nitpick, but although I like this, I feel like it's more natural to check first if the authorized origin is either root or none, and then check to make sure that if that's the case fees are zero.

TDemeco (approved):
Nice work 💯! Left some nitpick comments but overall looks good

pallets/proofs-dealer/src/lib.rs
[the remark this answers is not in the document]
HermanObst: e.g: having that allows us to test like this:

https://github.com/Moonsong-Labs/storage-hub/pull/412/files#diff-d961a92fa44fcaf734c2368626243fc923dcdae8450d9a250d68fa2c4db80fc2R541-R545

```
      assert(
        priorityChallengeEventDataBlob.who.isRoot,
        "The priority challenge should be initiated by Root origin"
      );
    });
```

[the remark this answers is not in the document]
snowmead: I don't really see this granularity being applied anywhere, and I don't really see the benefit either. 

Our test has to cover non root signer validation (which you already do), and if the extrinsic passes, we know that root signed it in the first place.

snowmead (approved):

PR #413, opened 2025-07-07
Remove file deletion functionality from StorageHub

snowmead:
Deleted components:
  - Remove `delete_file` and `pending_file_deletion_request_submit_proof` extrinsics
  - Remove MSP file deletion task and related client-side handling
  - Remove all file deletion storage items, events, and errors
  - Clean up associated benchmarks, tests, and type definitions
  - Remove file deletion RPC endpoints from node implementation


test/suites/integration/bsp/submit-proofs.test.ts
HermanObst, line 1: We can keep this test suite with latest changes in main

PR #419, opened 2025-07-17
feat: request_delete_file extrinsic

HermanObst:
This PR adds the basic functionality to support user to request a file deletion and tests that check happy path + edge cases.


pallets/file-system/src/utils.rs
```diff
@@ -1150,6 +1160,74 @@ where
...
+    /// 4. Computing the file key from provided metadata and verifying it matches the signed message
+    /// 5. Ensuring the operation type is Delete
+    ///
+    /// The signature verification prevents malicious file deletion requests and ensures
+    /// only the legitimate file owner can request deletion. The computed file key validation
+    /// ensures the metadata provided matches the file being deleted.
+    ///
+    /// Note: This function only validates the deletion request but does not perform the actual
+    /// file deletion. It serves as a preliminary check before the deletion process can proceed.
+    /// TODO: we probably want to hold a user deposit to prevent users to spam the network
```
snowmead, line 1178: I don't think we have to since there is no spam vector. 
Fisherman nodes and providers can ignore this request if they so chose.

pallets/file-system/src/utils.rs
```diff
@@ -1150,6 +1160,74 @@ where
...
+        ensure!(
+            <T::Providers as ReadBucketsInterface>::is_bucket_owner(&who, &bucket_id)?,
+            Error::<T>::NotBucketOwner
+        );
+
+        // Encode the message for signature verification
+        let message_encoded = signed_message.encode();
+
+        // Verify the signature (This is not strictly needed to do here, but prevents the fisherman node to do useless work)
+        let account_id_32 = Self::account_id_to_account_id32(&who)?;
```
snowmead, line 1207: All signatures require 32 byte account id?

pallets/file-system/src/utils.rs
```diff
@@ -102,6 +103,15 @@ impl<T> Pallet<T>
...
     T: pallet::Config,
 {
+    /// Helper function to convert T::AccountId to AccountId32 for signature verification
+    fn account_id_to_account_id32(
+        account_id: &T::AccountId,
+    ) -> Result<sp_core::crypto::AccountId32, Error<T>> {
+        let account_bytes = account_id.encode();
+        sp_core::crypto::AccountId32::try_from(account_bytes.as_slice())
+            .map_err(|_| Error::<T>::InvalidSignature)
+    }
```
snowmead, line 113: This would fail in Moonbeam for example since accounts are EVM based and are 20 bytes in length, I believe we should have the signature scheme and account id conversions be configurable types in the pallet config to be agnostic over any chain.

snowmead (changes requested):
Nice job! I have a few comments that we should discuss before merging.

TDemeco (changes requested):
Looks good but there are a few things (such as naming and format) that I feel could be improved so left some comments to discuss! Thanks bro 🤝

snowmead (approved):

pallets/file-system/src/utils.rs
```diff
@@ -1150,6 +1151,67 @@ where
...
+        ensure!(
+            signed_intention.file_key == computed_file_key,
+            Error::<T>::InvalidFileKeyMetadata
+        );
+
+        // Verify that the operation is Delete
+        ensure!(
+            signed_intention.operation == FileOperation::Delete,
+            Error::<T>::InvalidFileKeyMetadata
+        );
```
TDemeco, line 1210:
```suggestion
        // Check that the user that's sending the deletion request is not currently insolvent.
        // Insolvent users can't interact with the system and should wait for all MSPs and BSPs
        // to delete their files and buckets using the available extrinsics or resolve their
        // insolvency manually.
        ensure!(
            !<T::UserSolvency as ReadUserSolvencyInterface>::is_user_insolvent(&who),
            Error::<T>::OperationNotAllowedWithInsolventUser
        );

        // Check if sender is the owner of the bucket.
        ensure!(
            <T::Providers as ReadBucketsInterface>::is_bucket_owner(&who, &bucket_id)?,
            Error::<T>::NotBucketOwner
        );
        
        // Verify that the operation is Delete
        ensure!(
            signed_intention.operation == FileOperation::Delete,
            Error::<T>::InvalidFileKeyMetadata
        );

        // Encode the intention for signature verification
        let signed_intention_encoded = signed_intention.encode();

        let is_valid = signature.verify(&signed_intention_encoded[..], &who);
        ensure!(is_valid, Error::<T>::InvalidSignature);

        // Compute file key from the provided metadata
        let computed_file_key =
            Self::compute_file_key(who.clone(), bucket_id, location.clone(), size, fingerprint)
                .map_err(|_| Error::<T>::FailedToComputeFileKey)?;

        // Verify that the file_key in the signed intention matches the computed one
        ensure!(
            signed_intention.file_key == computed_file_key,
            Error::<T>::InvalidFileKeyMetadata
        );
```

TDemeco (approved):
Looks buen fla 💪🏻

PR #420, opened 2025-07-17
fix(ci): lint job

Karrq:


.github/workflows/lint.yml
```diff
@@ -18,8 +18,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - uses: actions-rust-lang/setup-rust-toolchain@v1.8
+        with:
+          components: rustfmt
```
TDemeco, line 23: FYI this was removed in PR #403 which means we shouldn't need it (but `uses` in the code below was incorrectly used instead of `run`, so that change we should keep)

TDemeco (approved):
Nice catch! We were running without the format and linter checks for a while there 😬 luckily the errors were easy to fix

.github/workflows/lint.yml
Karrq: I know it's installed but I added it back to be consistent with clippy (assuming the action loads the toolchain specified in the rust-toolchain file)

PR #421, opened 2025-07-18
fix: :green_heart: use GitHub token in all CI workflows to avoid rate limiting

TDemeco:


.github/workflows/rust-tests.yml
Karrq, line 1: fmt?

Karrq (approved):

PR #422, opened 2025-07-23
feat: Add fisherman service foundation and task infrastructure

snowmead:
feat: Add fisherman service foundation and task infrastructure

- Create fisherman-service crate with actor-based architecture
- Add ProcessFileDeletionRequest event and handler structure
- Implement fisherman_process_file_deletion task (placeholder)
- Add fisherman-service to workspace members
- Register fisherman task module

The fisherman service monitors file deletion requests and will
construct proofs of inclusion for storage providers to safely
remove files from their Merkle forests.

🤖 Generated with Claude Code
Co-Authored-By: Claude <noreply@anthropic.com>

feat: Add fisherman-service dependency to workspace and crates


snowmead:
* **#424** <a href="https://app.graphite.dev/github/pr/Moonsong-Labs/storage-hub/424?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#423** <a href="https://app.graphite.dev/github/pr/Moonsong-Labs/storage-hub/423?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#422** <a href="https://app.graphite.dev/github/pr/Moonsong-Labs/storage-hub/422?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Moonsong-Labs/storage-hub/422?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

client/src/tasks/fisherman_process_file_deletion.rs
```diff
@@ -0,0 +1,75 @@
...
+    async fn handle_event(&mut self, event: ProcessFileDeletionRequest) -> anyhow::Result<()> {
+        info!(
+            target: LOG_TARGET,
+            "Processing file deletion request for file key: {:?} and deletion target: {:?}",
+            event.file_key,
+            event.deletion_target
+        );
+
+        // TODO: Implement file deletion request handling:
+        // 1. Fetch file metadata and identify storage providers
```
ffarall, line 68: Bare in mind that the file might not exist

client/fisherman-service/src/handler.rs
```diff
@@ -0,0 +1,193 @@
...
+    ) -> impl std::future::Future<Output = ()> + Send {
+        async move {
+            match message {
+                FishermanServiceCommand::ProcessFileDeletionRequest => {
+                    info!(
+                        target: LOG_TARGET,
+                        "🎣 ProcessFileDeletionRequest received - constructing proof of inclusion"
+                    );
+                    // TODO: Emit ProcessFileDeletionRequest event for every Bucket and BSP found to hold the file key in the forest
+                }
```
ffarall, line 104: This would mean that you have to fetch information from the indexer HERE. So the `FishermanService` would need to have access to the indexer (as I see that it does). If instead you would just emit one single event, and let the task handle finding the BSPs and buckets, you wouldn't need that. Moreover, you'd avoid longer executions in this loop that is sort of critical.

client/fisherman-service/src/events.rs
```diff
@@ -0,0 +1,26 @@
...
+///
+/// Should contain all the data required to construct a proof of inclusion for a file key
+/// to be deleted from the [`FileDeletionTarget`].
+#[derive(Clone, ActorEvent)]
+#[actor(actor = "fisherman_service")]
+pub struct ProcessFileDeletionRequest {
+    /// The file key for which deletion proof is requested
+    pub file_key: FileKey,
+    /// The deletion target containing provider and bucket information
+    pub deletion_target: FileDeletionTarget,
```
ffarall, line 21: If the task ends up being the one finding the BSPs and Bucket, this wouldn't be needed.

client/fisherman-service/src/handler.rs
```diff
@@ -0,0 +1,193 @@
...
+            service: actor,
+            receiver,
+        }
+    }
+
+    async fn run(mut self) {
+        info!(target: LOG_TARGET, "🎣 Fisherman service event loop started");
+
+        // Get import notification stream (not finality stream) to monitor all blocks
+        let import_notification_stream = self.service.client.import_notification_stream();
```
ffarall, line 153: Was there a reason for using new imported blocks and not finalised blocks? Bare in mind that this is triggered also for non-best blocks and re-orgs. Also, the indexer only indexes finalised data.

client/fisherman-service/src/handler.rs
```diff
@@ -0,0 +1,193 @@
...
+
+    /// Monitor new blocks for file deletion request events
+    async fn monitor_block(
+        &mut self,
+        block_number: BlockNumber,
+        block_hash: H256,
+    ) -> Result<(), FishermanServiceError> {
+        debug!(target: LOG_TARGET, "🎣 Monitoring block #{}: {}", block_number, block_hash);
+
+        // TODO: Send FileDeletionRequest command when FileDeletionRequest event is detected
```
ffarall, line 75: Shouldn't this be sending the event? What you seem to be doing here is:
1. You get an event, so you emit a command to yourself.
2. You get the message of the command, you process it and send the event.

client/fisherman-service/src/handler.rs
ffarall: Actually, this is correct.

client/fisherman-service/src/handler.rs
```diff
@@ -0,0 +1,193 @@
...
+/// The main FishermanService actor
+///
+/// This service monitors the StorageHub blockchain for file deletion requests,
+/// constructs proofs of inclusion from Bucket/BSP forests, and submits these proofs
+/// to the StorageHub protocol to permisionlessly mutate (delete the file key) the merkle forest on chain.
+pub struct FishermanService<RuntimeApi> {
+    /// Substrate client for blockchain interaction
+    client: Arc<ParachainClient<RuntimeApi>>,
+    /// Database pool for accessing indexed data
+    _db_pool: DbPool,
```
ffarall, line 49: This is not needed if you don't lookup who are the BSPs and Bucket inside of this service, and delegate that to the task.

client/fisherman-service/src/events.rs
```diff
@@ -0,0 +1,22 @@
...
+/// Event triggered when fisherman detects a file deletion request
+///
+/// Contains the signed deletion intention data to be processed by the task.
+#[derive(Clone, ActorEvent)]
+#[actor(actor = "fisherman_service")]
+pub struct ProcessFileDeletionRequest {
+    /// The file key from the signed intention
+    pub signed_file_operation_intention: FileOperationIntention,
+    /// The signed intention
+    pub signature: MultiSignature,
```
HermanObst, line 19: This is `MultiSignature` vs `OffChainSignature` because the client do not "see" the configuration of the Runtime?

client/fisherman-service/src/handler.rs
```diff
@@ -0,0 +1,197 @@
...
+    ) -> impl std::future::Future<Output = ()> + Send {
+        async move {
+            match message {
+                FishermanServiceCommand::ProcessFileDeletionRequest { .. } => {
+                    info!(
+                        target: LOG_TARGET,
+                        "🎣 ProcessFileDeletionRequest received"
+                    );
+                    // TODO: Emit ProcessFileDeletionRequest event to trigger task
+                }
```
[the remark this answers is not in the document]
snowmead: what shouldn't be here?

client/fisherman-service/src/handler.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    /// Monitor new blocks for file deletion request events
+    async fn monitor_block(
+        &mut self,
+        block_number: BlockNumber,
+        block_hash: H256,
+    ) -> Result<(), FishermanServiceError> {
+        debug!(target: LOG_TARGET, "🎣 Monitoring block #{}: {}", block_number, block_hash);
+
+        // TODO: Emit FileDeletionRequest event
```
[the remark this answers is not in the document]
snowmead: This is the event we emit to trigger the fisherman task which runs all the logic for submitting a file deletion

client/fisherman-service/src/handler.rs
```diff
@@ -0,0 +1,197 @@
...
+    RuntimeApi::RuntimeApi: StorageEnableApiCollection + Send,
+{
+    type Message = FishermanServiceCommand;
+    type EventLoop = FishermanServiceEventLoop<RuntimeApi>;
+    type EventBusProvider = FishermanServiceEventBusProvider;
+
+    fn handle_message(
+        &mut self,
+        message: Self::Message,
+    ) -> impl std::future::Future<Output = ()> + Send {
```
HermanObst, line 97: From clippy:
async fn handle_message(
```suggestion
    fn handle_message(
        &mut self,
        message: Self::Message,
    ) -> impl std::future::Future<Output = ()> + Send {
```
    &mut self,
    message: Self::Message,
)

client/fisherman-service/src/handler.rs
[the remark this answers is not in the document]
HermanObst: Wouldn't be better to call it something like `InternalFileDeletionRequest`? Having the exact name than runtime is very confusing

client/fisherman-service/src/handler.rs
HermanObst, line 97:
```suggestion
    async fn handle_message(
        &mut self,
        message: Self::Message,
    ) {
```

client/fisherman-service/src/handler.rs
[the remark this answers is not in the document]
snowmead: The event is called `FileDeletionRequested` in the runtime. And since this is executing in the context of a client, I don't personally find this confusing. I could name it `ProcessFileDeletionRequest`

client/fisherman-service/src/handler.rs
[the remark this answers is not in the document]
ffarall: I can't think of a case like that. It would have to be a task telling the FishermanService to emit an event so that a task can be spawned. Would be a weird flow.

I'd remove it and if we ever face such a case, we'll discuss it then.

client/fisherman-service/src/lib.rs
```diff
@@ -0,0 +1,50 @@
+//! # Fisherman Service
+//!
+//! The Fisherman Service is responsible for handling file deletion requests in the StorageHub network.
+//! It monitors the blockchain for file deletion events and constructs the necessary proofs for
```
HermanObst, line 4:
```suggestion
//! It monitors the blockchain for file deletion request events and constructs the necessary proofs for
```

client/fisherman-service/src/handler.rs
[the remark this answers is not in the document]
HermanObst: Actually, this should be the event defined above: `ProcessFileDeletionRequest` ✅

client/fisherman-service/src/handler.rs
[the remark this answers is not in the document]
snowmead: I am keeping the command enum itself since I use it to implement the catch up command for the fisherman task to call

HermanObst (approved):
Nice job 💯

PR #425, opened 2025-07-23
Feat: SDK Filemanager

ftheirs:
Add to wasm package InMemoryFileDataTrie that performs the computation from fingerprint and FileMetadata that allows you to compute the fileKey
Create a wrapper class FileManager to handle files
Add tests for FileKey and Fingerprint (values taken from existing tests)
Minor fixes and updates on the project configuration


sdk/core/wasm/src/merkle.rs
```diff
@@ -0,0 +1,181 @@
...
+//  graph of Substrate-runtime crates that pull in features which cannot be
+//  compiled to `wasm32-unknown-unknown` (e.g. `std` I/O, rocksdb, proc-macro
+//  helpers, etc.).
+//
+//  For the SDK we only need the minimal in-memory Merkle-Patricia trie logic
+//  (`InMemoryFileDataTrie`) to compute Merkle roots client-side.  Therefore we
+//  copy the few small types (ChunkId, ChunkWithId, hashing adapter) instead of
+//  depending on the heavy client crate.  This keeps the WASM package tiny and
+//  avoids compilation failures in browser/Node environments.
+// ────────────────────────────────────────────────────────────────────────────
```
Karrq, line 22: I like the approach (keeping the wasm small), but ideally we would move out these items in a separate crate to depend on instead: we avoid duplication and get a semi-sdk in rust. 
Up to you if you want to do that at a later point, but I think sooner is better

sdk/core/wasm/src/merkle.rs
```diff
@@ -0,0 +1,181 @@
...
+    pub fn new() -> FileTrie {
+        FileTrie {
+            inner: InMemoryFileDataTrie::new(),
+            next_id: 0,
+        }
+    }
+
+    #[wasm_bindgen]
+    pub fn push_chunk(&mut self, bytes: &[u8]) {
+        assert!(
```
Karrq, line 98: Is it really worth a panic here and not just returning an error? I guess the original code is with this assert

sdk/core/wasm/src/merkle.rs
```diff
@@ -0,0 +1,181 @@
...
+    pub fn get_root(&self) -> Vec<u8> {
+        self.inner.get_root().to_vec()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    macro_rules! artefact_test {
```
Karrq, line 118: Typo: `artifact_test`?

sdk/package.json
```diff
@@ -7,19 +7,19 @@
     "packageManager": "pnpm@8.12.1",
     "workspaces": [
         "core",
-        "client"
+        "msp-client"
```
Karrq, line 10: This is not relevant to this PR, right? It's just a fix on the previous one if I understand correctly

sdk/core/src/index.ts
```diff
@@ -1,3 +1,2 @@
-export { add } from './wasm';
-
-// TODO: re-export crypto, filemanager, storage utilities once implemented
+export { FileTrie, FileMetadata } from '@storagehub/wasm';
```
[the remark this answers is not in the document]
Karrq: Ties with my comment below w/ types on the wasm_bindgen functions: in my view, wasm should have strict types where possible, and have a JS/TS wrapper expose idiomatic APIs that do the correct type conversions

For example, wasm shouldn't care that the bytes come from a `Buffer`, `ArrayBuffer` or whatever, it just cares that it's X bytes long. It's the caller library that should convert those types into the target for wasm to work with... At least that's what I think :)

sdk/core/wasm/src/file_metadata.rs
```diff
@@ -0,0 +1,58 @@
...
+        fingerprint: &[u8],
+    ) -> Result<FileMetadata, JsValue> {
+        // Basic validation to return early, preventing panics deeper inside.
+        if owner.len() != H_LENGTH
+            || bucket_id.len() != H_LENGTH
+            || fingerprint.len() != H_LENGTH
+            || location.is_empty()
+            || size == 0
+        {
+            return Err(JsValue::from_str("Invalid FileMetadata parameters"));
```
snowmead, line 37: I would be more specific here to return an error message based on what invalid parameters were passed. 

I imagine being a user of the SDK and not understanding what I might have done wrong.

sdk/core/wasm/src/merkle.rs
snowmead: I agree with @Karrq 

But I believe we can move perhaps the types to one of the common crates we have if it does not depend on any std deps

sdk/core/src/index.ts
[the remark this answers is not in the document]
ftheirs: Nope, probably we won't be exporting these types for general usage. Removed them from index.ts!

sdk/core/wasm/src/merkle.rs
ftheirs: This is not necessary at all. The inner functions don't have any limitations regarding the size and it will be restricted from the TS part. All the checks will be there!

sdk/core/wasm/src/file_metadata.rs
```diff
@@ -0,0 +1,58 @@
...
+
+#[wasm_bindgen]
+impl FileMetadata {
+    /// Constructs a new `FileMetadata`.
+    /// * `owner`, `bucket_id`, `fingerprint` – 32-byte arrays (passed as slices)
+    /// * `location` – arbitrary byte string (file path)
+    /// * `size` – file size in bytes
+    #[wasm_bindgen(constructor)]
+    pub fn new(
+        owner: &[u8],
```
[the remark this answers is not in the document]
ftheirs: I've removed FileTrie and FileMetadata from the exported types from SDK, I think that the better types are related to that, isn't it?

sdk/core/src/file-manager.ts
```diff
@@ -0,0 +1,85 @@
+import { statSync, existsSync } from 'node:fs';
```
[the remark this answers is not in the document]
ftheirs: 100% right! I had to make some changes to make it environment agnostic 💪

sdk/core/wasm/src/merkle.rs
snowmead: I see, I think it is reasonable to add a TODO to refactor this later than

sdk/core/wasm/src/file_metadata.rs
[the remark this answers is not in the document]
Karrq: Hmm no I meant about the arguments themselves. For example, I'd type owner as `&[u8; H_LENGTH]` directly (maybe a ref also is fine), and similar for the other  arguments. In TS, we'd cast the `Buffer` (or whatever type) to the appropriate fixed-length one after doing the required checks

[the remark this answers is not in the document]
ftheirs: In the inner class (real Rust implementation) there are no constrains regarding the size of the inputs (only a not empty is checked).
Probably this is something that was meant from the design. I mean, changing from AccountId (32 bytes) to Account20 (the one for EVM addresses) shouldn't cause any issue

Karrq (approved):

sdk/core/wasm/Cargo.toml
```diff
@@ -10,4 +10,12 @@ repository = "https://github.com/moonsong-labs/storage-hub"
...
\ No newline at end of file
+wasm-bindgen = "0.2"
+hex = { workspace = true }
+parity-scale-codec = "3.7.4"
+sp-trie = { workspace = true, default-features = false }
+sp-core = { workspace = true, default-features = false }
+sp-runtime-interface = { workspace = true, default-features = false, features = ["disable_target_static_assertions"] }
+
+shp-file-metadata = { workspace = true, default-features = false }
+shp-constants = { workspace = true, default-features = false }
```
TDemeco, line 21: Let's keep our dependencies alphabetically ordered

sdk/core/wasm/src/merkle.rs
```diff
@@ -0,0 +1,180 @@
+use wasm_bindgen::prelude::*;
+
+use parity_scale_codec::{Compact, Decode, Encode};
```
TDemeco, line 3: Remove unused imports

sdk/core/wasm/src/merkle.rs
```diff
@@ -0,0 +1,180 @@
...
+use shp_constants::FILE_CHUNK_SIZE;
+use shp_file_metadata::{Chunk, ChunkId, ChunkWithId};
+use sp_core::{hashing::blake2_256, Hasher};
+use sp_trie::{LayoutV1, MemoryDB, TrieDBMutBuilder, TrieMut};
+use std::collections::hash_map::DefaultHasher;
+
+// ────────────────────────────────────────────────────────────────────────────
+//  NOTE:
+//  The full StorageHub client crate (`client/file-manager`) depends on a large
```
TDemeco, line 12:
```suggestion
//  The full StorageHub `file-manager` crate (`client/file-manager`) depends on a large
```

TDemeco (changes requested):
Looks good! Requested some minor changes

sdk/core/__tests__/file-manager.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { describe, it, expect } from 'vitest';
+import { join } from 'node:path';
+import { statSync, createReadStream } from 'node:fs';
+import { Readable } from 'node:stream';
+// @ts-ignore – wasm types emitted after build
```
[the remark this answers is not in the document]
ftheirs: I just changed it. The FileManager type is generated once the wasm package is compiled. Until then, the type doesn't exist and the linter complain about it. However, I tweak that a little bit and imported the type that is expose from index and removed the `ts-ignore`

sdk/core/src/file-manager.ts
```diff
@@ -0,0 +1,92 @@
...
+    const registry = new TypeRegistry();
+    const trie = new FileTrie();
+
+    const stream = this.file.stream();
+    const reader = stream.getReader();
+    let buffer = new Uint8Array();
+    let bufferOffset = 0;
+
+    try {
+      while (true) {
```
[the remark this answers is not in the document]
ftheirs: Yeeeah, me too but this is Typescript  💔

sdk/core/src/file-manager.ts
```diff
@@ -0,0 +1,92 @@
...
+        // The default chunk size for a Blob stream is typically 64 KiB
```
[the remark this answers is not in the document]
ftheirs: I replaced this with a `NOTE`. You can configure the chunk size from stream but it's typically 64kB. Just a helper to understand that you're not handling the whole file in memory (if it's a big one)

sdk/core/src/file-manager.ts
```diff
@@ -0,0 +1,92 @@
...
+          buffer = newBuffer;
+          bufferOffset = 0;
+
+          while (buffer.length - bufferOffset >= CHUNK_SIZE) {
+            const chunk = buffer.subarray(bufferOffset, bufferOffset + CHUNK_SIZE);
+            trie.push_chunk(chunk);
+            bufferOffset += CHUNK_SIZE;
+          }
+        }
+      }
```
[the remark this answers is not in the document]
ftheirs: Yes sir 🫡

PR #426, opened 2025-07-24
feat: add delete file extrinsic

HermanObst:
Implements a new delete_file extrinsic that allows any entity (fishermen, MSPs, BSPs) to execute file deletions on behalf of users using signed intentions and forest proof verification.


snowmead (approved):
Beast!

PR #427, opened 2025-07-28
test(rpc): :white_check_mark: remote files integration

Karrq:
This PR adds a simple set of integration tests for remote URLs feature added to the RPC endpoints for save/load files


ftheirs (approved):
It looks good 🚀

TDemeco (approved):
Looks good! Tests could be improved a bit to avoid sleeps and to be a bit more organized, but they serve their purpose so 👍🏻 from me

client/rpc/src/remote_file/ftp.rs
```diff
@@ -57,14 +57,17 @@ impl AsyncRead for FtpStreamReader {
                         this.current_chunk = None;
                         this.position = 0;
                     }
+
+                    // We've read some bytes, return Ready immediately
+                    return Poll::Ready(Ok(()));
```
TDemeco, line 62: This is the main change you did, right? I think I'm missing some context, is there a reason to stop reading more from the next chunk if there's available space in the buffer still?

TDemeco (approved):
Nice updates to the tests 💯

PR #436, opened 2025-08-01
docs: 📝  update import ordering rules

Karrq:
It's from what I gathered and from personal preference, decided to add the the CLAUDE.md since it's something we always want to respect, and we could easily add a lint rule/job to enforce this :) 

Feel free to discard or change it as you see fit


Karrq:
Closing as rustfmt doesn't allow this pattern natively so we'd just end up overriding it all the time

PR #438, opened 2025-08-01
feat(backend): ✨ mock endpoints

Karrq:
This PR adds a mock version of all the designed endpoints for the backend - the implementation will be filled in later, but all the endpoints should verify that the requests data has a valid shape and also respond with similarly valid shaped data

### DIFF
--- a/backend/lib/src/api/validation.rs
+++ b/backend/lib/src/api/validation.rs
@@ -1,6 +1,6 @@
 use axum_extra::headers::{authorization::Bearer, Authorization};
 use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine};
-use jsonwebtoken::{decode, decode_header, Algorithm, DecodingKey, Validation};
+use jsonwebtoken::{decode, Algorithm, DecodingKey, Validation};
 use rand::Rng;
 use serde_json::Value;
 


backend/lib/src/api/handlers.rs
```diff
@@ -1,28 +1,290 @@
...
+    State(services): State<Services>,
+    headers: HeaderMap,
+    Path((bucket_id, file_key)): Path<(String, String)>,
+) -> Result<impl IntoResponse, Error> {
+    let token = extract_token(&headers)?;
+    validate_token(&token)?;
+
+    let response = services.msp.distribute_file(&bucket_id, &file_key).await?;
+    Ok(Json(response))
+}
```
TDemeco, line 261: We can leave this be for now but we might remove it later since we were thinking of doing the distribution directly in the MSP client instead of in the backend. Just to let you know

backend/lib/src/api/handlers.rs
```diff
@@ -1,28 +1,290 @@
...
     }
+
+    // Mock implementation - return success response
+    let response = FileUploadResponse {
+        status: "upload_successful".to_string(),
+        file_key: file_key.clone(),
+        bucket_id: bucket_id.clone(),
+        fingerprint: "5d7a3700e1f7d973c064539f1b18c988dace6b4f1a57650165e9b58305db090f".to_string(),
+        location: format!("/files/{}/{}", bucket_id, file_name),
+    };
```
TDemeco, line 246: I'm not entirely sure this `file_name` thing is correct, mostly because the file key depends on the location, and the location can be anything since it lives only inside its bucket. We would have to verify that the file name is the same as the last string after "/" in the actual location. If this is only for the mock then that's ok

backend/lib/src/api/handlers.rs
```diff
@@ -1,28 +1,290 @@
...
+        .split('.')
+        .nth(1)
+        .and_then(|payload| {
+            use base64::{engine::general_purpose, Engine};
+            general_purpose::STANDARD.decode(payload).ok()
+        })
+        .and_then(|bytes| String::from_utf8(bytes).ok())
+        .and_then(|s| serde_json::from_str::<serde_json::Value>(&s).ok())
+        .and_then(|v| v.get("address").and_then(|a| a.as_str()).map(String::from))
+        .unwrap_or_else(|| "0xf24FF3a9CF04c71Dbc94D0b566f7A27B94566cac".to_string());
```
TDemeco, line 286: I see that this is used twice, I'd extract it into a descriptive function to make the code cleaner

backend/lib/src/api/routes.rs
```diff
@@ -1,14 +1,61 @@
...
 /// Creates the router with all API routes
 pub fn routes(services: Services) -> Router {
+    // we use a separate router for the upload path
+    // so we can disable the request body limit
+    let file_upload = Router::new()
+        .route(
+            "/buckets/:bucket_id/upload/:file_key",
+            put(handlers::upload_file),
+        )
+        .route_layer(DefaultBodyLimit::disable());
```
TDemeco, line 20: What would be the default body limit otherwise?

backend/lib/src/api/routes.rs
```diff
@@ -1,14 +1,61 @@
...
+        .route(
+            "/buckets/:bucket_id/download/:file_key",
+            get(handlers::download_by_key),
+        )
+        .route(
+            "/buckets/:bucket_id/download/path/*file_location",
+            get(handlers::download_by_location),
+        )
+        // Payment route
+        .route("/payment_stream", get(handlers::payment_stream))
```
TDemeco, line 57:
```suggestion
        // Payment streams routes
        .route("/payment_stream", get(handlers::payment_stream))
```

backend/lib/src/api/validation.rs
TDemeco, line 1: Let's unify all imports at the top of the file and, as a general comment that applies to all the PR, let's label every function and mock-related data with a comment or something so we can easily identify and remove them/update them accordingly as we start implementing the actual functionality

backend/lib/src/models/msp_info.rs
```diff
@@ -0,0 +1,61 @@
...
+#[derive(Debug, Serialize)]
+pub struct StatsResponse {
+    pub capacity: Capacity,
+    #[serde(rename = "activeUsers")]
+    pub active_users: u64,
+    #[serde(rename = "lastCapacityChange")]
+    pub last_capacity_change: u64,
+    #[serde(rename = "valuePropsAmount")]
+    pub value_props_amount: u64,
+    #[serde(rename = "BucketsAmount")]
```
TDemeco, line 30:
```suggestion
    #[serde(rename = "bucketsAmount")]
```

TDemeco (changes requested):
Nice work! 💯  Left some minor comments and a general change in regards to clearly separating what's mocked and what should we should keep for the actual implementation, to make it easier for us when we start implementing

backend/lib/src/api/handlers.rs
```diff
@@ -1,28 +1,290 @@
...
```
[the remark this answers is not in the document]
Karrq: Yep - I can probably put it in a dedicated function...
The "mock" part is the `.unwrap_or_else` but the rest of the logic is pretty much the same. Maybe axum already has a way to decode the base64 header and get it in the handler directly

TDemeco (approved):
Good mocks 🚀

PR #440, opened 2025-08-05
fix: 🐛 correctly implement config parsing + integration test fixes

TDemeco:
This PR:
- Adds the `RpcConfig` struct which will hold all configurations related to RPCs (for now, only the configuration for remote files in `remote_file`).
  - This gets populated from the CLI or, if possible, from the configuration file for BSPs (`bsp_config.toml`) or MSPs (`msp_config.toml`) depending on which type of provider this is.
  - Default values have been provided that should work for most providers.
- Changes the `chunk` from the RPC remote file methods to have nothing to do with `FILE_CHUNK_SIZE`, since the former is related to the buffers used when fetching and uploading files to FTP or HTTP servers while the later is related to the size of the leafs of our file tries.
- Changes all docker containers from our integration tests to spawn under the `storage-hub` compose stack instead of having some spawining under the `docker` compose stack and others with no stack at all.
- Adds to the keystore the keys for a third MSP since it's used in more than one test (and we already had the constants for it).
- Fixes the `msp-move-bucket-low-capacity` test to ensure that no storage request is left open before issuing the move bucket request, preventing this test to error out if BSPs don't confirm storing all files quick enough.
- Fixes the `send-file-to-provider` test which wasn't asserting for the correct logs and as such was giving us false positives.

### DIFF
--- a/bsp_config.toml
+++ b/bsp_config.toml
@@ -46,9 +46,9 @@ max_submission_attempts = 5
 extrinsic_retry_timeout = 10
 
 # RPC configuration
-[provider.rpc]
+[provider.rpc_config]
 # Remote file configuration
-[provider.rpc.remote_file]
+[provider.rpc_config.remote_file]
 # Maximum file size in bytes (default: 10GB)
 max_file_size = 10737418240
 # Connection timeout in seconds (default: 30)
@@ -61,13 +61,13 @@ follow_redirects = true
 max_redirects = 10
 # User agent string (default: "StorageHub-Client/1.0")
 user_agent = "StorageHub-Client/1.0"
-# Chunk size in bytes (default: 8192)
+# Chunk size in bytes. This is different from the FILE_CHUNK_SIZE constant in the runtime, as it only affects file upload/download. (default: 8KB)
 chunk_size = 8192
-# Number of FILE_CHUNK_SIZE chunks to buffer (default: 512)
+# Number of chunks to buffer (default: 512)
 chunks_buffer = 512
 
 # Optional indexer configuration
 [indexer]
 # Set to true to enable the indexer
 indexer = true
-database_url = "postgresql://postgres:postgres@docker-sh-postgres-1:5432/storage_hub"
+database_url = "postgresql://postgres:postgres@storage-hub-sh-postgres-1:5432/storage_hub"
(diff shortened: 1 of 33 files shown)


client/rpc/src/remote_file/http.rs
```diff
@@ -205,8 +205,8 @@ impl RemoteFileHandler for HttpFileHandler {
                 let stream = response.bytes_stream();
                 let reader = StreamReader::new(stream.map_err(std::io::Error::other));
 
-                // Wrap the reader in a buffered reader with buffer size based on chunk_size * chunks_buffer
-                let buffer_size = self.config.chunk_size * self.config.chunks_buffer.max(1);
+                // Wrap the reader in a buffered reader with buffer size based on chunks_buffer
+                let buffer_size = self.config.chunks_buffer.max(1) * self.config.chunk_size;
```
snowmead, line 209: I am curious why this change?

client/rpc/src/remote_file/mod.rs
```diff
@@ -88,8 +88,8 @@ impl RemoteFileConfig {
             follow_redirects: true,
             max_redirects: 10,
             user_agent: "StorageHub-Client/1.0".to_string(),
-            chunk_size: shc_common::types::FILE_CHUNK_SIZE as usize, // 1KB (FILE_CHUNK_SIZE)
-            chunks_buffer: 512, // 512 chunks * 1KB = 512KB buffer
+            chunk_size: 8192,   // 8KB default
+            chunks_buffer: 512, // 512 FILE_CHUNK_SIZE chunks default (512KB)
```
snowmead, line 92: why do we mention here `FILE_CHUNK_SIZE`? I thought the point was to remove the relation

snowmead (approved):

bsp_config.toml
```diff
@@ -61,13 +61,13 @@ follow_redirects = true
 max_redirects = 10
 # User agent string (default: "StorageHub-Client/1.0")
 user_agent = "StorageHub-Client/1.0"
-# Chunk size in bytes (default: 1024 - FILE_CHUNK_SIZE)
-chunk_size = 1024
-# Number of FILE_CHUNK_SIZE chunks to buffer (default: 512)
+# Chunk size in bytes. This is different from the FILE_CHUNK_SIZE constant in the runtime, as it only affects file upload/download. (default: 8KB)
+chunk_size = 8192
+# Number of chunks to buffer (default: 512)
```
Karrq, line 66:
```suggestion
# Number of `chunk_size` chunks to buffer during upload/download. (default: 512)
```

I'd clarify it again here (and in the other explanations) which chunk this refers to

PR #448, opened 2025-08-20
feat: ✨ add EVM precompiles

TDemeco:
This PR adds the required EVM precompiles to StorageHub, allowing its user-facing extrinsics to be called through a Solidity smart contract that lives on-chain. To do so it:
- Adds the `/precompiles` folder and the `pallet-evm-precompile-file-system` package inside it, which holds the Substrate logic to expose the required extrinsics through a EVM interface. Such interface is found in the associated `FileSystem.sol` contract.
  - This package contains the core logic of the precompiles, plus a mocked EVM-compatible runtime, comprehensive unit testing for all the exposed extrinsics and some integration testing using bare runtime calls to better simulate what would be going on in the Substrate side.
  - The exposed extrinsics are:
    - `create_bucket`
    - `request_move_bucket`
    - `update_bucket_privacy`
    - `create_and_associate_collection_with_bucket`
    - `delete_bucket`
    - `issue_storage_request`
    - `revoke_storage_request`
    - `request_delete_file`
  - Two helper functions are also added:
    - `get_pending_file_deletions_request_count`, which allows the caller to get the pending file deletions that a specific user has.
    - `derive_bucket_id`, which allows the caller to derive a bucket's ID by passing its metadata, using the same logic that the runtime does.
  -  Solidity events were added to be emitted when calling extrinsics through the EVM, parallel to the Substrate ones:
     - `BucketCreated` when a new bucket has been created.
     - `BucketMoveRequested` when a move of a bucket has been requested.
     - `BucketPrivacyUpdated` when the privacy of a bucket has been updated.
     - `CollectionCreated` when a new NFT collection has been created for a private bucket.
     - `BucketDeleted` when a bucket has been deleted.
     - `StorageRequestIssued` when a new storage request has been issued.
     - `StorageRequestRevoked` when an existing storage request has been revoked and deleted before being fulfilled.
     - `FileDeletionRequested` when a user has requested a file to be deleted.

Extras:
- A minor bug in the integrity tests of `pallet-proofs-dealer` has been fixed.
- `jsonrpsee` has been updated to a newer version to avoid an incompatibility issue.
- `scale-info` and `sp-core` std features have been added to the runtime API of the `file-system` pallet since otherwise it wasn't compiling correctly in an std environment.

### TODO
- [ ] ~Add the recently added `delete_file` and `delete_file_for_incomplete_storage_request` extrinsics~. This are to be executed by fishermen or providers, so no need to add them to the precompile.
- [ ] ~Add user-facing operations from pallets other than `file-system`~. Will be done in a subsequent PR
- [x] Once the EVM-compatible StorageHub runtime is ready, add the precompiles to it
- [ ]  ~Create an integration test that actually executes the EVM calls.~ Future PR, when we have a testing framework that works with the new EVM-compatible runtime.


precompiles/pallet-file-system/src/lib.rs
```diff
@@ -0,0 +1,579 @@
...
+            name,
+            private,
+            value_prop_id,
+        };
+
+        // TODO: Consult about what storage growth argument is
+        RuntimeHelper::<Runtime>::try_dispatch(handle, Some(origin).into(), call, 0)?;
+
+        // Emit EVM log for BucketCreated event
+        // Event signature: BucketCreated(address indexed who, bytes32 indexed bucketId, bytes32 indexed mspId)
```
snowmead, line 217: Is this some standard comment you see in other precompiles, because I don't see the point in it.

TDemeco:
> Was there a reason for not exposing the Provider pallet extrinsics?

Providers are expected to run StorageHub nodes and as such should interact directly through Substrate extrinsics. These precompiles have been designed only for user-facing operations. There are a few missing such as `pay_outstanding_debt` and `clear_insolvent_flag` from the payment streams pallet which will be added later but the ones present here are the core ones that needed to be integrated into the SDK ASAP

precompiles/pallet-file-system/src/mock.rs
```diff
@@ -0,0 +1,758 @@
...
+        .concat();
+        let hashed_subject = blake2_256(&subject_concat_block);
+        (
+            H256::from_slice(&hashed_subject),
+            frame_system::Pallet::<Test>::block_number().saturating_sub(3),
+        )
+    }
+}
+
+/****** Commit-Reveal Randomness pallet ******/
```
snowmead, line 250: nit: Why is this commented like this and why do we have it and not on other pallets?

precompiles/pallet-file-system/src/mock.rs
snowmead, line 1: Is it possible to use the mock module from the file system pallet crate? 
I think there is a lot repeated here

precompiles/pallet-file-system/src/mock.rs
TDemeco: I got most of the mock structure from the actual mock of the `file-system` pallet, including this.
IIRC, the StorageHub runtime file was getting quite messy and when I added the CR randomness pallet to it I decided to clean its structure up using this type of "isolation" for each pallet. I did not do it for the mocked runtimes of each pallet since that would have cluttered that PR a lot

TDemeco: We should definitely map a task to clean up the mocked runtimes we have scattered across the codebase

precompiles/pallet-file-system/src/lib.rs
```diff
@@ -0,0 +1,580 @@
...
+    )
+}
+
+/// FileSystem precompile.
+///
+/// Provides EVM-compatible interface to the file system pallet functionality.
+#[derive(Debug, Clone)]
+pub struct FileSystemPrecompile<Runtime>(PhantomData<Runtime>);
+
+// TODO: Change all concrete types (AccountId20, EthereumSignature, etc.) to `storage_hub_runtime::` types once the EVM-compatible SH runtime is ready
```
ffarall, line 178: I think it is fair to assume that a runtime using a precompile will use `AccountId20` and `EthereumSignature`. Pluse if you add dependencies from the new EVM compatible runtime, it will cause a circular dependency, because that runtime will use this precompile likely as well.

ffarall (approved):

snowmead (approved):
Nice job!

PR #464, opened 2025-09-03
feat: handle fisherman incomplete storage request

snowmead:
Notable changes:

- monitor block now checks for `IncompleteStorageRequest` event 
- add revoking and expired storage request integration tests 
- do not return err result if peer id and file key already registered

### DIFF
--- a/client/file-transfer-service/src/handler.rs
+++ b/client/file-transfer-service/src/handler.rs
@@ -37,7 +37,7 @@ use sc_network::{
     IfDisconnected, NetworkPeers, NetworkRequest, ProtocolName, ReputationChange,
 };
 use sc_network_types::PeerId;
-use sc_tracing::tracing::{debug, error, info, warn};
+use sc_tracing::tracing::{debug, error, info, trace, warn};
 
 use shc_actors_framework::actor::{Actor, ActorEventLoop};
 use shc_common::{
@@ -371,17 +371,16 @@ impl<Runtime: StorageEnableRuntime> Actor for FileTransferService<Runtime> {
                     file_key,
                     callback,
                 } => {
-                    let result = match self.peer_file_allow_list.insert((peer_id, file_key)) {
-                        true => Ok(()),
-                        false => Err(RequestError::FileAlreadyRegisteredForPeer),
-                    };
+                    if !self.peer_file_allow_list.insert((peer_id, file_key)) {
+                        trace!(target: LOG_TARGET, "File already registered for peer id {} and file key {:?}", peer_id, file_key);
+                    }
 
                     self.peers_by_file
                         .entry(file_key)
                         .or_insert_with(Vec::new)
                         .push(peer_id);
 
-                    match callback.send(result) {
+                    match callback.send(Ok(())) {
                         Ok(()) => {}
                         Err(_) => error!(
                             target: LOG_TARGET,
--- a/client/fisherman-service/src/handler.rs
+++ b/client/fisherman-service/src/handler.rs
@@ -1,6 +1,6 @@
 use codec::Decode;
 use futures::stream::{self, StreamExt};
-use log::{debug, error, info, warn};
+use log::{debug, error, info, trace, warn};
 use pallet_file_system_runtime_api::FileSystemApi;
 use sc_client_api::{BlockImportNotification, BlockchainEvents, HeaderBackend};
 use shc_common::types::{FileOperation, OpaqueBlock, StorageEnableEvents};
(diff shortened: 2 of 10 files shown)


client/file-transfer-service/src/handler.rs
```diff
@@ -371,17 +371,16 @@ impl<Runtime: StorageEnableRuntime> Actor for FileTransferService<Runtime> {
                     file_key,
                     callback,
                 } => {
-                    let result = match self.peer_file_allow_list.insert((peer_id, file_key)) {
-                        true => Ok(()),
-                        false => Err(RequestError::FileAlreadyRegisteredForPeer),
-                    };
+                    if !self.peer_file_allow_list.insert((peer_id, file_key)) {
+                        trace!(target: LOG_TARGET, "File already registered for peer id {} and file key {:?}", peer_id, file_key);
+                    }
```
ffarall, line 376: Good catch! Weirdly I can't think of a reason why we would want to consider this an error.

client/fisherman-service/src/handler.rs
```diff
@@ -107,41 +107,11 @@ impl<Runtime: StorageEnableRuntime> FishermanService<Runtime> {
                     self.emit(event);
                 }
                 StorageEnableEvents::FileSystem(
-                    pallet_file_system::Event::StorageRequestExpired { file_key },
+                    pallet_file_system::Event::IncompleteStorageRequest { file_key },
```
ffarall, line 110: Smart for maintainability

pallets/file-system/src/lib.rs
```diff
@@ -785,6 +785,11 @@ pub mod pallet {
             file_key: MerkleHash<T>,
             provider_id: ProviderIdFor<T>,
         },
+        /// Notifies that a storage request was marked as incomplete.
+        ///
+        /// This is important for fisherman nodes to listen and react to, to delete
+        /// the file key from the BSPs and/or MSP storing that file from their forest.
```
ffarall, line 791:
```suggestion
        /// the file key from the BSPs and/or Bucket storing that file from their forest.
```

pallets/file-system/src/utils.rs
```diff
@@ -1033,6 +1033,8 @@ where
                         &file_key,
                         incomplete_storage_request_metadata,
                     );
+
+                    Self::deposit_event(Event::IncompleteStorageRequest { file_key });
```
ffarall, line 1037: How about making a small helper function called `add_incomplete_storage_request` that adds it and emits the event. So that if we make changes in the future, we don't forget to ALWAYS send this event alongside it.

client/src/tasks/fisherman_process_file_deletion.rs
```diff
@@ -82,21 +83,32 @@ where
...
-        shc_indexer_db::models::bsp::BspFile::get_bsps_for_file_key(&mut conn, file_key.as_ref())
-            .await
-            .map_err(|e| anyhow!("Failed to query BSPs for file: {:?}", e))?;
+    let bsp_ids = BspFile::get_bsps_for_file_key(&mut conn, file_key.as_ref())
+        .await
+        .map_err(|e| anyhow!("Failed to query BSPs for file: {:?}", e))?;
+
+    // Check if MSP file association exists for this file
+    let bucket_target = match MspFile::get_msp_for_file_key(&mut conn, file_key.as_ref()).await {
+        Ok(Some(_msp_id)) => {
```
ffarall, line 92: Why only in the case that a bucket is linked to an MSP do you delete the file from the bucket? Couldn't it be that the bucket temporarily has no MSP (the MSP stopped storing it for instance)? And in that case, why would we prevent the user from removing a file from a bucket?

test/util/fisherman/fishermanHelpers.ts
```diff
@@ -36,6 +36,42 @@ export async function waitForDeleteFileExtrinsic(
...
 }
 
+/**
+ * Helper function to wait for delete_file_for_incomplete_storage_request extrinsic in transaction pool
+ * @param api - The API instance to use
+ * @param expectedCount - Number of expected delete_file_for_incomplete_storage_request extrinsics (default: 1)
+ * @param timeout - Timeout in milliseconds (default: 10000)
+ * @returns Promise<boolean> - True if expected number of extrinsics found, false if timeout
+ */
+export async function waitForDeleteFileForIncompleteStorageRequestExtrinsic(
```
ffarall, line 46: Well that's a hella long name lol

test/suites/integration/fisherman/process-file-deletion.test.ts
```diff
@@ -317,26 +386,36 @@ describeMspNet(
...
+      // Seal block to process the extrinsic
+      const deletionResult = await userApi.block.seal();
+
+      // Verify FileDeletedFromIncompleteStorageRequest event
+      assertEventMany(
+        userApi,
+        "fileSystem",
+        "FileDeletedFromIncompleteStorageRequest",
+        deletionResult.events
+      );
```
ffarall, line 418: In at least one of this tests, I think it would be worth adding a check for changing roots, additional to the on-chain event.

ffarall (approved):
Great job! I like the criteria you took in some decisions you made.

Remember to re-build `api-augment` for the CI.

test/suites/integration/fisherman/process-file-deletion.test.ts
```diff
@@ -260,39 +302,48 @@ describeMspNet(
       )
         .unwrap()
         .asRuntimeConfig.asStorageRequestTtl.toNumber();
-      await userApi.block.skipTo(currentBlockNumber + storageRequestTtl);
 
-      await waitForIndexing(userApi);
+      await userApi.block.skipTo(currentBlockNumber + storageRequestTtl - 1);
```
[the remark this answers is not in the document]
snowmead: we skip to a block before the ttl because we want to first verify if the extrinsic is in the mempool

test/util/bspNet/waits.ts
```diff
@@ -19,6 +19,7 @@ export const waitForTxInPool = async (api: ApiPromise, options: WaitForTxOptions
     checkQuantity,
     strictQuantity = false,
     shouldSeal = false,
+    finalizeBlock = true,
```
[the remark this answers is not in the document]
snowmead: imo I didn't introduce any complexity and all tests expect finalization by default (hence the default) - it is equivalent to having `shouldSeal` as a parameter in the function.
if you insist, I will remove it

node/src/service.rs
```diff
@@ -694,6 +698,9 @@ where
     configure_and_spawn_fisherman(
```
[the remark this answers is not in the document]
snowmead: Running as a provider or fisherman roles are now mutually exclusive

node/src/service.rs
```diff
@@ -685,15 +704,6 @@ pub fn new_partial_parachain(
         telemetry
     });
 
-    // FIXME: The `config.transaction_pool.options` field is private, so for now use its default value
```
ffarall, line 688: Let's keep this

.github/workflows/api-augment-publish.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 jobs:
   publish-api-augment:
-    runs-on: ubuntu-latest
+    runs-on: blacksmith-4vcpu-ubuntu-2404
```
ffarall, line 11: Rollback the runners in this PR.

PR #467, opened 2025-09-08
feat(backend): implements backend download endpoint

HermanObst:
### Summary
- Add GET `/download/{file_key}` to stream files as attachments.
- Use RPC (`storagehubclient_saveFileToDisk`) via `MspService::get_file_from_key`.
- On Unix, unlink temp file immediately after open (race-free, crash-safe).

### Notable changes
- RPC client: `call(...)` added, `ToRpcParams` used end-to-end.
- Mock RPC: concrete behavior for `saveFileToDisk`: writes mock file into `uploads/`.
- SDK: `downloadByKey(fileKey)` (removed `bucketId`). Examples/docs updated.
- Tests: added download test (verifies status, body, and temp file cleanup).

### Breaking
- Call sites must pass `rpc_params![]` instead of `()` for empty params.
- SDK: `downloadByKey(bucketId, fileKey)` → `downloadByKey(fileKey)`.

### Config
- `MspService::new` now requires `msp_callback_url`.

### DIFF
--- a/backend/lib/src/services/msp.rs
+++ b/backend/lib/src/services/msp.rs
@@ -334,13 +334,13 @@ impl MspService {
         let temp_path = format!("uploads/{}", file_key);
         let upload_url = format!("{}/{}", self.msp_callback_url, temp_path);
 
-        // Create params
-        let params = jsonrpsee::rpc_params![file_key, &upload_url];
-
         // Make the RPC call to download file and get metadata
         let rpc_response: SaveFileToDisk = self
             .rpc
-            .call("storagehubclient_saveFileToDisk", params)
+            .call(
+                "storagehubclient_saveFileToDisk",
+                (file_key, upload_url.as_str()),
+            )
             .await
             .map_err(|e| Error::BadRequest(e.to_string()))?;
 


backend/lib/src/api/handlers.rs
```diff
@@ -169,19 +171,52 @@ pub async fn download_by_location(
...
+
+    // Open file for streaming
+    let file = File::open(&download_result.temp_path)
+        .await
+        .map_err(|e| Error::BadRequest(format!("Failed to open downloaded file: {}", e)))?;
+
+    let stream = ReaderStream::new(file);
+    let file_stream_resp = FileStream::new(stream).file_name(&filename).into_response();
+
+    // Schedule cleanup of temp file after response
```
Karrq, line 214: This isn't scheduled tho - it will most likely just be done right after we return from this task.
The FileStream probably doesn't buffer the entire file in memory.

backend/lib/src/api/routes.rs
```diff
@@ -41,18 +45,12 @@ pub fn routes(services: Services) -> Router {
...
         )
-        .route(
-            "/buckets/{bucket_id}/download/{file_key}",
-            get(handlers::download_by_key),
-        )
-        .route(
-            "/buckets/{bucket_id}/download/path/{*file_location}",
-            get(handlers::download_by_location),
-        )
+        .route("/files/{file_key}", get(handlers::download_by_key))
```
Karrq, line 53: I guess you removed teh `buckets/{bucket_id}` since it's not needed, right?

backend/lib/src/api/routes.rs
```diff
@@ -19,6 +19,10 @@ pub fn routes(services: Services) -> Router {
         )
         .route_layer(DefaultBodyLimit::disable());
 
+    let internal_file_upload = Router::new()
+        .route("/uploads/{file_key}", put(handlers::internal_upload_by_key))
```
Karrq, line 23: I don't quite understand this internal path. and also it doesn't seem internal because we don't authenticate it at all.

btw the separate router was because `route_layer` applies to all routes in the router, so if you made a new router just for the body limit you can include it in the previous one

backend/lib/src/config.rs
```diff
@@ -47,6 +50,8 @@ pub struct StorageHubConfig {
     pub max_concurrent_requests: Option<usize>,
     /// Whether to verify TLS certificates for secure connections
     pub verify_tls: bool,
+    /// URL for the node to reach the MSP backend
+    pub msp_callback_url: String,
```
Karrq, line 54: So this will be used by the MSP node, right? Probably to upload the file or something.
Sooo we have the callback_url configured to be reachable by the MSP node, and then the backend does the RPC call to the backend node passing in something like `{callback_url}/uploads/file_key`, using the internal endpoint.
Did I get that right?

backend/lib/src/constants/mod.rs
```diff
@@ -24,6 +24,9 @@ pub mod rpc {
     /// Default RPC WebSocket URL
     pub const DEFAULT_RPC_URL: &str = "ws://localhost:9944";
 
+    /// Default MSP callback URL
+    pub const DEFAULT_MSP_CALLBACK_URL: &str = "http://host.docker.internal:8080";
```
Karrq, line 28: would probably have localhost here like above

backend/lib/src/services/mod.rs
```diff
@@ -65,6 +67,7 @@ impl Services {
         // Create in-memory storage
         let memory_storage = InMemoryStorage::new();
         let storage = Arc::new(BoxedStorageWrapper::new(memory_storage));
+        let callback_base_url = String::from("http://localhost:8080");
```
Karrq, line 70: use the same default value as the config

msp.toml.example
Karrq, line 1: we already have `backend/backend_config.toml`

backend/lib/src/services/msp.rs
```diff
@@ -308,6 +336,35 @@ impl MspService {
...
+            .map_err(|e| Error::BadRequest(format!("Failed to parse RPC response: {}", e)))?;
+
+        // Convert location bytes to string
+        let location = String::from_utf8_lossy(&rpc_response.success.location).to_string();
+
+        Ok(FileDownloadResult {
+            file_size: rpc_response.success.file_size,
+            location,
+            fingerprint: rpc_response.success.fingerprint,
+            temp_path: temp_path,
```
Karrq, line 365: Ok I see how this works now. I have a small doubt tho, we save the file to a temporary path, which we expect the MSP Node to start uploading the file to our "internal" endpoint, right?

We probably shouldn't expect the file to be written to completely already at this stage, so what happens to the responded `FileStream`? Does it get updated as more data is written to file?

Actually, i the other case, where the RPC returns when the entire file is written, do we want to start streaming only until after the file is completely transfered from Node to Backend?

Karrq (changes requested):
I have some questions but other than that it's fine. Not what I was thinking, but we can improve it later, as long as we acknowledge and correctly mark what we will need to change later.

backend/lib/src/api/routes.rs
HermanObst: yup

backend/lib/src/constants/mod.rs
HermanObst: I added that as Im using it that way to test manually: 
- I have a MSP node running in a container 
- I run the backend locally 

But I dont have a strong preference

backend/lib/src/services/msp.rs
HermanObst: So we are not going to have the file_path until the receiving FileStream finishes

backend/bin/src/main.rs
```diff
@@ -65,7 +69,7 @@ async fn main() -> Result<()> {
 
     let postgres_client = create_postgres_client(&config).await?;
     let rpc_client = create_rpc_client(&config).await?;
-    let services = Services::new(storage, postgres_client, rpc_client);
+    let services = Services::new(storage, postgres_client, rpc_client, config.clone());
```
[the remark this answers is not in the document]
Karrq: I suggested passing in the entire config to the services struct as I have this change in my endpoints and also in auth - in general I don't think it's a bad idea to pass in the config in to `Services` so some items can be "self configured"

backend/backend_config.toml
```diff
@@ -21,6 +21,9 @@ max_page_size = 100
 # WebSocket RPC endpoint for the StorageHub node
 # Use ws:// for unencrypted or wss:// for TLS-encrypted connections
 rpc_url = "ws://localhost:9944"
+# URL for node to communicate with the backend (e.g for uploading a file)
+# e.g: if running an msp in a container: http://host.docker.internal:8080
+msp_callback_url = "http://localhost:8080"
```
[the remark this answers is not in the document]
HermanObst: Leaving this for now

backend/bin/src/main.rs
```diff
@@ -44,6 +44,10 @@ struct Args {
     /// Override RPC URL
     #[arg(long)]
     rpc_url: Option<String>,
+
+    /// Override MSP callback URL
+    #[arg(long)]
+    msp_callback_url: Option<String>,
```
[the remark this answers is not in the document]
HermanObst: Leaving this for now

backend/lib/src/api/handlers.rs
```diff
@@ -169,19 +171,71 @@ pub async fn download_by_location(
...
 
-    // TODO(MOCK): return proper data
-    let file_data = b"Mock file content for download".to_vec();
-    let stream = ReaderStream::new(Cursor::new(file_data));
-    let file_stream_resp = FileStream::new(stream).file_name("by_key.txt");
+    // Validate file_key is a hex string
+    let key = file_key.trim_start_matches("0x");
+    if hex::decode(key).is_err() {
+        return Err(Error::BadRequest("Invalid file key".to_string()));
+    }
```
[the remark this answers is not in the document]
HermanObst: addressed above

backend/lib/src/api/handlers.rs
```diff
@@ -169,19 +171,71 @@ pub async fn download_by_location(
...
+        .await
+        .map_err(|e| Error::BadRequest(format!("Failed to open downloaded file: {}", e)))?;
+
+    // On Unix, unlink the path immediately; the open fd remains valid for streaming
+    // TODO: we should implement proper cleanup after the stream is closed
+    // But as we will probably change implementation to just redirect the RPC stream to user, leaving it as is for now (not a problem if we run on unix).
+    #[cfg(unix)]
+    {
+        let _ = tokio::fs::remove_file(&download_result.temp_path).await;
+    }
```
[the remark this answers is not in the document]
HermanObst: Yup! Since we’re Unix-only, unlink-after-open is safe (I wanted to add the flag just to have it very clear).
tempfile would helps only if we change how we create/clean the temp path. Right now the RPC uploads by URL path (internal_upload_by_key writes to a path), not an open file handle.

We could workaround that, but as in a soon-to-come follow up we will just redirect the stream, leaving as is for now :)

backend/lib/src/data/rpc/mock_connection.rs
```diff
@@ -144,38 +180,39 @@ impl Default for MockConnection {
...
+        let response: Value = match method {
+            "storagehubclient_saveFileToDisk" => self.mock_save_file_to_disk(params).await,
+            _ => {
+                let responses = self.responses.read().await;
+                responses
+                    .get(method)
+                    .cloned()
+                    .unwrap_or(serde_json::json!(null))
+            }
         };
```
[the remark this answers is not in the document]
HermanObst: Created a ticket in our board to keep track of this

backend/lib/Cargo.toml
```diff
@@ -26,6 +26,7 @@ tokio = { workspace = true, features = ["full"] }
 tokio-util = { workspace = true }
 toml = { workspace = true }
 tower-http = { workspace = true }
+shc-rpc = { workspace = true, features = ["std"] }
 
 # Workspace dependencies
```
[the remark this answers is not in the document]
HermanObst: You are right :P

backend/lib/src/api/routes.rs
```diff
@@ -61,10 +59,14 @@ pub fn routes(services: Services) -> Router {
 
 #[cfg(all(test, feature = "mocks"))]
 mod tests {
+    use crate::constants::mocks::DOWNLOAD_FILE_CONTENT;
+    use crate::services::health::HealthService;
```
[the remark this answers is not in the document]
HermanObst: amended

backend/lib/src/services/msp.rs
```diff
@@ -308,6 +336,41 @@ impl MspService {
...
+
+        // Make the RPC call to download file and get metadata
+        let res: Value = self
+            .rpc
+            .call("storagehubclient_saveFileToDisk", params)
+            .await
+            .map_err(|e| Error::BadRequest(e.to_string()))?;
+
+        let rpc_response: RpcResponse = serde_json::from_value(res)
+            .map_err(|e| Error::BadRequest(format!("Failed to parse RPC response: {}", e)))?;
```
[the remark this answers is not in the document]
HermanObst: addressed

PR #469, opened 2025-09-10
Feat: missing GETs

ftheirs:
Add missing GETs to SDK to retrieve information from MSP
- getInfo
- getStats
- getValueProposition
- getFileMetadata

Include all these new methods into E2E tests
Update the examples projects

Make token field public in MspClient in order to allow the FrontEnd to correctly handle when it's None


TDemeco (approved):
Nice job 💯

PR #476, opened 2025-09-16
fix: Use MSP backend binary as entrypoint in Docker image (don't hardcode args)

stiiifff:


TDemeco:
#473 has been merged so this fix is not needed anymore, closing

PR #488, opened 2025-09-22
refactor(ci): Update GitHub Actions workflow to build and push sh-msp-backend

snowmead:
Changes to Docker workflows:

- Renamed job from `build_image` to `build_binaries` for clarity.
- Updated build steps to compile both `storage-hub-node` and `sh-msp-backend`.
- Added new jobs to push Docker images for both node and backend binaries.
- Adjusted artifact handling to include both binaries in the workflow.
- Ensured all dependent jobs reference the new `build_binaries` job.

### DIFF
--- a/.github/workflows/parachain.yml
+++ b/.github/workflows/parachain.yml
@@ -75,7 +75,11 @@ jobs:
       - name: Set tag name
         id: set-tag
         run: |
-          if [[ "${{ env.SKIP_BUILD_LABEL_PRESENT }}" == "true" && "${{ env.SKIP_BACKEND_BUILD_LABEL_PRESENT }}" == "true" ]] || [[ "${{ steps.node_check.outputs.changed }}" == 'false' && "${{ steps.backend_check.outputs.changed }}" == 'false' ]] || "${{ github.ref }}" == 'refs/heads/main'; then
+          if [[ "${{ github.ref }}" == 'refs/heads/main' ]]; then
+            echo "tag=latest" >> $GITHUB_OUTPUT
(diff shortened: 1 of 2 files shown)


snowmead:
Not sure if this is the only place we should be checking for changes for the rebuilding the node - seems correct to me
https://github.com/Moonsong-Labs/storage-hub/blob/49807d812b0ef8ea1287f7d4f0634d58943b1479/.github/workflows/parachain.yml#L65

santikaplan:
> Not entirely sure why we are not pushing to docker when we merge to main (followed for now the same workflow process as was implemented for the node image)
> 
> https://github.com/Moonsong-Labs/storage-hub/blob/49807d812b0ef8ea1287f7d4f0634d58943b1479/.github/workflows/parachain.yml#L80-L83

Hey! What do you mean here? We should push with every merge to main yes (if there where any changes on the Backend).

snowmead:
> > Not entirely sure why we are not pushing to docker when we merge to main (followed for now the same workflow process as was implemented for the node image)
> > https://github.com/Moonsong-Labs/storage-hub/blob/49807d812b0ef8ea1287f7d4f0634d58943b1479/.github/workflows/parachain.yml#L80-L83
> 
> Hey! What do you mean here? We should push with every merge to main yes (if there where any changes on the Backend).

It runs on pull requests not just pushes to main. So my understanding is that as it stands we are pushing to Docker before merging.
https://github.com/Moonsong-Labs/storage-hub/blob/49807d812b0ef8ea1287f7d4f0634d58943b1479/.github/workflows/parachain.yml#L26-L30

PR #492, opened 2025-09-24
Feat delete file and bucket

ftheirs:
Improve requestDeleteFile input params
Add integration tests for `requestDeleteFile` and `deleteBucket`

### DIFF
--- a/sdk/core/src/evm/storageHubClient.ts
+++ b/sdk/core/src/evm/storageHubClient.ts
@@ -38,10 +38,11 @@ type FileSystemContract<TClient extends EvmClient> = GetContractReturnType<
 >;
 
 /**
- * Internal constant precompile address for FileSystem on StorageHub runtimes.
+ * Default precompile address for FileSystem on StorageHub runtimes.
  * If a chain uses a different address, this constant should be updated accordingly.
  */
-const FILE_SYSTEM_PRECOMPILE_ADDRESS = "0x0000000000000000000000000000000000000064" as Address;
+export const SH_FILE_SYSTEM_PRECOMPILE_ADDRESS =
+  "0x0000000000000000000000000000000000000064" as Address;
 
 export class StorageHubClient {
   private readonly publicClient: PublicClient; // Internal for gas estimation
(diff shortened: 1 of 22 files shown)


sdk/core/src/evm/storageHubClient.ts
```diff
@@ -365,36 +412,39 @@ export class StorageHubClient {
...
-   * @param size - file size as bigint (storage units)
-   * @param fingerprint - 32-byte file fingerprint
-   * @param options - optional gas and fee overrides
+   * Request deletion of a file from the network.
+   * @param fileInfo File information containing all required data
+   * @param operation File operation to perform (defaults to Delete)
+   * @param options Optional transaction options
+   * @returns Transaction hash
    */
   async requestDeleteFile(
```
TDemeco: It's called `requestDeleteFile` so the `operation` should never be allowed to be anything else other than `FileOperation.Delete`, which means it shouldn't even be an argument to this function

TDemeco: Or the other option, rename the function to `requestFileOperation` and don't default to deletions but force the user to pass the actual file operation it wants to perform (although for now we only have the deletion operation)

sdk/msp-client/src/modules/files.ts
```diff
@@ -101,29 +101,57 @@ export class FilesModule extends ModuleBase {
...
     const headers = this.withAuth(baseHeaders);
-    const res = await this.ctx.http.getRaw(path, {
-      ...(headers ? { headers } : {}),
-      ...(options?.signal ? { signal: options.signal } : {})
-    });
 
-    if (!res.body) {
-      throw new Error("Response body is null - unable to create stream");
-    }
+    try {
```
TDemeco: I changed this to not throw but to return the error code in the `DownloadResult`'s status field. We should do the same for other endpoints that return `XyzResult` so we can actually process the result of the requests instead of throwing. I'd even venture that it would be nice to have a field where we can specify what failed so the requester can debug the issue.
All this is for another PR though, we should add a task for it

TDemeco (approved):
Approved with some comments! Feel free to merge once they are resolved 👍🏻

PR #493, opened 2025-09-24
Feat: use biome in SDK

ftheirs:
Replace Prettier + ESLint with Biome in SDK
This PR replaces Prettier and ESLint with Biome and unifies the formatting and linting commands to match the conventions already used in the repo:
- pnpm fmt → checks formatting
- pnpm fmt:fix → applies formatting
- pnpm lint → checks for lint issues
- pnpm lint:fix → fixes lint issues

The PR includes a large number of changes, but most of them are simple replacements (e.g., switching from single to double quotes) to align with the repo’s original formatting rules. It’s recommended to review this commit by commit, as the changes are grouped by functionality.


.github/workflows/sdk.yml
```diff
@@ -105,15 +105,18 @@ jobs:
           key: wasm-pack-v0.12.1
       - name: Install wasm-pack
         run: curl -sSf https://rustwasm.github.io/wasm-pack/installer/init.sh | sh
-      - name: Install dependencies
+      - name: Install root dependencies
         run: pnpm install --frozen-lockfile
+        working-directory: .
+      - name: Install SDK dependencies
+        run: pnpm install --frozen-lockfile
```
TDemeco, line 112: I might be wrong but, since we have all SDK packages in the `pnpm-workspace.yaml` file of the repo, wouldn't doing `pnpm i` in the root repo just install everything required? Why do we need to do this twice?

.github/workflows/sdk.yml
```diff
@@ -105,15 +105,18 @@ jobs:
...
+      - name: Format check
+        run: pnpm fmt
+      - name: Lint
+        run: pnpm lint
```
TDemeco, line 116: I just realized, do we have an action that prevents us pushing badly formatted SDK code? Basically a job that runs this and checks if there is any badly formatted or linter errors, and if so, errors

sdk/package.json
```diff
@@ -17,26 +17,19 @@
...
     "clean": "node scripts/clean.js",
     "coverage": "pnpm --filter ./core run coverage && pnpm --filter ./msp-client run coverage",
-    "format": "pnpm --filter ./core run format && pnpm --filter ./msp-client run format",
-    "format:check": "pnpm --filter ./core run format:check && pnpm --filter ./msp-client run format:check",
-    "lint": "pnpm --filter ./core run lint && pnpm --filter ./msp-client run lint",
-    "lint:fix": "pnpm -r --if-present run lint:fix",
+    "fmt": "../node_modules/.bin/biome format .",
+    "fmt:fix": "../node_modules/.bin/biome format . --write",
+    "lint": "../node_modules/.bin/biome lint .",
+    "lint:fix": "../node_modules/.bin/biome lint . --write",
```
TDemeco, line 23: Hmm is this the cleanest way to access the biome binary? I see that in the `package.json` under the `test/` folder we directly call biome, like `"fmt": "biome format . --write"`

sdk/biome.json
TDemeco, line 1: Couldn't we just extend the workspace biome config? By making this file only be this:
```json
{
  "extends": ["../biome.json"]
}

TDemeco (approved):
Nice work! Left a few comments to make the change a bit cleaner and more maintainable 👍🏻  trust you can take care of them

sdk/biome.json
ftheirs: I'll check this again 🔍  but I was getting a different behaviour when I call `pnpm fmt` from root or sdk folder. Probably it was not getting correctly the json.
Anyways, I have some specific ignores like `"!**/_wasm_embed.ts"` that is to prevent to verify the generated file with wasm code from format ---> If you run format after building the SDK, biome will complain about this file but the generated code is not pushed into the repo.

TDemeco: Makes sense! That's the nice part about extending the root config: you can add extra stuff on top of it specific to the SDK while inheriting all the common config

.github/workflows/sdk.yml
ftheirs: These commands are not applying the linter suggestion or the format. They will fail if they detect something different. The one that do stuff are the one with `...:fix`

TDemeco: Damn I had a brainfart moment 😅 . I know these ones don't fix the formatting or linting but while reviewing this I was thinking that I remembered we having a specific action to check something like this in another part of the repo, but it was the typegen stuff

sdk/biome.json
ftheirs: Done! Each package references to the biome.json from SDK folder. This one contains special config.
The lint and fmt commands can be executed from root, sdk or internal packages (core, msp-client) and it works well on every case

PR #496, opened 2025-09-24
feat: ✨ Add CLI config param for MSP to distribute files

ffarall:
1. Add `enable_msp_distribute_files` flag in Blockchain Service, that determines whether or not a running MSP should even look for storage requests it has to distribute files for.
2. Avoid looking at existing storage requests for this MSP to distribute, or listen to related events if `enable_msp_distribute_files = false`.
3. Add flag to MSP sample config file.
4. Add flag to MSP CLI config params.

### DIFF
--- a/node/src/cli.rs
+++ b/node/src/cli.rs
@@ -175,10 +175,7 @@ pub struct ProviderConfigurations {
 
     /// Enable MSP file distribution to BSPs (disabled by default unless set via config/CLI).
     /// Only applicable when running as an MSP provider.
-    #[arg(long, value_name = "BOOLEAN", required_if_eq_all([
-        ("provider", "true"),
-        ("provider_type", "msp"),
-    ]))]
+    #[arg(long, value_name = "BOOLEAN")]
     pub msp_distribute_files: bool,
 
     // ============== Provider RPC options ==============


client/src/builder.rs
```diff
@@ -790,6 +794,10 @@ pub struct BlockchainServiceOptions {
     pub max_blocks_behind_to_catch_up_root_changes: Option<u32>,
     /// The peer ID of this node.
     pub peer_id: Option<Vec<u8>>,
+
```
HermanObst, line 797: remove this line

docker/fullnet-base-template.yml
```diff
@@ -83,6 +84,7 @@ services:
         '--sealing=manual',
         '--base-path=/data',
         '--msp-charging-period=12',
+        '--msp-distribute-files',
```
HermanObst, line 87: Shouldn't this be =true or something like that?

node/src/cli.rs
```diff
@@ -173,6 +173,11 @@ pub struct ProviderConfigurations {
     #[arg(long, default_value = "10")]
     pub max_blocks_behind_to_catch_up_root_changes: Option<u32>,
 
+    /// Enable MSP file distribution to BSPs (disabled by default unless set via config/CLI).
+    /// Only applicable when running as an MSP provider.
+    #[arg(long, value_name = "BOOLEAN")]
+    pub msp_distribute_files: bool,
```
snowmead, line 179: Not sure if there is a similar field which restricts this field to only be supplied if this condition is met `("provider_type", "msp"),`

snowmead (approved):
Left a small comment regarding the CLI option 👍

node/src/cli.rs
ffarall: I can't add that condition directly from the `arg` macro, but this flag is ignored when processed, unless the provider type is MSP.

PR #500, opened 2025-09-26
test: :bug: Ensure BSP volunteered in upload file integration test

snowmead:
Hit an [error](https://github.com/Moonsong-Labs/storage-hub/actions/runs/18046378830/job/51358587248?pr=499) where the BSP confirmation transaction was not found in the transaction pool. 

This change will ensure that at least the volunteer extrinsic happened in the previous block which normally makes this flow deterministic.


TDemeco (approved):
Nice catch! Now that we are distributing the file we care for at least one BSP having volunteered 👍🏻

PR #503, opened 2025-10-01
feat: unify authentication process

ftheirs:
WIP


sdk/msp-client/src/modules/buckets.ts
```diff
@@ -1,10 +1,41 @@
...
+  }
+
+  /** List files/folders under a path for a bucket (root if no path) */
+  getFiles(bucketId: string, options?: GetFilesOptions): Promise<FileListResponse> {
+    const headers = this.withAuth();
+    const path = `/buckets/${encodeURIComponent(bucketId)}/files`;
+    return this.ctx.http.get<FileListResponse>(path, {
+      ...(headers ? { headers } : {}),
+      ...(options?.signal ? { signal: options.signal } : {}),
+      ...(options?.path ? { query: { path: options.path.replace(/^\/+/, "") } } : {})
```
Karrq: Could you add a comment explaining what this replace does, in words? I'm not that good at regex, but to me it seems like it removed all `/` instances at the start of the string, right?
So `/foo/bar` becomes `foo/bar`?

sdk/msp-client/src/modules/files.ts
```diff
@@ -1,10 +1,289 @@
...
+    type FileInfoWire = Omit<FileInfo, "uploadedAt"> & { uploadedAt: string };
+    return this.ctx.http
+      .get<FileInfoWire>(path, {
+        ...(headers ? { headers } : {}),
+        ...(options?.signal ? { signal: options.signal } : {})
+      })
+      .then(
+        (wire): FileInfo => ({
+          ...wire,
+          uploadedAt: new Date(wire.uploadedAt)
```
Karrq: This is to convert the unix timestamp into a ISO style string, right?

sdk/msp-client/src/modules/files.ts
```diff
@@ -1,10 +1,289 @@
...
+  }
+
+  /** List files under a bucket path (root if no path) */
+  getFiles(bucketId: string, options?: GetFilesOptions): Promise<FileListResponse> {
+    const headers = this.withAuth();
+    const path = `/buckets/${encodeURIComponent(bucketId)}/files`;
+    return this.ctx.http.get<FileListResponse>(path, {
+      ...(headers ? { headers } : {}),
+      ...(options?.signal ? { signal: options.signal } : {}),
+      ...(options?.path ? { query: { path: options.path.replace(/^\/+/, "") } } : {})
```
Karrq: Same as the other comment related to the regex

sdk/msp-client/src/modules/files.ts
```diff
@@ -1,10 +1,289 @@
...
+      stream: res.body,
+      status: res.status,
+      contentType,
+      contentRange,
+      contentLength
+    };
+  }
+
+  /** Download a file by its location path under a bucket */
+  async downloadByLocation(
```
Karrq: This doesn't exist anymore

sdk/e2e/page/msp.html
```diff
@@ -489,13 +400,9 @@ <h3>Errors</h3>
...
-        if (!window.__msp.token) {
-          const jwt = generateMockJwt();
-          client.setToken(jwt);
-          window.__msp.token = jwt;
-          console.log('[TOKEN][AUTO]', jwt);
-        }
-        const dl = await client.downloadByLocation(bucketId, path);
+        const auth = await client.auth.getAuthStatus();
+        if (!auth || auth.status !== 'Authenticated') throw new Error('Not authenticated');
+        const dl = await client.files.downloadByLocation(bucketId, path);
```
Karrq, line 405: yeah this one we should remove since that endpoint doesn't exist anymore

sdk/e2e/page/msp.html
```diff
@@ -345,107 +298,53 @@ <h3>Errors</h3>
...
+              if (method === 'eth_chainId') return '0x2c3e6'; // 181222
+              throw new Error('Unsupported JSON-RPC method in test transport');
+            }
+          })
+        });
+
+        await client.auth.SIWE(wallet);
+        const auth = await client.auth.getAuthStatus();
+        window.__msp.status = auth?.status;
+        setText('authStatus', auth?.status || 'unknown');
```
TDemeco: No need for the `unknown` fallback

PR #504, opened 2025-10-03
fix: FileList response type

ftheirs:
FileListResponse type didn't contain children field in its type definition


Karrq (approved):

TDemeco (approved):
Good enough for now to have type completion but we should have a nicer way of displaying files

PR #506, opened 2025-10-06
feat(file-system): batch file deletion support

snowmead:
Enables multiple files to be deleted in the `delete_files` and `delete_files_for_incomplete_storage_request` extrinsics.

This will increase the file deletion throughput on a per block basis since prior to this change, fisherman nodes would have to submit sequential extrinsics with incrementing nonces to delete multiple files from the same Bucket Forest or BSP Forest. Without this, they would only be able to delete a single file per block per Forest.

The number of files that can be deleted are bounded by a new config parameter `MaxFileDeletionsPerExtrinsic`.

Note: Client side batching implementation will be done in another PR.

### DIFF
--- a/api-augment/package.json
+++ b/api-augment/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@storagehub/api-augment",
-  "version": "0.2.6",
+  "version": "0.2.7",
   "description": "",
   "scripts": {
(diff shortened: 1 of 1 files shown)


ffarall (approved):
Just some quick changes that I trust you to take care of. Otherwise neat job!

PR #507, opened 2025-10-06
feat: Support for Payment streams on SDK

ftheirs:
Add getPaymentStreams method to MspClient in SDK
Add E2E and integration tests for this new feature

### DIFF
--- a/test/suites/integration/solochain-evm/sdk-precompiles.test.ts
+++ b/test/suites/integration/solochain-evm/sdk-precompiles.test.ts
@@ -302,14 +302,26 @@ await describeMspNet(
     });
 
     it("Should fetch payment streams using the SDK's MspClient", async () => {
-      // Retrieve payment streams for the authenticated user and print details for review
-      const streamsResp = await mspClient.getPaymentStreams();
-      strictEqual(streamsResp.streams.length, 1, "payment streams array length should be 1");
+      // Get on-chain information for payment streams
+      const mspId = userApi.shConsts.DUMMY_MSP_ID;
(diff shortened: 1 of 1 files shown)


test/suites/integration/solochain-evm/sdk-precompiles.test.ts
```diff
@@ -301,6 +301,17 @@ await describeMspNet(
...
+    it("Should fetch payment streams using the SDK's MspClient", async () => {
+      // Retrieve payment streams for the authenticated user and print details for review
+      const streamsResp = await mspClient.getPaymentStreams();
+      strictEqual(streamsResp.streams.length, 1, "payment streams array length should be 1");
+
+      const ps = streamsResp.streams[0];
+      strictEqual(ps.providerType, "msp", "providerType should be 'msp'");
+      strictEqual(ps.totalAmountPaid, "100000", "totalAmountPaid should be '100000'");
+      strictEqual(ps.costPerTick, "90664", "costPerTick should be '90664'");
+    });
```
TDemeco, line 313: Let's not hardcode this, but instead:
- Query the chain for the expected payment stream between the user and the MSP using `await userApi.query.paymentStreams.fixedRatePaymentStreams(mspId, account.address)`
- Use the SDK to get all the payment streams of the user, and filter the one between the user and this MSP, making sure it exists.
- Compare what's received from the SDK to what's on chain (rate, provider) to ensure they match. We don't have a `totalAmountPaid` on chain so we can safely ignore that one

TDemeco (approved):
Looks good 👍🏻 minor test change

PR #717, opened 2026-03-26
refactor(indexer): 🗑️ Remove `bucket_id` column from `file` table

ffarall:
## Summary

This PR removes the legacy `file.bucket_id` dependency from the indexer and backend code, and treats `onchain_bucket_id` as the single canonical bucket identifier throughout the codebase.

It also adds the required database migrations to enforce `bucket.onchain_bucket_id` uniqueness, drop the redundant `file.bucket_id` column and index, and update the generated schema snapshot accordingly.

## Key Changes

- Add indexer DB migrations to make `bucket.onchain_bucket_id` unique and remove `file.bucket_id`.
- Update Diesel schema and models to create, query, and maintain files and bucket stats using `onchain_bucket_id` directly.
- Refactor repository, DB client, MSP service, and indexer event handlers to stop translating through the internal bucket row ID.
- Refresh mocks, snapshots, and tests to match the new identifier flow and keep the `mocks` feature compiling cleanly.

## Why

This simplifies the data model around the chain-native bucket identifier, removes redundant joins and lookups, and aligns the backend/indexer behaviour with the assumption that each on-chain bucket ID maps to exactly one canonical bucket row.

## ⚠️ Breaking Changes ⚠️

- **Short description**

`bucket_id` column removed from `file` table, which now only relies on `onchain_bucket_id`.

- **Who is affected**

  - `🟤 [Indexer operators]`: A migration will be auto-run when the client upgrades, eliminating the `bucket_id` field and making `onchain_bucket_id` unique (provided `SH_INDEXER_DB_AUTO_MIGRATE` env var is NOT set to `false`)
  
- **Suggested code changes**

None. Migrations should apply automatically.


client/indexer-db/migrations/2026-03-25-000001_make_bucket_onchain_bucket_id_unique/up.sql
```diff
@@ -0,0 +1,6 @@
+-- Enforce bucket on-chain ID uniqueness.
+-- This matches current code assumptions that bucket lookups by onchain_bucket_id
+-- resolve to a single canonical bucket row.
+DROP INDEX IF EXISTS idx_bucket_blockchain_id;
+
+CREATE UNIQUE INDEX idx_bucket_blockchain_id ON bucket(onchain_bucket_id);
```
ffarall, line 6: We should check if this is actually true and under no circumstance there can be a duplicated `onchain_bucket_id` in the indexer db

client/indexer-db/src/models/bucket.rs
```diff
@@ -177,12 +177,12 @@ impl Bucket {
     /// Increment file count and update total size
     pub async fn increment_file_count_and_size<'a>(
         conn: &mut DbConnection<'a>,
-        bucket_id: i64,
+        onchain_bucket_id: Vec<u8>,
```
TDemeco: Why not `&[u8]` instead of `Vec<u8>`? The Diesel `.filter()` accepts it and it avoid us having to `.clone` the vec when we call this

client/indexer-db/src/models/bucket.rs
```diff
@@ -195,12 +195,12 @@ impl Bucket {
     /// Decrement file count and update total size
     pub async fn decrement_file_count_and_size<'a>(
         conn: &mut DbConnection<'a>,
-        bucket_id: i64,
+        onchain_bucket_id: Vec<u8>,
```
TDemeco: Same here

backend/lib/src/data/indexer_db/repository/postgres.rs
```diff
@@ -130,15 +130,15 @@ impl IndexerOps for Repository {
...
     async fn get_files_by_bucket(
         &self,
-        bucket: i64,
+        bucket: &Hash,
         limit: i64,
         offset: i64,
     ) -> RepositoryResult<Vec<File>> {
         let mut conn = self.pool.get().await?;
 
         // Same as File::get_by_bucket_id but with pagination
```
TDemeco: Stale comment, that fn does not exist anymore

TDemeco (approved):
Looks good, just some nitpicks 💯

